PR #1, opened 2021-04-29
Add initial foundations

illusional:
Adds initial foundations for sample metadata system:

- A DB schema, applied using `liquibase`
- A set of DB classes + models
    - Tables + Layers that can interact with a MySQL database
    - If `export SM_ENVIRONMENT="PRODUCTION"`, then it will pull credentials from `projects/sample_metadata/secrets/databases/`, maintains one connection per database.
- A web API with Flask + `apispec`:
    - uses OpenAPI3 to generate a swagger UI, and an installable `sample_metadata` python API.
    - Authentication of the web API using the GCP auth tokens.
    - Currently NO authorization - except the web API will be deployed behind an Google Identity-aware proxy (IAP)
    - Capturing the current user from this bearer token
- Installable API
    - Generated from OpenAPI3 annotations, using the [openapi-generator](https://openapi-generator.tech/docs/installation/). 
    - `regenerate_api.py` is a script that can generate the installable sample_metadata API. Can use an existing instance, or spin up an instance (including using Docker)

PR #7, opened 2021-05-27
Make update_sample_status pass future linting

illusional:
This [fails #1 's new linting](https://github.com/populationgenomics/sample-metadata/runs/2682781340), so fixing it here so the pull-request will pass.

Linting isn't enabled in this branch, but a few errors:

- end-of-file-fixer
- trailing-whitespace
- MD031/blanks-around-fences Fenced code blocks should be surrounded by blank lines
- Q002 Remove bad quotes from docstring


lgruen (approved):
Thanks!

PR #9, opened 2021-06-03
Backup db

vivbak:


db/backup/backup.py
```diff
@@ -0,0 +1,64 @@
...
+        logger.log_text(text, severity='INFO')
+
+    except subprocess.SubprocessError:
+        text = f'Failed to export backup {timestamp_str} UTC.'
+        logger.log_text(text, severity='ERROR')
+        return
+
+    # Write backup to file
+    filepath = f'backups/backup_{timestamp_str}.sql'
+    with open(filepath, 'xb') as f:
```
illusional, line 41: This file isn't ever deleted, and could fill the db hdd pretty quickly.

db/backup/backup.py
```diff
@@ -0,0 +1,64 @@
...
+    log_name = 'backup_log'
+    logger = logging_client.logger(log_name)
+
+    # Get timestamp
+    utc_now = pytz.utc.localize(datetime.utcnow())
+    timestamp_str = utc_now.strftime('%d_%m_%Y_%H-%M-%S')
+
+    # Export SQL Data
+    try:
+        backup_sql = subprocess.check_output(
```
illusional, line 28: Probably want to write this to a temporary file, otherwise we're keeping the whole db in memory, to then write. Looking at the tempfile module in Python, you could create a tmpdir name, wrap it in a `with` (context block) and it get's cleaned up automatically.

```python
with tempfile.TemporaryFile() as fp:
    result = subprocess.run([command], stdout=fp)
    if result.rc != 0:
        # fail state
        pass
    else:
        # success
        blob.upload_from_filename(filepath)
```

db/backup/backup.py
vivbak: Great point! Have addressed in the latest commit (hopefully)

db/backup/backup.py
vivbak: This suggestion was really useful, but I had issues implementing it with the new mariadb backup function. The created folder is quite difficult to remove without using a sudo rm. Would be very interested if you knew of an alternative fix though.

db/backup/backup.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+        text = f'Performed mariabackup to pull data {timestamp_str} UTC.'
+        logger.log_text(text, severity='INFO')
+
+    except subprocess.SubprocessError:
+        text = f'Failed to export backup {timestamp_str} UTC.'
+        logger.log_text(text, severity='ERROR')
+        return
+
+    # Grant appropraite permissions for tmp_dir
```
illusional, line 46:
```suggestion
    # mariabackup creates awkward permissions for the output files,
    # so we'll grant appropriate permissions for tmp_dir to later remove it
```

db/backup/backup.py
```diff
@@ -0,0 +1,78 @@
...
+from google.cloud import storage
+from google.cloud import logging
+import pytz
+
+
+def perform_backup():
+    """Completes a backup of the databases within a local mariadb instance
+    and uploads this backup to GCS."""
+
+    # Logging
```
illusional, line 17:
```suggestion
    # Logging: Any log with `severity >= ERROR` get's logged to #software-alerts
```

illusional (approved):
Great work on the back-up script! I've added a couple of small suggestions to comments, but looks great 👍

PR #10, opened 2021-07-14
Add create_project script

illusional:
Adds script that can be ran on database server to create a new project and updates the secret with the new credentials.
After this script, a restart of the sample-metadata cloud run service is required. Not sure how to do that yet.


vivbak (approved):
Works wonderfully!

PR #12, opened 2021-07-26
Update Cloud Function with Upload Processor 

vivbak:
Workflow is as follows
1. gVCFs and CRAMS are moved.
2. Following a successful move, analysis objects are created.
3. Once analysis objects are successfully created, sequencing metadata is updated


vivbak:
Updated Workflow: 
`main.py` defines an endpoint, that when hit, formats a file object. This file object, and the associated qc_metadata is added to the sample metadata db. 
`process_backlog.py` exists to process previous batches.

update_sample_status/main.py
```diff
@@ -1,63 +1,120 @@
...
-    return ('', 204)
+
+    if not project or not external_id or not status or not batch or not qc_metadata:
+        raise web.HTTPBadRequest(
+            reason=f'Invalid request. Ensure that the project, sample, status, batch and metadata are provided.'
+        )
+
+    # Determine the internal ID
+    external_id_dict = {'external_ids': [external_id]}
+    internal_id_map = sapi.get_sample_id_map_by_external(project, external_id_dict)
```
illusional, line 80: This recently changed, so you don't need to wrap it in a dictionary anymore.

```suggestion
    internal_id_map = sapi.get_sample_id_map_by_external(project, [external_id])
```

update_sample_status/main.py
```diff
@@ -1,63 +1,120 @@
...
+
+    reads = []
+    for key in FILES_HANDLED:
+        reads = create_file_object(external_id, key, project, batch)
+        # TODO: Added the QC Metadata Here. Should this be moved to sample metadata?
+        metadata = {'reads': reads, 'qc_metrics': qc_metadata}
+        # Update the sequencing metadata.
+        update_sequence_meta(project, internal_id, status, metadata)
+
+    return web.Response(text=f'{external_id} lodged.')
```
illusional, line 91: It looks like you set `reads` to an empty array, and then set on every iteration, but I think it'd be better to not set everything against reads (as initially suggested), and instead against the different types:

- `reads` for (`.cram`)
- `gvcf` for (`.gvcf`)

If you add `'meta-field'` to the FILES_HANDLED dict, your code could look something like:

```suggestion
    # TODO: Added the QC Metadata Here. Should this be moved to sample metadata?
    meta = {'qc_metrics': qc_metadata}
    for key in FILES_HANDLED:
        reads = create_file_object(external_id, key, project, batch)
        
        field_name = FILES_HANDLED[key]['meta-field']
        meta[field_name] = reads
        meta[field_name + '_type'] = key
        
    # Update the sequencing metadata.
    update_sequence_meta(project, internal_id, status, metadata)

    return web.Response(text=f'{external_id} lodged.')
```

update_sample_status/main.py
```diff
@@ -1,63 +1,120 @@
...
+from sample_metadata.models import SequencingStatus
 
 
-secret_manager = secretmanager.SecretManagerServiceClient()
+routes = web.RouteTableDef()
 
+FILES_HANDLED = {
+    'gvcf': [{'primary': '.g.vcf.gz', 'secondary': ['.g.vcf.gz.tbi', '.g.vcf.gz.md5']}],
+    'cram': [{'primary': '.cram', 'secondary': ['.cram.crai', '.cram.md5']}],
+}
```
illusional, line 23: To address future comment.

```suggestion
FILES_HANDLED = {
    'gvcf': [{'primary': '.g.vcf.gz', 'secondary': ['.g.vcf.gz.tbi', '.g.vcf.gz.md5'], 'meta-field': 'gvcf'}],
    'cram': [{'primary': '.cram', 'secondary': ['.cram.crai', '.cram.md5'], 'meta-field': 'reads'}],
}
```

illusional (changes requested):
This is looking pretty good!

PR #13, opened 2021-07-27
Format intermal sample ids with hyphens for readability

vladsaveliev:
What do you think about additionally formatting sample numbers with hyphens for readability? CPG141243 -> CPG-1412-43. 
I  usually find it to improve the experience when doing the data analysis.

(We could look into turning IDs into random meaningful expressions - similar to the docker container names (`nostalgic_stallman` and `focused_yonath`), that would improve communicating samples even further. Or even follow Loic's suggestion and include some real metadata into the IDs, like the abbreviated dataset name.)

### DIFF
--- a/models/models/sample.py
+++ b/models/models/sample.py
@@ -42,7 +42,7 @@ def sample_id_transform_to_raw(
     identifier: Union[List[Union[str, int]], Union[str, int]], strict=True
 ):
     """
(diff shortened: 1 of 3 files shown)


illusional:
I don't mind the hyphens, my first thought is that someone will mistake the chunks into having some meaning for the separation. A part from this I'm happy with it. Can you post this in the sample-metadata slack to get a consensus, then I'm happy to merge.

PR #14, opened 2021-07-28
Add first draft of ingestion script

illusional:
- Parse format
- Construct new File object (with md5 + secondaries)
- Add to SM system

### DIFF
--- a/scripts/parse_vcgs_manifest.py
+++ b/scripts/parse_vcgs_manifest.py
@@ -1,0 +1,381 @@
+from typing import List, Dict, Union, Optional
+
+import os
+import re
+import csv
+import logging
+from io import StringIO
+from itertools import groupby
+from collections import defaultdict
+
+from sample_metadata.model.sequence_type import SequenceType
+
+from sample_metadata.model.sample_type import SampleType
+
+from sample_metadata.model.sequence_status import SequenceStatus
+
+from sample_metadata.apis import SampleApi, SequenceApi
+from sample_metadata.model.new_sample import NewSample
+from sample_metadata.model.new_sequence import NewSequence
+
+
+rmatch = re.compile(r'_[Rr]\d')
+
+logger = logging.getLogger(__file__)
+logger.addHandler(logging.StreamHandler())
+logger.setLevel(logging.INFO)
+
+
+class Columns:
+    PHS_ACCESSION = 'phs_accession'
+    SAMPLE_NAME = 'sample_name'
+    LIBRARY_ID = 'library_ID'
+    TITLE = 'title'
+    LIBRARY_STRATEGY = 'library_strategy'
+    LIBRARY_SOURCE = 'library_source'
+    LIBRARY_SELECTION = 'library_selection'
+    LIBRARY_LAYOUT = 'library_layout'
+    PLATFORM = 'platform'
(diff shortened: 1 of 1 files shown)


scripts/parse_vcgs_manifest.py
```diff
@@ -0,0 +1,381 @@
...
+        type_ = types[0].lower()
+        if type_ == 'wgs':
+            return 'wgs'
+        if type_ in ('single-cell', 'ss'):
+            return 'single-cell'
+
+        raise ValueError(f'Unrecognised sequencing type {type_}')
+
+    def parse_reads(self, reads: List[str]) -> Union[List[Dict], Dict]:
+        if all(r.endswith('.fastq') or r.endswith('.fastq.gz') for r in reads):
```
vladsaveliev, line 241: Not relevant for this specific manifest, but to make the code reusable in future, can also check for the `.fq.gz` common extention

scripts/parse_vcgs_manifest.py
```diff
@@ -0,0 +1,381 @@
...
+
+        type_ = types[0].lower()
+        if type_ == 'wgs':
+            return 'wgs'
+        if type_ in ('single-cell', 'ss'):
+            return 'single-cell'
+
+        raise ValueError(f'Unrecognised sequencing type {type_}')
+
+    def parse_reads(self, reads: List[str]) -> Union[List[Dict], Dict]:
```
vladsaveliev, line 240:
```suggestion
    def parse_reads(self, reads: List[str]) -> Union[List[List[Dict]], List[Dict]]:
```

scripts/parse_vcgs_manifest.py
```diff
@@ -0,0 +1,381 @@
...
+                for col in Columns.sequence_columns()
+            }
+            collapsed_type = self.parse_sequencing_type(
+                sample_name, [r[Columns.LIBRARY_STRATEGY] for r in rows]
+            )
+            collapsed_sample_meta = {
+                col: ",".join(set(r[col] for r in rows))
+                for col in Columns.sample_columns()
+            }
+            collapsed_sample_meta['reads'] = reads
```
vladsaveliev, line 170: Should we also add some field for the file type (`fastq`/`cram`/`bam`)? Though probably not a big deal, the loading pipeline would check the extension (or whether the object is a list of lists for fastqs).

vladsaveliev (approved):
Awesome!!

Just minor comments, approving.

scripts/parse_vcgs_manifest.py
```diff
@@ -0,0 +1,381 @@
...
+            return files
+
+        extensions = set(os.path.splitext(r)[-1] for r in reads)
+
+        joined_reads = ''.join(f'\n\t{i}: {r}' for i, r in enumerate(reads))
+        raise ValueError(
+            f'Mixed, or unrecognised extensions ({", ".join(extensions)}) for reads:{joined_reads}'
+        )
+
+    def parse_fastqs_structure(self, fastqs):
```
vladsaveliev, line 278: Maybe worth adding type hints here. And probably clarify about the return structure. Is it a list of pairs (R1/R2), or a pair of lists (different lanes for the same read orientation)? Realised that I got confused :)

PR #16, opened 2021-08-11
Genericise ingestor

illusional:
- Genericise the importer, so can be easily subclassed, and a manifest can be imported.
- It's the subclasser's responsibility to form the sample and sequence meta fields, including the read objects
- The `Parser` base provides methods to help construct file objects, including with secondaries and interacting with GS.

FYI @vladsaveliev, @lgruen, @vivbak

### DIFF
--- a/scripts/parser.py
+++ b/scripts/parser.py
@@ -201,9 +201,7 @@ class GenericParser:
         for external_sample_id in sample_map:
             logger.info(f'Preparing {external_sample_id}')
             rows = sample_map[external_sample_id]
-            # TODO: remove this
-            if external_sample_id not in external_id_map:
-                continue
+
             if len(rows) == 1:
                 rows = rows[0]
             # now we have sample / sequencing meta across 4 different rows, so collapse them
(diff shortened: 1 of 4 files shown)


scripts/parser.py
```diff
@@ -0,0 +1,433 @@
...
+        # by external_sample_id
+        sequencing_to_add: Dict[str, List[NewSequence]] = defaultdict(list)
+
+        samples_to_update: Dict[str, SampleUpdateModel] = {}
+        sequences_to_update: Dict[int, SequenceUpdateModel] = {}
+
+        for external_sample_id in sample_map:
+            logger.info(f'Preparing {external_sample_id}')
+            rows = sample_map[external_sample_id]
+            # TODO: remove this
```
lgruen, line 204: Leftover?

scripts/parser.py
```diff
@@ -0,0 +1,433 @@
...
+                    )
+                )
+
+        if not self.sample_metadata_project:
+            logger.info('No sample-metadata project set, so skipping add into SM-DB')
+            return samples_to_add, sequencing_to_add
+
+        logger.info(f'Adding {len(samples_to_add)} samples to SM-DB database')
+        ext_sample_to_internal_id = {}
+        # for new_sample in samples_to_add:
```
lgruen, line 249: Leftover?

lgruen (changes requested):

scripts/parser.py
illusional: Oh phew!

scripts/parser.py
illusional: Yes, thank you!!

scripts/parser.py
```diff
@@ -0,0 +1,433 @@
...
+                secondaries.append(self.create_file_object(sec_file))
+
+        return secondaries
+
+
+def apply_secondary_file_format_to_filename(
+    filepath: Optional[str], secondary_file: str
+):
+    """
+    You can trust this function to do what you want
```
vladsaveliev, line 407: Just a minor comment, I usually try to prefix internal functions with an underscore: `_apply_secondary_file_format_to_filename`, in which case linter doesn't require a docstring. But if a function is meant to be imported, better to add a meaningful docstring.

scripts/parser.py
```diff
@@ -0,0 +1,433 @@
...
+        raise ValueError(
+            f'Mixed, or unrecognised extensions ({", ".join(extensions)}) for reads: {joined_reads}'
+        )
+
+    @staticmethod
+    def parse_fastqs_structure(fastqs) -> List[List[str]]:
+        """
+        Takes a list of fastqs, and a set of nested lists of each R1 + R2 read.
+
+        >>> GenericParser.parse_fastqs_structure(['20210727_PROJECT1_L002_R2.fastq.gz', '20210727_PROJECT1_L002_R1.fastq.gz', '20210727_PROJECT1_L001_R2.fastq.gz', '20210727_PROJECT1_L001_R1.fastq.gz'])
```
vladsaveliev, line 338: Note that fastq names can also have a "segment ID" in the end (e.g. `SampleName_S1_L001_R1_001.fastq.gz`):
https://support.illumina.com/help/BaseSpace_OLH_009008/Content/Source/Informatics/BS/NamingConvention_FASTQ-files-swBS.htm
Which is though always `_001`, so shouldn't cause issues.

illusional: Ooh that's a great resource, thanks for sharing!

PR #19, opened 2021-08-12
Fix more parser imports

illusional:
Can't run through analysis-runner without fixing this import
I don't know why pycharm can run it correctly 😓


vladsaveliev (approved):

vladsaveliev:
Not a good idea to import from a script in `scripts`, but can refactor later :)

PR #22, opened 2021-08-13
Fix sequences route

illusional:
+ cheeky fix to getSamples active check


vivbak (approved):
🙏 Thanks so much Michael!!

PR #23, opened 2021-08-17
Add project-related permissions

illusional:
Look-up permissions per project based on secrets. The name of the secret for read / write operations are specified in the database, and are inferred on creation. 

In your local development (you'll have to set `SM_ALLOWALLACCESS=true`) to not use secrets.  There is some caching (~1 second) applied to looking up secrets, but after this PR merge (before production), I'll be checking secret access times. 

Broadly, there are very few changes to the conda API, there have been a few small route name changes.

Routes are slightly trickier to add now:

- If your route requires a project, ensure any resources are limited to that project
- If your route DOESN'T require a project, you'll need to implement a reverse-lookup:
  - Match the resource back to its parent's project, eg (sequence -> sample)

Note the following resources that belong to projects:
- Family
- Person
- Sample
- Analysis
  - If an analysis has sample ids, you're only expected to have access to the relevant analysis. Note, the creator of an analysis should have access to projects for all specified sample Ids

There are methods on the `ProjectPermissionsTable` to make this easier, for example:

```python3
async def update_sample(self, id_, ...)
  # check user has access to sample before updating
  project = await self.connection.fetch_val('SELECT project FROM sample WHERE id = :id', {'id': id_})
  ptable = ProjectPermissionsTable(self.connection)
  await ptable.check_access_to_project_id(self.author, project)
  # remainder of code to update sample ...
```


api/server.py
```diff
@@ -72,9 +67,16 @@ async def exception_handler(_: Request, e: Exception):
         code = determine_code_from_error(e)
         name = str(type(e).__name__)
 
+    base_params = {'name': name, 'description': str(e)}
+
+    if add_stacktrace:
+        st = traceback.format_exc()
+        logger.error(traceback.format_exc())
+        base_params['stacktrace'] = st
+
```
illusional, line 75: I reordered this, so the stacktrace appears at the bottom of the json response, dictionary insert order matters here.

vladsaveliev: Cool! 
Could you write `logger.error(st)` in the line 74?

db/python/tables/project.py
```diff
@@ -1,51 +1,309 @@
...
+        project_id = await self.connection.fetch_val(_query, values)
+
+        if create_test_project:
+            values = {
+                'name': project_name + '-test',
+                'dataset': dataset_name,
+                'gcp_id': gcp_project_id,
+                'author': author,
+                'read_secret_name': read_secret_name.replace('-main-', 'test'),
+                'write_secret_name': write_secret_name.replace('-main-', 'test'),
```
vladsaveliev, line 304: Perhaps `"test"` should be surrounded by dashes in these 2 lines too?
```suggestion
                'write_secret_name': write_secret_name.replace('-main-', '-test-'),
```

db/python/tables/project.py
```diff
@@ -1,51 +1,309 @@
...
+
+        project_id = await self.connection.fetch_val(_query, values)
+
+        if create_test_project:
+            values = {
+                'name': project_name + '-test',
+                'dataset': dataset_name,
+                'gcp_id': gcp_project_id,
+                'author': author,
+                'read_secret_name': read_secret_name.replace('-main-', 'test'),
```
vladsaveliev, line 303:
```suggestion
                'read_secret_name': read_secret_name.replace('-main-', '-test-'),
```

vladsaveliev (approved):
That's awesome!!

I pushed a few commits with minor fixes, and also updated the tests in `test`, and added a README for the tests.

I might think about how to set up CI to run those tests automatically. Can do in a separate PR, so approving this now :)

PR #25, opened 2021-08-20
Dev deploy fixes

illusional:
Introduce dev-deploy for the sample-metadata API.

- The `.github/workflows/deploy.conf` has been updated to deploy a different cloud run deployment with new DB.
- Adds `dev{sha}` onto the end of the version for a dev conda deploy.
- Modifies the sample prefix, and adjusts the checksum so sample IDs can't clash between environments.

Notably, `SM_ENVIRONMENT` is only used by the installable conda API

- `SM_ENVIRONMENT=PRODUCTION`: goes to production SM server
- `SM_ENVIRONMENT=DEVELOPMENT`: Goes to managed-development SM server
- `SM_ENVIRONMENT=LOCAL`: Goes to local API server (`localhost:${PORT || 8000}`)

Breaks some deploy options into separate environment variables:
- `SM_CREDS` instead of hard-coded secret path. The expectation is the CloudRun service will make these available as an environment variable.
- `SM_IGNORE_GCP_CREDENTIALS_ERROR`: Ignore errors on GCP init for credential issues, useful for generating the API, as you don't need GCP.
- `SM_ENABLE_GCP_LOGGING`: Enables GCP logging
- `SM_LOGGING_LEVEL`: Configure logging level, default DEBUG.
- `SM_SAMPLEPREFIX`: Sample prefix, default `CPGLCL` - please don't change this
- `SM_SAMPLECHECKOFFSET` - sample checksum offset, default 2 - please don't change this.


api/utils/gcp.py
```diff
@@ -4,11 +4,18 @@
...
-IGNORE_GCP_CREDENTIALS_ERROR = bool(os.getenv('SM_IGNORE_GCP_CREDENTIALS_ERROR'))
+levels_map = {'DEBUG': logging.DEBUG, 'INFO': logging.INFO, 'WARNING': logging.WARNING}
 
+IGNORE_GCP_CREDENTIALS_ERROR = os.getenv('SM_IGNORE_GCP_CREDENTIALS_ERROR') in (
+    'y',
+    'true',
+    '1',
+)
+USE_GCP_LOGGING = os.getenv('SM_ENABLE_GCP_LOGGING', '0').lower() in ('y', 'true', '1')
+LOGGING_LEVEL = levels_map.get(os.getenv('SM_LOGGING_LEVEL'), logging.DEBUG)
```
vladsaveliev, line 15: Just to make sure the case is ignored:

```suggestion
LOGGING_LEVEL = levels_map[os.getenv('SM_LOGGING_LEVEL', 'DEBUG').upper()]
```

.github/workflows/deploy.yaml
```diff
@@ -3,7 +3,8 @@ on:
   push:
     branches:
       - main
-      - add-foundations # test the deploy
+      - dev
+      - dev-deploy
```
vladsaveliev, line 7: Awesome! So to rebuild the dev server/package/image, one should push to either one of `dev` or `dev-deploy`?

vladsaveliev (approved):

.github/workflows/deploy.yaml
illusional: Should be just `dev`, I'll remove dev-deploy :)

PR #28, opened 2021-08-27
Validate Backup

vivbak:


db/backup/recovery_test.py
```diff
@@ -0,0 +1,246 @@
...
+                '--prepare',
+                f'--target-dir={LOCAL_BACKUP_FOLDER}/{backup_folder}',
+            ],
+            check=True,
+        )
+        # Stop the MariaDB server
+        subprocess.run(['sudo', 'systemctl', 'stop', 'mariadb'], check=True)
+
+        if os.path.isdir('/var/lib/mysql/'):
+            print('Backup cannot be performed unless /var/lib/mysql is empty.')
```
illusional, line 57: Should `isdir` always return true, or are you checking if the folder doesn't exist?

db/backup/recovery_test.py
```diff
@@ -0,0 +1,246 @@
...
+        # Fix permissions
+        subprocess.run(
+            ['sudo', 'chown', '-R', 'mysql:mysql', '/var/lib'],
+            check=True,
+        )
+        # Start the MariaDB server
+        subprocess.run(['sudo', 'systemctl', 'start', 'mariadb'], check=True)
+
+    def setUp(self):
+        self.prod_connection = ConnectionDetails('10.152.0.2', 'backup')
```
illusional, line 79: For the prod connection:

- We can't guarantee the internal IP won't change, so we should use internal url.
- Aren't there passwords that go with this? Can we put most of these into environment variables (or a secret) so they can be changed and not captured with the script?

(Might need to add another comment to the script on how to set this up / run it).

db/backup/recovery_test.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # Tables with unique id fields.
+        tables = [
+            'analysis',
+            'family',
+            'participant',
+            'project ',
+            'sample',
+            'sample_sequencing',
+        ]
```
illusional, line 135: This now has a unique ID field:
```suggestion
            'family_participant',
        ]
```

db/backup/recovery_test.py
```diff
@@ -0,0 +1,246 @@
...
+            sample_ids = []
+            for sample in restored_random_results:
+                sample_ids.append(sample[0])
+
+            if len(sample_ids) == 0:
+                continue
+
+            sample_string = ','.join([str(elem) for elem in sample_ids])
+            prod_random_results = get_results(
+                f"SELECT * FROM {table} FOR SYSTEM_TIME AS OF TIMESTAMP'{self.timestamp}' WHERE id in ({sample_string}) ;",
```
illusional, line 151: Nice catch, running this for `TIMESTAMP'{self.timestamp}` is great!

Where we're using _uncontrolled_ values in a query, we need to escape the values before we use them in the query, [Relevant XKCD](https://xkcd.com/327/). The `mysql.connector` library has functionality for this: [Prevent SQL Injection
](https://www.w3schools.com/python/python_mysql_where.asp#:~:text=mycursor.execute(sql%2C%20adr))

db/backup/recovery_test.py
```diff
@@ -0,0 +1,246 @@
...
+    conn = mysql.connector.connect(user=connection.user, host=connection.address)
+    cursor = conn.cursor()
+    cursor.execute('show databases')
+    databases = list(cursor)
+    return databases
+
+
+def get_results(query: str, connection: ConnectionDetails):
+    """ Returns the results from a provided query at a given connection """
+    conn = mysql.connector.connect(user=connection.user, host=connection.address)
```
illusional, line 241: Can we create the connection once (rather than pass the credentials), and reuse it (by creating a new cursor) for each request?

A `mysql.connector` allows you to specify the `database` to connect to up front (to avoid the separate USE statement): https://dev.mysql.com/doc/connector-python/en/connector-python-connectargs.html#:~:text=Table%C2%A07.1%C2%A0Connection%20Arguments%20for%20Connector/Python

illusional (changes requested):
It's super important that we prevent sql injection by using `mysql.connector`'s string escaping for us.


I think it's possible to parameterize the testing interface, and then use [parameterized](https://github.com/wolever/parameterized).



For example:

```py
from typing import Optional, Tuple

tables = [
	('analysis_sample', ('analysis_id', 'sample_id')),
	('sample', ('id', )),
	# ... more configs here
]


def test_random_rows(self, tablename: str, id_fields: Tuple[str]):
	"""
	Test 10 random rows in the analysis_sample table
	"""
	id_fields_str = ', '.join(id_fields)
	restored_random_results = get_results(
	    f'SELECT {id_fields}, * FROM {tablename} ORDER BY RAND() LIMIT 10',
	    self.local_connection,
	)

	for row in restored_random_results:
	    ids = row[:len(id_fields)]

	    # Note the uncontrolled value (that we get from a database, is represented by %s)
	    wheres_str = ' AND '.join(f'{field} = %s' for field in id_fields)

	    _query = f"""\
SELECT {id_fields_str}, * 
FROM analysis_sample 
FOR SYSTEM_TIME AS OF TIMESTAMP'{self.timestamp}' 
WHERE {wheres_str};"""

	    prod_random_results = get_results(
	        self.prod_connection,
	        _query,
	        ids,
	    )
	    self.assertEqual(sample, prod_random_results[0])
```

WHERE this is the updated get_results:

```python
def get_results(connection: ConnectionDetails, query: str, values: Optional[Tuple]=None):
    """ Returns the results from a provided query at a given connection """
    conn = mysql.connector.connect(user=connection.user, host=connection.address, database=DATABASE)
    cursor = conn.cursor()
    cursor.execute(query, values)
    results = list(cursor)
    return results

```

db/backup/recovery_test.py
vivbak: Checking that the folder doesn't exist. In hindsight, we should catch and exit here. I'll amend :)

db/backup/recovery_test.py
vivbak: Great point regarding the internal IP -- I've switched this out to the URL. I didn't originally set passwords, because I instead set the permissions for the new accounts to only work at certain IP addresses -- but in hindsight, that's not a great idea. I've created passwords and moved all this information including into a secret.

db/backup/recovery_test.py
vivbak: Awesome ty

db/backup/recovery_test.py
vivbak: Great point, have hopefully fixed now.

vivbak:
Hi @illusional, thanks so much for the review! That was super helpful in cleaning up + securing this script. Hopefully, I've addressed everything now -- let me know if not.

illusional (changes requested):
This is looking really great @vivbak! I'm stoked I could see all the tests pass. A few things to look at before we merge it:

1. I'm noticing a lot of warnings for google-drive. It seems to work but do you know where these errors are coming from?

```
/home/michael_franklin/.local/lib/python3.7/site-packages/google/cloud/storage/acl.py:643: ResourceWarning: unclosed <ssl.SSLSocket fd=8, family=AddressFamily.AF_INET, type=SocketKind.
SOCK_STREAM, proto=6, laddr=('10.152.0.17', 47972), raddr=('142.250.66.208', 443)>
  super(ObjectACL, self).__init__()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/michael_franklin/.local/lib/python3.7/site-packages/google/cloud/storage/acl.py:643: ResourceWarning: unclosed <socket.socket fd=7, family=AddressFamily.AF_INET, type=SocketKind.
SOCK_STREAM, proto=6, laddr=('10.152.0.17', 46390), raddr=('169.254.169.254', 80)>
  super(ObjectACL, self).__init__()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
./usr/lib/python3.7/json/decoder.py:353: ResourceWarning: unclosed <socket.socket fd=11, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('10.152.0.17', 46880
), raddr=('169.254.169.254', 80)>
  obj, end = self.scan_once(s, idx)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/usr/lib/python3.7/json/decoder.py:353: ResourceWarning: unclosed <ssl.SSLSocket fd=12, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('10.152.0.17', 41388)
, raddr=('142.250.66.176', 443)>
  obj, end = self.scan_once(s, idx)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```


2. Can we add instructions on how to use this script, there are a few dependencies that I had to install to test (maybe a `requirements.txt`), and how to set this script up + run the tests.

Thanks!!

illusional (changes requested):
Looking really great, thanks for putting the docs together. I've got a few requests docs to make it more straight forward to answer: "how do I do the DB validation" step. You've already done the hard work of creating the db user and secret.

db/backup/README.md
```diff
@@ -0,0 +1,159 @@
...
+2. In the production instance of the database, create a user that has read access to all of the tables.
+3. In the Secret Manager, create a secret `db-validate-backup`, with a JSON config as follows, where p_username and p_password matches the user created in step 2 above.
+
+```json
+{
+  "Dbname": "sm_production",
+  "P_host": "sm-db-vm-instance.australia-southeast1-b.c.sample-metadata.internal",
+  "P_username": "backup",
+  "p_password": "example_password123",
+  "l_host": "localhost",
```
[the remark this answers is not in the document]
vivbak: Good point -- updated in the latest.

db/backup/recovery_test.py
```diff
@@ -0,0 +1,160 @@
...
+).payload.data.decode('UTF-8')
+config = json.loads(config_str)
+
+DATABASE = config['dbname']
+
+PROD_HOST = config['p_host']
+PROD_USER = config['p_username']
+PROD_PASSWORD = config['p_password']
+
+LOCAL_USER = config['l_username']
```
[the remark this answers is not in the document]
vivbak: Done :)

vivbak:
Hey @illusional thanks so much for the review! I've tried to address all of the feedback provided. I seem to have forgotten what you said would be involved for this step though (In terms of testing the SM API) 

> 1. Update the configuration to point to the new VM as the production VM.

https://github.com/populationgenomics/sample-metadata/blob/disaster-recovery/db/backup/README.md#yearly-procedures

db/backup/README.md
```diff
@@ -1,5 +1,86 @@
...
+
+   > ```bash
+   > python3 restore.py
+   > ```
+
+#### Rebuild VM
+
+1. Create a new VM instance in the GCP Console.
+2. Enable storage write permissions. On the create instance page, scroll down to Access Scopes. Select Set access for each API. Set Storage to Read/Write.
+3. Select a service account. The Compute Engine default service account will be selected by default.
```
illusional, line 55: I think this should be the `sm-db-sa@sample-metadata.iam.gserviceaccount.com` service-account to mirror the existing db instance.

db/backup/README.md
```diff
@@ -1,5 +1,86 @@
...
+   > cd sample-metadata/db/backup
+   > ```
+
+5. Restore the database
+
+   > ```bash
+   > python3 restore.py
+   > ```
+
+#### Rebuild VM
```
illusional, line 51:
```suggestion
#### Rebuild Database VM
```

db/backup/README.md
```diff
@@ -1,5 +1,86 @@
...
+   1. [Validation Procedure](#Validation-Procedure)
+   2. [Validation Plan](#Validation-Plan)
+
+## Recovery Procedures
+
+The full recovery process is detailed below. Follow the recovery stages that are applicable. It is worth noting that individual tables or databases cannot be restored. In the event that only one table or database has been lost, the entire instance will need to be restored from the last backup point.
+
+### Recover Sample Metadata Database
+
+1. If the `sm-dm-vm-instance` needs to be restored, create a new instance from the [sm-db-vm-image](https://console.cloud.google.com/compute/machineImages/details/sm-dm-vm-image?project=sample-metadata). If you can't access the image, you can rebuild the instance from scratch. See [Rebuild VM](#Rebuild-VM) for further instructions.
```
illusional, line 24:
```suggestion
1. If the `sm-dm-vm-instance` needs to be restored, create a new instance from the [sm-db-vm-image](https://console.cloud.google.com/compute/machineImages/details/sm-dm-vm-image?project=sample-metadata). If you can't access the image, you can rebuild the instance from scratch. See [Rebuild Database VM](#Rebuild-Database-VM) for further instructions.
```

illusional (approved):
Looks great! A few small suggestions, but this is awesome. Great work!!

PR #35, opened 2021-09-06
Add test github workflow

vladsavelyev:
Added `.github/workflows/test.yaml` that sets up a local dev environment, populates the database, and runs a test that simulates the SM queries performed by a joint-calling workflow.


test/README.md
```diff
@@ -1,84 +1,109 @@
...
+-u$SM_DEV_DB_USER -p$SM_DEV_DB_PASSWORD \
+-e 'use '$SM_DEV_DB_PROJECT'; insert into project \
+(id, name, author, dataset, gcp_id, read_secret_name, write_secret_name) \
+values \
+(1, "'$INPUT_PROJECT'", "'$USER'", "'$INPUT_PROJECT'", "'$GCP_ID'", "'$INPUT_PROJECT'-ci-sample-metadata-main-read-members-cache", "'$INPUT_PROJECT'-ci-sample-metadata-main-write-members-cache"), \
+(2, "'$OUTPUT_PROJECT'", "'$USER'", "'$OUTPUT_PROJECT'", "'$GCP_ID'", "'$OUTPUT_PROJECT'-ci-sample-metadata-main-read-members-cache", "'$OUTPUT_PROJECT'-ci-sample-metadata-main-write-members-cache");'
+
+mysql --host=$SM_DEV_DB_HOST --port=$SM_DEV_DB_PORT \
+-u $SM_DEV_DB_USER -p$SM_DEV_DB_PASSWORD  \
+-e 'use '$SM_DEV_DB_PROJECT'; select * from project;'
```
illusional, line 87:
```suggestion
-u $SM_DEV_DB_USER -p$SM_DEV_DB_PASSWORD $SM_DEV_DB_PROJECT  \
-e 'select * from project;'
```

.github/workflows/test.yaml
```diff
@@ -0,0 +1,125 @@
...
+      - name: gcloud setup
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: sample-metadata
+          service_account_key: ${{ secrets.GCP_SERVER_DEPLOY_KEY }}
+          export_default_credentials: true
+
+      - name: gcloud docker auth
+        run: |
+          gcloud auth configure-docker australia-southeast1-docker.pkg.dev
```
illusional, line 29: I don't think we need this for the unittesting

.github/workflows/test.yaml
```diff
@@ -0,0 +1,125 @@
...
+        run: python regenerate_api.py
+
+      - name: install python api
+        run: sudo pip install -e .
+
+      - name: populate samples
+        run: |
+          python3 -m api.server &
+          sleep 3  # to make sure server finished starting
+          python test/test_add_samples_for_joint_calling.py
```
illusional, line 115: Can all tests be run in `SM_ALLOWALLACCESS=0`?

illusional (changes requested):
This is looking awesome, I'm so excited! This will make a big difference to the sample-metadata server code-base.

Is it possible for us to add tests in some testing framework?

Each test could declare sample it wants in `setupClass`, and then we can use `unittest` to manage running the tests (and ask it to run the occasional doctests that are littered around).

.github/workflows/test.yaml
vladsaveliev: Will it be able to check secrets without that?

Also happy to use a separate service account rather than the deploy one

.github/workflows/test.yaml
vladsaveliev: Of course, but there might be a benefit in covering the permission checking feature with tests too?

illusional:
Hey @vladsavelyev, I'm going to close this due to the newish testing framework, and this being out-of-date. Feel free to reopen at anytime

PR #36, opened 2021-09-07
Improvements to seqr individual meta parser + generic SM runner script

illusional:
- Improvements seqr individual metadata parser
- Adds a generic sample-metadata runner script for running in analysis runner

For example, I want to upload some metadata to sample-metadata from the upload bucket `gs://cpg-$dataset-main-upload/individual-metadata-form.csv`. This would involve:

```python
import sample_metadata
import_api = sample_metadata.api.ImportApi()
response = import_api.import_individual_metadata_manifest(project='', file='<path>')
```

I've wrapped that up, so you can call:

```shell
cd scripts
analysis-runner \
    --dataset $dataset --description "upload $dataset individual metadata' \
    -o metadata-upload --access-level standard \
    arbitrary_sm.py import import_individual_metadata_manifest \
    --file-to-localise "gs://cpg-$dataset-main-upload/individual-metadata-form.csv" \
    --json "{\"project\": \"$dataset\", \"file\": \"individual-metadata-form.csv\"}"
```

HELP:
```shell
./arbitrary_sm.py --help
# usage: Arbitrary sample-metadata script [-h]
#                                         [--file-to-localise FILE_TO_LOCALISE]
#                                         [--json JSON] [--args ARGS [ARGS ...]]
#                                         api_name method_name
# 
# positional arguments:
#   api_name
#   method_name
# 
# optional arguments:
#   -h, --help            show this help message and exit
#   --file-to-localise FILE_TO_LOCALISE
#                         List of GS files to localise
#   --json JSON           JSON encoded dictionary for kwargs
#   --args ARGS [ARGS ...]
#                         any positional arguments to pass to the API
```

You can test this works like with:

```shell
python arbitrary_sm.py sample get_all_sample_id_map_by_internal \
    --json "{\"project\": \"$dataset\"}"
```


db/python/layers/family.py
```diff
@@ -162,6 +191,7 @@ async def get_pedigree(
         # pylint: disable=invalid-name
         replace_with_family_external_ids=False,
         empty_participant_value='',
+        include_header=True,
```
vivbak, line 194: Should this default to True here?  It looks like it's set as `False` by default here https://github.com/populationgenomics/sample-metadata/blob/d3b75a5cec9f54e606403d2ea9a7a66c5c15f196/api/routes/family.py#L51. 
That appears to be the only place that `get_pedigree` is called.

vivbak: I'm not too sure of the use case where `include_header` will be set to True as default as it seems it will always be provided when called?

vivbak (approved):
Hey Michael, `arbitrary_sm` is such a great idea :) Just one tiny comment.

db/python/layers/family.py
illusional: Nice catch, I've actually flipped this around. I think it makes sense for the route to default for include_header to True, but if you're requesting the pedigree around sample-metadata, you probably don't want the header.

PR #38, opened 2021-09-08
Add file map parser

illusional:
Notably, `reads` is going to `sequence.meta` (and I've updated the vcgs parser) - this is to better align with tob-wgs.

```tsv
Individual ID	Filenames
<sample-id>	<comma_R1.fastq.gz>,<separated_R1.fastq.gz>
```

Also supports:

```tsv
Individual ID	Filenames
<sample-id>	<sample_R1.fastq.gz>
<sample-id>	<sample_R2.fastq.gz>
```

### DIFF
--- a/scripts/parse_file_map.py
+++ b/scripts/parse_file_map.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 # pylint: disable=too-many-instance-attributes,too-many-locals,unused-argument,no-self-use,wrong-import-order,unused-argument
 import logging
 import os
--- a/scripts/parse_file_map.py
+++ b/scripts/parse_file_map.py
@@ -48,7 +48,7 @@ class SampleMapParser(GenericParser):
 
     def populate_filename_map(self, search_locations: List[str]):
         """
-        ParseFileMap uses search locations based on the filename,
+        FileMapParser uses search locations based on the filename,
         so let's prepopulate that filename_map from the search_locations!
         """
         self.filename_map = {}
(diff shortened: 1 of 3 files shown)


scripts/parse_file_map.py
```diff
@@ -0,0 +1,169 @@
...
+            delimeter=delimeter,
+            confirm=confirm,
+        )
+        self.search_locations = search_locations
+        self.filename_map = {}
+        self.populate_filename_map(self.search_locations)
+
+    def populate_filename_map(self, search_locations: List[str]):
+        """
+        ParseFileMap uses search locations based on the filename,
```
vladsaveliev, line 51: What is ParseFileMap?

scripts/parse_file_map.py
```diff
@@ -0,0 +1,169 @@
...
+    def get_sample_id(self, row: Dict[str, any]):
+        """Get external sample ID from row"""
+        external_id = row[Columns.INDIVIDUAL_ID]
+        return external_id
+
+    def get_sample_meta(self, sample_id: str, row: GroupedRow):
+        """Get sample-metadata from row"""
+        return {}
+
+    def get_sequence_meta(self, sample_id: str, row: GroupedRow):
```
vladsaveliev, line 85: I find it super helpful to specify the return type as well, helps to read code, and also can benefit from `mypy` static analysis :)

vladsaveliev (approved):
Awesome!!

scripts/parse_file_map.py
illusional: Renamed :)

scripts/parse_file_map.py
illusional: Thanks! I've added static types to them all.

PR #39, opened 2021-09-09
Add generic md5 checker

illusional:
For manual checking of md5 files in a directory.
Not very efficient, it'll load the list of files in the directory in memory and create a batch job for each pair of md5s.

Example: https://batch.hail.populationgenomics.org.au/batches/5026/jobs/1


vivbak (approved):
👏  Looks great.

PR #45, opened 2021-09-14
Improvements to analysis

illusional:
- Adds `meta`, `active` and `on_behalf_of` to analysis schema
    - `on_behalf_of` will be used in the future in the analysis runner to track the submitting user (vs the SM-authenticated request)
    - `active` is implemented as a follow-up to (#23) so hidden analysis can be created in test-projects - but is useful as a replacement to the airtable list of jobs run through the analysis-runner.
- Adds `unknown` to `AnalysisStatus`
- Adds `exome` to `SequenceType`

### DIFF
--- a/api/routes/analysis.py
+++ b/api/routes/analysis.py
@@ -160,23 +160,29 @@ async def get_latest_complete_analysis_for_type_post(
     return analysis
 
 
-@router.get(
-    '/{project}/{analysis_type}/latest-complete',
-    operation_id='getLatestCompleteAnalysisForType',
+@router.post(
+    '/{project}/{analysis_type}/{latest-complete}/for-samples',
+    operation_id='getLatestAnalysisForSamplesAndType',
 )
-async def get_latest_complete_analysis_for_type(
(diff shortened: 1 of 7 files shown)


api/routes/analysis.py
```diff
@@ -106,36 +119,52 @@ async def get_incomplete_analyses(
...
-    return results
+    return analysis
 
 
 @router.get(
     '/{project}/{analysis_type}/latest-complete',
     operation_id='getLatestCompleteAnalysisForType',
 )
-async def get_latest_complete_analysis_for_type(
+async def get_latest_complete_analysis_for_type_meta(
```
vladsaveliev, line 167: What's the difference between `get_latest_complete_analysis_for_type_meta` and `get_latest_complete_analysis_for_type`

vladsaveliev (approved):
Looks great!

api/routes/analysis.py
illusional: Nothing, this was incorrectly refactored (and I've renamed back).

PR #54, opened 2021-09-22
Script to populate a test project

vladsaveliev:
Usage example:

```
python scripts/create_test_subset.py --project perth-neuro -n 10
```

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -15,6 +15,7 @@ from sample_metadata import (
     NewSequence,
     NewSample,
     AnalysisModel,
+    SampleUpdateModel,
     exceptions,
 )
 
@@ -38,6 +39,7 @@ seqapi = SequenceApi()
     '-n',
     required=True,
     type=int,
+    default=10,
     help='Number of samples to subset',
 )
 def main(
@@ -49,6 +51,16 @@ def main(
     A new project with a prefix -test is created, and for any files in sample/meta,
     sequence/meta, or analysis/output a copy in the -test namespace is created.
     """
+    if n < 1:
+        raise click.BadParameter('Please specify n higher than 0')
+
+    if n >= 100:
+        resp = str(
+            input(f'You requested a subset of {n} samples. Please confirm (y): ')
+        )
+        if resp.lower() != 'y':
+            raise SystemExit()
+
     samples = sapi.get_samples(
         body_get_samples_by_criteria_api_v1_sample_post={
             'project_ids': [project],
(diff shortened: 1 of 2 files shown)


scripts/create_test_subset.py
```diff
@@ -0,0 +1,174 @@
...
+    seq_info_by_s_id = dict(zip(sample_ids, seq_infos))
+
+    target_project = project + '-test'
+
+    for s in samples:
+        logger.info(f'Processing sample {s["id"]}')
+
+        # processing sample entries
+        try:
+            new_s = sapi.get_sample_by_external_id(s['external_id'], target_project)
```
illusional, line 74: You could save yourself the loop by calling this upfront:

```python
sapi.get_sample_id_map_by_external([s['external_id'] for s in samples])
```

scripts/create_test_subset.py
```diff
@@ -0,0 +1,174 @@
...
+            )
+        else:
+            new_s_id = new_s.id
+
+        # processing sequence entries
+        seq_info = seq_info_by_s_id.get(s['id'])
+        if seq_info:
+            new_meta = _copy_files_in_dict(seq_info.get('meta'), project)
+            print(new_meta)
+            seqapi.create_new_sequence(
```
illusional, line 92: There's no try-catch here, I don't think you'd need to update if it already exists, but don't want to halt the script.

scripts/create_test_subset.py
```diff
@@ -0,0 +1,174 @@
...
+                    meta=new_meta,
+                    type=seq_info['type'],
+                    status=seq_info['status'],
+                )
+            )
+
+        # processing analysis entries
+        for a_type in ['cram', 'gvcf']:
+            try:
+                analysis = aapi.get_latest_analysis_for_samples_and_type(
```
illusional, line 104: You can request these for all sample Ids up front to save yourself the loop.

setup.py
```diff
@@ -33,6 +33,8 @@
         'urllib3 >= 1.25.3',
         'python-dateutil',
         'requests',
+        'click',
+        'google-cloud-storage',
```
illusional, line 37:
```suggestion
```

These requirements are for the install-able (through openapi) python module, not for the scripts. Traditionally scripts are run through the analysis-runner which already has these modules installed.

scripts/create_test_subset.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    samples = sapi.get_samples(
+        body_get_samples_by_criteria_api_v1_sample_post={
+            'project_ids': [project],
+            'active': True,
+        }
+    )
+    logger.info(f'Found {len(samples)} samples')
+    random.seed(42)  # for reproducibility
+    samples = random.sample(samples, n)
```
illusional, line 60: Should we cap the number of samples, or require a minimum?

illusional (approved):
Awesome script, going to be super useful! 
I've approved, if you can remove the requirements from `setup.py` that would be awesome.  There are a few non-critical changes to avoid lots of calls within loops, but not critical as this script won't be called that often.

PR #66, opened 2021-10-12
Test subset: add `--families` option

vladsaveliev:
Example:

```
create_test_subset.py --project acute-care --families 3
```


scripts/create_test_subset.py
```diff
@@ -182,11 +285,14 @@ def _copy_files_in_dict(d, dataset: str):
     return d
 
 
-def _pretty(samples: List[Dict]) -> str:
+def _pretty_format_samples(samples: List[Dict]) -> str:
     return ', '.join('/'.join([s['id'], s['external_id']]) for s in samples)
```
illusional, line 289:
```suggestion
    return ', '.join(f"{s['id']}/{s['external_id']}" for s in samples)
```
String format the inner, so you don't have issues with `None` / non-str types (though that shouldn't happen)

illusional (approved):
Awesome addition!

PR #70, opened 2021-10-25
Add kccg update endpoint

vivbak:


vivbak:
Handled :) 
https://github.com/populationgenomics/sample-metadata/pull/247
https://github.com/populationgenomics/sample-metadata/pull/227

PR #78, opened 2021-11-09
Add web app for documentation

illusional:
- Configures the openapi-generator to use templates
- Creates React app to serve this markdown documentation (including some URL rewriting)
- Adds a swagger page into the react app at `/swagger`.
- Adds a build step to the github action workflow to build the UI (which as a postbuild copies artifacts into the python api for distribution.
- Adds a static file mount to the FastAPI app
  - + rewrite all other URLs not starting with `/api` to return the index.html page, so a direct link to the SPA works


danielreti (commented):
Only very very minor notes/questions for Documentation.tsx:
Is there a reason you use props.href on line 17 and then assign it  to the link variable on line 21 and then use that?
Console log on line 34
What is the purpose of a useEffect with no dependencies?
Require on line 44 seems hacky but also makes sense and I'm not sure what the alternative is.

web/src/Documentation.tsx
```diff
@@ -0,0 +1,104 @@
...
+    articleId?: string
+}
+
+function RouterLink(props: any) {
+    if (props.href.match(/^(https?:)?\/\//))
+        return <a href={props.href}>{props.children}</a>
+
+
+    let link: string = props.href
+    if (!link.startsWith('/')) { link = '/documentation/' + link; }
```
illusional, line 22: Hey @danielreti, I'll reply to you in the inline comments :)
Just setting `link` to a variable here, as I want to rewrite if it's a relative link - as the relative links generated by the markdown don't work with this format.

web/src/Documentation.tsx
```diff
@@ -0,0 +1,104 @@
...
+}
+
+const DocumentationArticle: React.FunctionComponent<IDocumentationArticleProps> = (props: IDocumentationArticleProps) => {
+
+    const [text, setText] = React.useState<string | null>(null)
+
+    const match = useParams()
+
+    useEffect(() => {
+        console.log('Loading data')
```
illusional, line 34: The `useEffect` here with no dependencies should act like `componentDidMount` or `componentDidUpdate` functions, allowing me to load data on first load, or reload if the props change. An optimisation here could be to only reload if the articleId changes, but not required at the moment,

illusional, line 34: @danielreti 
The `useEffect` here with no dependencies should act like `componentDidMount` or `componentDidUpdate` functions, allowing me to load data on first load, or reload if the props change. An optimisation here could be to only reload if the articleId changes, but not required at the moment,

danielreti (approved):

PR #82, opened 2021-11-11
Fix tob parser

### LINKED ISSUE #81: TOB parser fails when ingesting new samples
When ingesting samples that don't exist already in the SM DB it will fail on a key error. https://batch.hail.populationgenomics.org.au/batches/6472/jobs/1

Current workaround is to simply run it again.

vivbak:
Fixes #81

### DIFF
--- a/sample_metadata/parser/generic_parser.py
+++ b/sample_metadata/parser/generic_parser.py
@@ -320,7 +320,7 @@ Updating {len(sequences_to_update)} sequences"""
                 project=self.sample_metadata_project, new_sample=new_sample
             )
             added_ext_sample_to_internal_id[new_sample.external_id] = sample_id
-
+existing_external_id_to_cpgid[new_sample.external_id] = sample_id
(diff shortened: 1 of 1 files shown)


sample_metadata/parser/generic_parser.py
```diff
@@ -321,14 +321,19 @@ def parse_manifest(
             )
             added_ext_sample_to_internal_id[new_sample.external_id] = sample_id
```
illusional, line 323: I think a simpler solution might be to include the added sample id in the list, eg:

```suggestion
existing_external_id_to_cpgid[new_sample.external_id] = sample_id
```

Otherwise we'd probably want `allow_missing=False` in the sapi call.

illusional (changes requested):

sample_metadata/parser/generic_parser.py
vivbak: great point, thanks!

illusional (approved):

PR #86, opened 2021-11-24
Add nagim parser

vladsaveliev:
A couple of additions into GenericParser:

1. `verbose=False` option to make the output more clear for big amounts of samples,
2. `skip_checking_gcs_objects=True` to skip all bucket object existence and size checks and md5 checks to make it work faster,
3. Error handling when manifest file has no data.

Added `parse_nagim.py`. The script has 2 commands: `transfer` and `parse`.

```
python scripts/parse_nagim.py transfer \
    --tmp-dir ~/tmp \
    --transfer-crams \
    --transfer-gvcfs
```

Transfers the CRAM and GVCF files, along with corresponding indices from a Terra workspace to the GCP upload bucket. Currently only works under a personal account, which works well for a test dataset, but might be harder for the US workspace on the full data. Ideally the Terra bucket can be shared with a service account, so the transfer
command can be submitted to Hail Batch (with `--use-batch`), however I haven't yet figured out how to shared with a service account.

```
python scripts/parse_nagim.py parse \
    nagim.samples_list.tsv \
    --tmp-dir ~/tmp \
    --skip-checking-objects \
    --confirm
```

Assuming CRAM and GVCFs are transferred, populates the sample metadata DB objects. The first parameter is a path to a TSV with 2 columns: sample IDs used in the NAGIM run, and project ID (in a format defined below in `PROJECT_ID_MAP`), which I got from Shyam.

### DIFF
--- a/sample_metadata/parser/generic_parser.py
+++ b/sample_metadata/parser/generic_parser.py
@@ -206,8 +206,7 @@ class GenericParser:
             sample_map[sample_id].append(row)
 
         if len(sample_map) == 0:
-            logger.error(f'No sample information read from the manifest')
-            return dict()
+            raise ValueError(f'The manifest file contains no records')
 
         # now we can start adding!!
         sapi = SampleApi()
--- a/sample_metadata/parser/generic_parser.py
+++ b/sample_metadata/parser/generic_parser.py
@@ -43,10 +43,14 @@ class GenericParser:
         default_sequence_type='wgs',
         default_sample_type='blood',
         confirm=False,
+        skip_checking_gcs_objects=False,
+        verbose=True,
     ):
 
         self.path_prefix = path_prefix
         self.confirm = confirm
+        self.skip_checking_gcs_objects = skip_checking_gcs_objects
+        self.verbose = verbose
 
         if not sample_metadata_project:
             raise ValueError('sample-metadata project is required')
@@ -170,7 +174,7 @@ class GenericParser:
         """Get sequence-metadata from row"""
 
     def get_qc_meta(self, sample_id: str, row: GroupedRow) -> Optional[Dict[str, any]]:
-        """Get qc-meta from row, creates a QC object"""
+        """Get qc-meta from row, creates a Analysis object of type QC"""
         return None
 
     def get_sample_type(self, sample_id: str, row: GroupedRow) -> str:
@@ -185,11 +189,13 @@ class GenericParser:
     def get_sequence_status(self, sample_id: str, row: GroupedRow) -> str:
         """Get sequence status from row"""
 
-    def parse_manifest(
+    def parse_manifest(  # pylint: disable=too-many-branches
         self, file_pointer, delimiter=',', dry_run=False
-    ):  # pylint: disable=too-many-branches
+    ) -> Dict[str, str]:
         """
         Parse manifest from iterable (file pointer / String.IO)
+
+        Returns a dict mapping external sample ID to CPG sample ID
         """
         # a sample has many rows
         sample_map = defaultdict(list)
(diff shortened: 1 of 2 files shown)


sample_metadata/parser/generic_parser.py
```diff
@@ -199,6 +205,10 @@ def parse_manifest(
             sample_id = self.get_sample_id(row)
             sample_map[sample_id].append(row)
 
+        if len(sample_map) == 0:
+            logger.error(f'No sample information read from the manifest')
```
illusional, line 209: I think this should raise an exception.

scripts/parse_nagim.py
```diff
@@ -0,0 +1,467 @@
...
+SRC_BUCKETS_TEST = [
+    # Australian Terra workspace
+    'gs://fc-7d762f69-bb45-48df-901b-b3bcec656ee0/2232b739-5183-4935-bb84-452a631c31ea',
+    # The US Terra workspace
+    'gs://fc-bda68b2d-bed3-495f-a63c-29477968feff/1a9237ff-2e6e-4444-b67d-bd2715b8a156',
+]
+
+SRC_BUCKETS_MAIN = []
+
+assert all(b.startswith('gs://fc-') for b in SRC_BUCKETS_MAIN + SRC_BUCKETS_TEST)
```
illusional, line 59: Could use a comment to say all terra buckets start with fc?

scripts/parse_nagim.py
```diff
@@ -0,0 +1,467 @@
...
+        with open(sample_tsv_file) as f:
+            parser.parse_manifest(f, dry_run=dry_run)
+
+
+def _fix_ids(
+    samples_from_found_files: List[Sample],
+    sample_to_project_path: str,
+    prod: bool,
+) -> List[Sample]:
+    sapi = SampleApi()
```
illusional, line 242: Docstring please

scripts/parse_nagim.py
```diff
@@ -0,0 +1,467 @@
...
+            f' | gsutil -m cp -I {upload_bucket}/'
+        )
+        if hbatch:
+            _add_batch_job(hbatch, cmd, f'Transfer GVCFs from {bucket}')
+        else:
+            _call(cmd)
+
+
+def _add_batch_job(hbatch, cmd, job_name):
+    j = hbatch.new_job(job_name)
```
illusional, line 323: Why default 32 cores?

scripts/parse_nagim.py
```diff
@@ -0,0 +1,467 @@
...
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def get_sample_id(self, row: Dict[str, any]) -> str:
+        return row['ext_id']
+
+    def get_sequence_meta(self, sample_id: str, row: GroupedRow) -> Dict[str, any]:
+        sequence_meta = dict()
+        if 'gvcf' in row and row['gvcf'] and row['gvcf'] != '-':
```
illusional, line 357: How does row['gvcf'] be set to "-"?

scripts/parse_nagim.py
```diff
@@ -0,0 +1,467 @@
...
+        )
+    return samples
+
+
+cli.add_command(transfer)
+cli.add_command(parse)
+
+if __name__ == '__main__':
+    # pylint: disable=no-value-for-parameter
+    cli(obj={})  # pylint: disable=unexpected-keyword-arg
```
illusional, line 467: Isn't this generally:
```suggestion
    cli()  # pylint: disable=unexpected-keyword-arg
```

illusional (approved):
A few constructive comments, but generally looks good

scripts/parse_nagim.py
vladsaveliev: Remaining artefact from sharing context between the commands

scripts/parse_nagim.py
vladsaveliev: The `-m` option for gsutil enables multi-threaded copy, I think the more cores the faster?

scripts/parse_nagim.py
vladsaveliev: Didn't want to duplicate the comment where the function is called, but will add

vladsaveliev: Actually, split it into 2 methods, one to fix sample IDs, another to set project IDs.

PR #92, opened 2021-11-30
Add family + sample methods

illusional:
Unrelated to larger version update change

Adds:
- Import families (from tsv)
- Get families
- Update family
- Get history of sample

### DIFF
--- a/db/python/tables/sample.py
+++ b/db/python/tables/sample.py
@@ -299,3 +299,23 @@ WHERE participant_id IS NULL AND project = :project
             _query, {'project': project or self.project}
         )
         return {row['id']: row['external_id'] for row in rows}
+
+    async def get_history_of_sample(self, id_: int):
+        """Get all versions (history) of a sample"""
+        keys = [
+            'id',
+            'external_id',
+            'participant_id',
+            'meta',
+            'active+1 as active',
+            'type',
+            'project',
+            'author',
+        ]
+        keys_str = ', '.join(keys)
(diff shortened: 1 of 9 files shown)


db/python/tables/sample.py
```diff
@@ -299,3 +299,23 @@ async def get_sample_with_missing_participants_by_internal_id(
...
         return {row['id']: row['external_id'] for row in rows}
+
+    async def get_history_of_sample(self, id_: int):
+        """Get all versions (history) of a sample"""
+        keys = [
+            'id',
+            'external_id',
+            'participant_id',
+            'meta',
+            'active+1 as active',
```
vladsaveliev, line 310: Why +1?

illusional: Forces it to an `int`, otherwise mariadb returns like `x01`: https://stackoverflow.com/questions/19256155/python-mysqldb-returning-x01-for-bit-values.

vladsaveliev: Cool, thanks!

vladsaveliev:
Maybe I'm misunderstanding, should Family also contain pointers to mom and dad?

illusional:
A family is the container, it doesn't really represent much information (hence why we haven't needed yet). The import pedigree method has just created the family (without a coded phenotype), but now I've got a use case where I have to update the family IDs.

vladsaveliev (approved):
Looks great!

PR #93, opened 2021-12-02
Nagim parser: set analysis meta

vladsaveliev:
Mostly updating the analysis model:
```python
AnalysisModel(
    sample_ids=['<none>'],
    type=AnalysisType(atype),
    status=AnalysisStatus('completed'),
    output=file_path,
    meta={
        # To distinguish TOB processed on Terra as part from NAGIM
        # from those processed at the KCCG:
        'source': 'nagim',
        # Indicating that files need to be renamed to use CPG IDs,
        # and moved from -upload to -test/-main. (For gvcf, also
        # need to reblock):
        'staging': True,
        'project': row.get('project'),
    },
)
```

(Skipped updated imports in `sample_metadata/parser/generic_parser.py`, to avoid conflicts with your upcoming PR :)

### DIFF
--- a/scripts/parse_nagim.py
+++ b/scripts/parse_nagim.py
@@ -66,10 +66,15 @@ import click
 import pandas as pd
 
 from cpg_pipes.pipeline import setup_batch
-
-from sample_metadata.models.analysis_model import AnalysisModel
+from sample_metadata.models import (
+    SequenceStatus,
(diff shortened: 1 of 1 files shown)


illusional (approved):
Just note the new sample metadata version hasn't been deployed to analysis runner yet

PR #96, opened 2021-12-08
Release sample-metadata

illusional:
This noteworthy 4.0.0 release introduces backwards incompatible changes to the installable python API.
Additionally, it brought mypy as a pre-commit and linting hook to reduce likelihood of errors.


vladsaveliev (approved):
❤️  for mypy!

PR #97, opened 2021-12-08
Add ability to authenticate with IAP

illusional:
- Adds support for authenticating using the IAP jwt (prioritised over bearer token to avoid forged JWTs)
- Add cryptography package for verifying the OAUTH audience


lgruen (approved):
Nice catch! Just wondering whether we want to move some of this to `cpg-utils` and potentially make the documentation / name of `email_from_id_token` clearer to avoid similar issues in the future?

illusional:
Yeah I think that's a great idea, I'll merge this for now to fix the issue, and move code to cpg-utils :)

PR #98, opened 2021-12-08
Nagim parser: MultiQC, refactor, add main

vladsaveliev:
* Use MultiQC to parse QC metrics
* Add production buckets
* Refactor to make more extensible to potentially reuse for other Terra projects

* Minor GenericParser updates: 
   * add default sequence uploaded status, 
   * add sequences only if get_sequence_meta returns,
   * change type of sequences_to_add as it has one sequence per sample

### DIFF
--- a/scripts/parse_nagim.py
+++ b/scripts/parse_nagim.py
@@ -459,19 +459,17 @@ def parse(
     # have them in the SMDB and fixing the external IDs.
     _fix_sample_ids(samples)
 
-    multiqc_json_fpath = None
+    multiqc_html_path = join(
+        f'gs://cpg-{NAGIM_PROJ_ID}-{NAMESPACE}-web/qc/multiqc.html'
+    )
+    multiqc_json_path = join(
+        f'gs://cpg-{NAGIM_PROJ_ID}-{NAMESPACE}-analysis/qc/multiqc_data.json'
+    )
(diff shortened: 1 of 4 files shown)


scripts/parse_nagim.py
```diff
@@ -403,21 +692,38 @@ def _add_batch_job(hbatch, cmd, job_name):
...
 def _call(cmd):
-    print(cmd)
+    """
+    Call subprocess command locally.
+    """
+    logger.info(cmd)
     subprocess.run(cmd, shell=True, check=True)
 
 
+def _batch_or_call(cmd, hbatch=None, job_name=None):
```
illusional, line 702: I don't really like this, seems a little dangerous to accidentally get it to run. I'd probably say make the user actively decide to _add_batch_job or _call, and have the logic explicitly above.

illusional (changes requested):
Looks great! I have pretty strong reservations about `_batch_or_call` which is why I've requested changes, happy to jump on a call to chat more :)

scripts/parse_nagim.py
vladsaveliev: But this function is not user exposed, it's just a helper for this script.

Removed it anyway, calling explicitly `_add_batch_job()` to transfer

PR #108, opened 2022-01-05
Update packages + fix db diagram

illusional:


vivbak:
I'm not sure if you're looking to update the db diagram in general, if so, should the participant table also include reported_sex, reported_gender, karyotype and meta? 
Screen Shot 2022-01-05 at 1 26 45 pm
Screen Shot 2022-01-05 at 1 26 54 pm

illusional:
I wasn't intending to yet, it would be good to but I'm just hoping to silence some of the warnings I get from GH.

vivbak (approved):

PR #110, opened 2022-01-06
Bump version: Create update single participant endpoint

vivbak:


api/routes/participant.py
```diff
@@ -148,3 +158,25 @@ async def get_external_participant_id_to_internal_sample_id_export(
         media_type=export_type.get_mime_type(),
         headers={'Content-Disposition': f'filename={filename}'},
     )
+
+
+@router.post(
+    '/{project}/update-single-participant', operation_id='updateSingleParticipant'
```
illusional, line 164: I think we should infer the `{project}` in the layer just before the write - otherwise this would also allow someone to specify any project they have write access to, even if the participant didn't belong to that project. 

Could we also put the participant ID in the URL, maybe something like

```suggestion
    '/{participant_id}/update', operation_id='updateParticipant'
```

api/routes/participant.py
```diff
@@ -148,3 +158,25 @@ async def get_external_participant_id_to_internal_sample_id_export(
...
+)
+async def update_single_participant(
+    participant_id: int,
+    participant: ParticipantUpdateModel,
+    connection: Connection = get_project_write_connection,
```
illusional, line 169:
```suggestion
    connection: Connection = get_projectless_db_connection,
```

db/python/layers/participant.py
```diff
@@ -621,3 +621,17 @@ async def update_participants(
...
+
+    async def update_single_participant(
+        self,
+        participant_id: int,
+        reported_sex: int = None,
+        reported_gender: str = None,
+        karyotype: str = None,
+        meta: Dict = None,
+        author=None,
+    ):
```
illusional, line 633: By removing the project earlier, can you add a `check_project_id=True` + the actual check before we do the update participant.

db/python/tables/participant.py
```diff
@@ -94,6 +94,42 @@ async def update_participants(
...
+        if reported_gender:
+            updaters.append('reported_gender = :reported_gender')
+            fields['reported_gender'] = reported_gender
+
+        if karyotype:
+            updaters.append('karyotype = :karyotype')
+            fields['karyotype'] = karyotype
+
+        if meta:
+            updaters.append('meta = JSON_MERGE_PATCH(COALESCE(meta, "{}"), :meta)')
```
illusional, line 125: Awesome!

illusional (changes requested):
Looking really great! I left a few comments about checking the project id :)

vivbak:
Thanks very much for the review @illusional! Hopefully i've interpreted your suggestions correctly, but let me know if not :)

illusional (approved):
Looks great! Can you add a `bumpversion minor` commit to this to ensure the new API gets released? Thanks!

PR #113, opened 2022-01-11
Parse participant metadata TOB-WGS

vivbak:
Quick script to leverage new update participant endpoint, and parse the master CSV file.


illusional (changes requested):
Great first pass! I've left some feedback about removing the global variable, and performing the get_id_map of participants once.

illusional (approved):
I've suggested changes to getting a file's contents on GCS, looks great!

PR #114, opened 2022-01-12
NAGIM parser final changes

vladsaveliev:
Some remaining parser changes from before-the-holidays:

* Add `move_uploads` command that moves gvcf/cram files to final destinations and updates analysis,
* NAGIM test subset: fix external ids, explicitly specify how many samples from each dataset we want;
* Add ancestry and sex metadata for the projects where it's known (1kg and hgdp);
* Parse QC metrics from multiqc json;
* Add more handy logging into generic_parser (when "verbose=True").


vladsaveliev:
CI error is weird 🤔

scripts/parse_nagim.py
```diff
@@ -793,21 +1233,103 @@ def get_sequence_meta(self, sample_id: str, row: GroupedRow) -> Dict[str, Any]:
...
+        new_bucket = SOURCES['GVCF'].get_staging_bucket(stack, namespace)
+        for s in [s for s in samples if s.project_id == proj]:
+            if s.gvcf:
+                j = _add_batch_job(hbatch, f'{smdb_proj}/{s.cpg_id}: Move GVCF uploads')
+                s.gvcf.moved_path = join(new_bucket, f'{s.cpg_id}.g.vcf.gz')
+                s.tbi.moved_path = s.gvcf.moved_path + '.tbi'
+                _move(j, s.gvcf)
+                _move(j, s.tbi)
+
+    hbatch.run(wait=True, dry_run=dry_run)
```
illusional, line 1325: Why `wait=true`, this could cause re-runs.

scripts/parse_nagim.py
```diff
@@ -308,18 +639,20 @@ def transfer(
...
         )
     else:
         hbatch = None
 
     for source in SOURCES_TO_PROCESS:
-        SOURCES[source].transfer(hbatch)
+        SOURCES[source].transfer_to_upload(hbatch, namespace)
 
     if use_batch:
         hbatch.run(wait=True, dry_run=dry_run)
```
illusional, line 658: Just again, I think wait=True is a little dangerous when directly running the workflow.

illusional (approved):
Lots of changes, I think I have the gist of what's happening, but it's hard for me completely understand. Nothing obvious stands out to me, I've left a couple clarifying questions that I don't think are blockers but would appreciate a couple clarifying docstrings. 

With my review, I think important to note it's a _one off_ style script.

scripts/parse_nagim.py
vladsaveliev: It's okay if the main process is interrupted, because all what the following code is doing is just the consistency checks. So if you kill the main process, you can manually watch the batch process finish, and then run the `parse` command that would start with those checks anyway

scripts/parse_nagim.py
```diff
@@ -146,142 +222,397 @@ class Source:
...
+class Sample:  # pylint: disable=too-many-instance-attributes
     """
     Represent a parsed sample, so we can check that all required files for
     a sample exist, and also populate and fix sample IDs.
     """
 
     nagim_id: str
     cpg_id: Optional[str] = None
+    test_cpg_id: Optional[str] = None
+    main_cpg_id: Optional[str] = None
```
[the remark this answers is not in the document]
vladsaveliev: For a test and main subset, we would have different internal IDs for the same external ID. I wanted to link the test internal ID with the main internal ID in this data structure so I could easily determine which files to move from main to test and how exactly to rename them.

scripts/parse_nagim.py
vladsaveliev: What do you mean? Why would it cause re-runs in contrast with wait=False?

Actually `wait=True` is not really necessary here, because it's the last line in the command

vladsaveliev:
> With my review, I think important to note it's a one off style script.

Even though it's a one-off script, I think we might want to refactor it or reuse parts of it to create something production-looking like. I just don't fully understand yet what would a more generalised version look like. With more projects coming to parse in the future, will revisit this.

PR #119, opened 2022-01-13
Fix PBMC sample participants 

vivbak:
https://github.com/populationgenomics/sample-metadata/pull/113 Assumed a 1:1 relationship between samples and participants for the entire TOB-WGS project. However, this is not the case for the PBMC samples. This script modifies the PBMC samples' participant data appropriately.


vivbak:
See https://github.com/populationgenomics/sample-metadata/pull/120

PR #120, opened 2022-01-13
Fix PBMC sample participants

vivbak:
#113 Assumed a 1:1 relationship between samples and participants for the entire TOB-WGS project. However, this is not the case for the PBMC samples. This script modifies the PBMC samples' participant data appropriately.


illusional (approved):
Looks great! Thanks for fixing this up!

Issue #107, opened 2022-01-04
SMAPI; get_pedigree endpoint returns None


MattWellie:
minimal example:

```
from sample_metadata.apis import FamilyApi
response = FamilyApi().get_pedigree(project='acute-care')
print(response)
```
> None

alternative queries work fine, e.g. 

`FamilyApi().get_families(project='acute-care')`. Querying 'manually' using the requests package is also fine.

Cause: 
This endpoint is returning a filestream, and parsing the response via requests requires decoding the returned content as a string. This differs from all other endpoints which are happily dumping JSON. At some point the client might be dropping this content, or isn't checking for 'content' within the response value.

Request:
- Please modify the client such that the text returned by default on this endpoint is handled correctly
- An alternative method, e.g. 'get_families_json' is possible, which will return a JSON-formatted version of the table contents

Ext: 
- alter the API so that the "Accept" parameter in the header determines how this data is presented (currently `*/*` for text)

PR #128, opened 2022-01-25
Resolves SMAPI; get_pedigree endpoint returns None #107

violetbrina:
PR to fix the issue where the SM api returns none. Now response type json is assumed (both for swagger and SM python api) and the response_type argument can be specified to have a tsv or csv returned.

Resolves SMAPI; get_pedigree endpoint returns None #107


api/routes/family.py
```diff
@@ -85,23 +94,35 @@ async def get_pedigree(
...
+            iter(output.getvalue()),
+            media_type=f'text/{response_type}',
+            headers={'Content-Disposition': f'filename={basefn}.ped'},
+        )
+    else:
+        header = pedigree_rows.pop(0)
+        header = [re.sub(r'^[^a-zA-Z0-9]+', '', x) for x in header]
+        data = [dict(zip(header, x)) for x in pedigree_rows]
+        encoded = jsonable_encoder(data)
+        return JSONResponse(content=encoded)
```
illusional, line 125: By default, a response will automatically be converted to json, so this can just be:

```suggestion
        data = [dict(zip(header, x)) for x in pedigree_rows]
        return data
```

api/routes/family.py
```diff
@@ -85,23 +94,35 @@ async def get_pedigree(
...
+            basefn += '-'.join(str(fm) for fm in internal_family_ids)
+
+        return StreamingResponse(
+            iter(output.getvalue()),
+            media_type=f'text/{response_type}',
+            headers={'Content-Disposition': f'filename={basefn}.ped'},
+        )
+    else:
+        header = pedigree_rows.pop(0)
+        header = [re.sub(r'^[^a-zA-Z0-9]+', '', x) for x in header]
```
illusional, line 122: This is great! On reflection, could we convert anything to camelcase, eg so: `"Participant ID"` would go to `participant_id`, I imagine something like `.lower().replace(' ', '_')` before the `re.sub` would be perfect.

api/routes/family.py
```diff
@@ -60,14 +67,15 @@ async def import_pedigree(
...
 
 @router.get(
-    '/{project}/pedigree', operation_id='getPedigree', response_class=StreamingResponse
+    '/{project}/pedigree', operation_id='getPedigree',
 )
 async def get_pedigree(
     internal_family_ids: List[int] = Query(None),
+    response_type: ContentType = ContentType.JSON,
     replace_with_participant_external_ids: bool = False,
     replace_with_family_external_ids: bool = False,
```
illusional, line 76: While we're changing this endpoint, could we set these to `True`, I feel like that's the more common use case, and I'm pretty confident that we don't have clients using this endpoint (because it didn't work in the Python API earlier).

```suggestion
    replace_with_participant_external_ids: bool = True,
    replace_with_family_external_ids: bool = True,
```

illusional (changes requested):
This is an awesome change, really excited to get this in!
I've left a couple of small points to feedback, we also have linting on this library (you can enable it with `pre-commit`. You can see the feedback in the action window next to the commit (on the PR page).

It's not introduced in the documentation of this repo (another thing to add!), but you can install the pre-commit hooks with:

```shell
pre-commit install --install-hooks
```

The pre-commit hooks do run on the changes when you commit locally, but you can test run everything on every file with:

```shell
pre-commit run --all-files
```

Edit: Could you also add a `bumpversion minor` commit to this PR, that should bump the sample_metadata version to 4.3.0. You might need to `pip install bumpversion` to make it work :)

api/routes/family.py
violetbrina: Yup done 👍

api/routes/family.py
violetbrina: Done 😉

api/routes/family.py
violetbrina: Done!

illusional (approved):
Nice, this is an awesome fix!

PR #117, opened 2022-01-12
Add project summary table

illusional:
- Adds a project summary table that allows you to select a project, and displays a paged list of samples in a table.

Technical:
- Adds automatic typescript binding generation from openapi
- Adds ability to store global context in react app
- Adds `reactstrap` UI library
- Adds local development option for default authorisation (without bearer token, useful for development of front-end)

PR #132, opened 2022-02-03
Adds project summary grid backend + typescript generator

illusional:
Partial changes from #117.

**Adds a project summary table that allows you to select a project, and displays a paged list of samples in a table.**

Technical:

- Adds automatic typescript binding generation from openapi
- Adds local development option for default authorisation (without bearer token, useful for development of front-end)
- Adds PAGED web route for getting a project summary

### DIFF
--- a/api/routes/web.py
+++ b/api/routes/web.py
@@ -1,0 +1,95 @@
+"""
+Web routes
+"""
+from typing import Optional, List
+
+from fastapi import APIRouter, Request
+from pydantic import BaseModel
+
+from models.models.sample import sample_id_format
+from db.python.layers.web import WebLayer, NestedParticipant
+
+from api.utils.db import (
+    get_project_readonly_connection,
+    Connection,
+)
+
+
+class PagingLinks(BaseModel):
+    """Model for PAGING"""
+
+    self: str
+    next: Optional[str]
+    token: Optional[str]
+
+
+class ProjectSummaryResponse(BaseModel):
+    """Response for the project summary"""
+
+    participants: List[NestedParticipant]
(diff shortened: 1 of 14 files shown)


api/utils/export.py
```diff
@@ -0,0 +1,26 @@
+from enum import Enum
+
+
+class ExportType(Enum):
```
vladsaveliev, line 4: But is it used anywhere? Sorry if I'm blind!

vladsaveliev: Can't see it used in the larger PR either

api/routes/web.py
```diff
@@ -0,0 +1,95 @@
...
+    """Response for the project summary"""
+
+    participants: List[NestedParticipant]
+    total_samples: int
+    sample_keys: List[str]
+    sequence_keys: List[str]
+
+    links: Optional[PagingLinks]
+
+    class Config:
```
vladsaveliev, line 36: Where is the Config class used?

vladsaveliev (approved):
Looks great! Though I haven't tried it - would love to at some point (both as a user and as a dev)

Thanks a lot for building the UI for the SM!

api/routes/web.py
illusional: It's used by pydantic to configure the serialization.

api/utils/export.py
illusional: I think this came from a bad merge (when I split the PRs), this was refactored into `extensions.py` which covered more use cases.

illusional: I've removed the file now, thanks!

Issue #130, opened 2022-02-01
Sample metadata parsers can fail midway through (and are slow)


illusional:
The sample-metadata parsing scripts prepares the samples, sequences and metadata ready for ingestion, and then hit the sample-metadata server for each sample, sequence and analysis to insert. This can be very slow, and could break if a request fails, requiring a retry, or worse a silent failure.

It would be great to have one endpoint, to send a formatted JSON to import the samples, sequences and analysis objects to create / update. The request will know whether it needs to be updated / created, so you could ask for an optional ID, and insert if it doesn't exist.

It could be formatted like:

- `samples` (null `.id` means insert)
    - list of sequences that belong to the sample (null `.id` means insert)
    - list of analysis objects to create with only one sample_id, shouldn't include if we're not creating. Though @vladsaveliev, does the parsing scripts correctly handle if the relevant analysis already exists?)

Preferably in one SQL transaction so rollbacks are handled correctly.

PR #134, opened 2022-02-07
syan/batch upload

violetbrina:
PR in progress for batch upload mode. Endpoint currently working, testing in progress.

TODO:
- re-write the generic parser to use the batch endpoint

Closing Sample metadata parsers can fail midway through (and are slow) #130


api/routes/sample.py
```diff
@@ -60,6 +88,46 @@ async def create_new_sample(
...
+    async with connection.connection.transaction():
+        # Create or update samples
+        upserts = [upsert_sample(s) for s in samples.samples]
+        runs = await asyncio.gather(*upserts)
+        sids = [sample_id_format(s) for s in runs]
+
+        # Upsert all sequences with paired sids
+        sequences = zip(sids, [x.sequences for x in samples.samples])
+        seqs = [upsert_sequences(sid, seqs) for sid, seqs in sequences]
+        results = await asyncio.gather(*seqs)
```
illusional, line 124: Does this bubble up errors correctly? I can't remember if an exception in a called function causes all other promises to get cancelled.

api/routes/sample.py
```diff
@@ -60,6 +88,46 @@ async def create_new_sample(
...
+
+        # Upsert sequences
+        print(results)
+        return dict(zip(sids, results))
```
illusional, line 128: I think it would make more sense to return `{internal_sample_id: external_sample_id}`.

api/routes/sample.py
```diff
@@ -60,6 +88,46 @@ async def create_new_sample(
...
+)
+async def batch_upsert_samples(
+    samples: SampleUpsertBatchModel,
+    connection: Connection = get_project_write_connection,
+) -> Dict[str, Any]:
+    """Upserts a list of samples with sequences, and returns the list of internal sample IDs"""
+
+    def upsert_sample(sample):
+        if not sample.id:
+            return create_new_sample(sample, connection)
```
illusional, line 102: It looks like `create_new_sample` creates a transaction within itself. Do the two transactions sort of cancel out, (tbh: I don't hink create_new_sample doesn't need to create it's own transaction),

illusional (commented):
Looks really great! I've left a few small comments, but great approach!

api/routes/sample.py
violetbrina: Fixed!

api/routes/sample.py
violetbrina: Ah good idea. Let me reformat that output

api/routes/sample.py
violetbrina: Double checked the error handling. All good using async/await

violetbrina:
Generic parser now uses batch endpoint for samples and sequences (not analyses).
Example return value for a two sample insert/update.

`{'808': {'sample_id': 'CPGLCL8088', 'sequences': [1044]}, '809': {'sample_id': 'CPGLCL8096', 'sequences': [1043]}}`

sm/generate_data.py
```diff
@@ -0,0 +1,126 @@
+from pprint import pprint
```
illusional, line 1: Thanks for including this in this PR, I lost it through splitting and combining other PRs. Could we put this in maybe `test/data/` with a small README?

illusional (changes requested):
This is looking great! A few things that are highlighted below:

- The layers / db levels should only know that sample ID is an integer, that might a be annoying for how models are defined.
- Moving the router logic into a layer somewhere (though you might have to do a transform of sample IDs on the way into the layer. Keeps the routers very lightweight.
- For the parser, I might not be understanding what the `upsert` flag is, but I think it's worth stripping out all the old insert / update logic. Ie: stripping out `sample_to_add, sample_to_update, sequence_to_add, sequence_to_update` in favour of `sample_to_upsert`. That will mean rewriting the warning messages to determine how many inserts, updates for samples / sequences from the list of upsert models, but very worthy.

api/main.py
```diff
@@ -0,0 +1,122 @@
+import os
```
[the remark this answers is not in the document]
violetbrina: Oh yeah, I had to move it to a separate file to get the auto reload working. Not sure why but the uvicorn command to run the server needs to be called with a string not an import for it to work and it didn't work in the file it was in. The reason I moved it was so that the run command from the command line would still be the same (running server.py) but the behaviour would include auto reloading.

I'm happy to refactor another way, but this seemed to work as an initial solution.

api/routes/sample.py
```diff
@@ -60,6 +75,37 @@ async def create_new_sample(
...
+        sequences = zip(sids, [x.sequences for x in samples.samples])
+        seqs = [await seqt.upsert_sequences(sid, seqs) for sid, seqs in sequences]
+
+        # Format and return response
+        seqs_w_ids = dict(zip(iids, seqs))
+        results = [
+            {'sample_id': sample_id_format(k), 'sequences': v}
+            for k, v in seqs_w_ids.items()
+        ]
+        return dict(zip(iids, results))
```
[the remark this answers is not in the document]
violetbrina: `SampleLayer`? Happy to move it, just not sure what you mean.

sample_metadata/parser/generic_parser.py
```diff
@@ -288,6 +291,7 @@ async def process_group(
         external_sample_id: str,
         cpg_sample_id: Optional[str],
         sequence_id: Optional[str],
+        upsert: Optional[bool]
```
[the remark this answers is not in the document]
violetbrina: Yeah I wasn't sure if that function was used elsewhere or not. Just didn't want to break something! But if this is it's only use case I'll remove the old code.

sm/generate_data.py
violetbrina: Sure!

api/routes/sample.py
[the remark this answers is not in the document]
illusional: Creating a function called `upsert_batch` in SampleLayer that takes the `SampleBatchUpsert` and creates the transaction and calls the upsert_samples and upsert_sequences. So if in future someone produces a SampleBatchUpsert model, they can call that method directly, rather than this logic in the route.

sm/generate_data.py
violetbrina: Done!

violetbrina:
All comments addressed. PR open for re-review

api/routes/sample.py
```diff
@@ -60,6 +62,35 @@ async def create_new_sample(
...
+    for sample in samples.samples:
+        sample.id = sample_id_transform_to_raw(sample.id)
+
+    async with connection.connection.transaction():
+        # Table interfaces
+        st = SampleLayer(connection)
+        seqt = SampleSequenceLayer(connection)
+
+        results = await st.batch_upsert_samples(samples, seqt)
+        print(results)
```
illusional, line 84: Can you remove the extra print statement?

db/python/layers/sample.py
```diff
@@ -14,6 +45,7 @@ class SampleLayer(BaseLayer):
     def __init__(self, connection: Connection):
         super().__init__(connection)
         self.st: SampleTable = SampleTable(connection)
+        self.seqt: SampleSequencingTable = SampleSequencingTable(connection)
```
illusional, line 48: Can you move this one into the upsert_sample, if it was used in more endpoints we could promote it up, but for now I think just leave down below.

db/python/layers/sample.py
```diff
@@ -206,3 +261,17 @@ async def get_history_of_sample(
             )
 
         return rows
+
+    async def batch_upsert_samples(
+        self, samples: SampleBatchUpsert, seqt: SampleSequenceLayer
```
illusional, line 266: This function can instantiate the `SampleSequenceLayer(self.connection)`, I think that would be cleaner than passing it in as a variable.

illusional (approved):
Looks awesome! I've left a few minor comments.
Can you add a `bumpversion minor` commit to this, and then should be all good to merge!

PR #138, opened 2022-02-16
Ensure typescript API is using correct URL

illusional:
The new API isn't working on sample-metadata.populationgenomics because the URL returned by openapi in the generation during deploy was defaulting to the "dev" environment.

Unfortunately the typescript-axios environment requires a fixed URL (relative base path doesn't work), so as a work-around I've made the deploy stage pass an SM_ENVIRONMENT url, so the server will return the "correct" base path for the environment we're building for.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -42,8 +42,8 @@ jobs:
 
       - uses: actions/setup-java@v2
         with:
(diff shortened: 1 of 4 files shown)


violetbrina:
All looks good to me!

violetbrina (approved):
Approved!

PR #141, opened 2022-02-17
Fix non ordinary display values in grid

illusional:
In the project summary grid, I try to block out anything with large nested values (like the reads block), but I missed the `gvcf` for tob-wgs. React doesn't like json objects displayed directly (this currently crashes the the react app and there's no error boundary 😞). 

This PR adds a `sanitiseValue` function that simply checks the type, and if it's not a supported type then it `json.stringify`-ies anything it doesn't recognise to avoid these crashes.

### DIFF
--- a/web/src/project/ProjectSummary.tsx
+++ b/web/src/project/ProjectSummary.tsx
@@ -12,6 +12,14 @@ import { Input, Button } from 'reactstrap'
 
 const PAGE_SIZES = [20, 40, 100, 1000]
 
+const sanitiseValue = (value: any) => {
+    const tvalue = typeof value
+    if (tvalue === 'string' || tvalue === 'number') return value
+    if (tvalue === 'boolean') return value ? 'true' : 'false'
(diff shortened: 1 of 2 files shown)


violetbrina (commented):
Looks good to me. Just a clarification. Sanitization of the string will allow for numbers and strings returned. Is that the expected behaviour?

web/src/project/ProjectSummary.tsx
```diff
@@ -12,6 +12,14 @@ import { Input, Button } from 'reactstrap'
 
 const PAGE_SIZES = [20, 40, 100, 1000]
 
+const sanitiseValue = (value: any) => {
+    const tvalue = typeof value
+    if (tvalue === 'string' || tvalue === 'number') return value
```
violetbrina, line 17: Returns a number or string, this is correct? All other types seem to be stringified.

violetbrina (approved):

PR #144, opened 2022-02-21
Fix update sample status

vivbak:
Addresses https://github.com/populationgenomics/sample-metadata/issues/4

### DIFF
--- a/db/python/layers/sequence.py
+++ b/db/python/layers/sequence.py
@@ -188,6 +188,16 @@ class SampleSequenceLayer(BaseLayer):
         seq_id = self.get_latest_sequence_id_for_sample_id(sample_id)
         return self.update_status(seq_id, status, check_project_id=False)
 
+    async def update_sample_sequence_status_by_type(
+        self, sample_id: int, status: SequenceStatus, sequence_type: SequenceType
+    ):
+        """Update the sequencing status from the internal sample id"""
+        seq_id = await self.seqt.get_latest_sequence_id_by_type(
+            sample_id, sequence_type
+        )
+
+        return await self.update_status(seq_id, status, check_project_id=False)
+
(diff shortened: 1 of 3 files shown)


db/python/tables/sequence.py
```diff
@@ -135,6 +135,27 @@ async def get_latest_sequence_id_for_sample_id(
...
+ORDER by sq.id DESC
+LIMIT 1
+"""
+        result = await self.connection.fetch_one(
+            _query, {'sample_id': sample_id, 'type': sequence_type.value}
+        )
+        if not result:
+            raise NotFoundError
+
+        return result['id']
```
illusional, line 157: This should return the ProjectID as well.

db/python/layers/sequence.py
```diff
@@ -188,6 +188,16 @@ async def update_sequencing_status_from_internal_sample_id(
...
 
+    async def update_sample_sequence_status_by_type(
+        self, sample_id: int, status: SequenceStatus, sequence_type: SequenceType
+    ):
+        """Update the sequencing status of a given type from the internal sample id"""
+        seq_id = await self.seqt.get_latest_sequence_id_by_type(
+            sample_id, sequence_type
+        )
+
+        return await self.update_status(seq_id, status, check_project_id=False)
```
illusional, line 199: Should check the project ID before, esp if you're asking to skip the project check on update status

illusional (changes requested):

vivbak:
Closing as I believe this has actually been handled already 

- You can pull the appropriate sequence ID here, then update accordingly https://github.com/populationgenomics/sample-metadata/blob/96649a7d0d8e8836c0e7709a55b46d78b6f7f1d2/api/routes/sequence.py#L118

PR #146, opened 2022-02-22
Merge samples endpoint

violetbrina:
Merge together two samples as well as all of their sequences and analyses


violetbrina:
Current merge approach for sequences and analysis is simply to override the foreign key in those tables with the new one. This may cause duplicates (if duplicate sequences or analyses were created) but will also ensure no loss of data if there are unique seqs/analyses in each sample.

db/python/tables/sample.py
```diff
@@ -97,6 +97,78 @@ async def update_sample(
...
+    async def merge_samples(
+        self,
+        id_keep: int = None,
+        id_merge: int = None,
+        author: str = None,
+    ):
+        """Merge two samples together"""
+        sid_merge = sample_id_format(id_merge)
+        _, sample_keep = await self.get_single_by_id(id_keep)
+        _, sample_merge = await self.get_single_by_id(id_merge)
```
illusional, line 109: Can do these all in one to avoid multiple rounds to the database, using:

```suggestion
        samples =  await get_samples_by(sample_ids=[id_keep, id_merge])
        samples_map = {s.id: s for s in samples}
        sample_keep = samples_map[id_keep]
        sample_merge = samples_map[id_merge]
```

db/python/tables/sample.py
```diff
@@ -97,6 +97,78 @@ async def update_sample(
...
+
+        def list_merge(l1: Any, l2: Any) -> List:
+            lst: List = [l1, l2]
+            if isinstance(l1, list) and isinstance(l2, list):
+                lst = l1 + l2
+            elif isinstance(l1, list):
+                lst = l1 + [l2]
+            elif isinstance(l2, list):
+                lst = [l1] + l2
+            return lst
```
illusional, line 120: I think pylint would prefer you to return early:

```
if <condition1>:
    return l1 + l2
if <condition2>:
   return l1 + [l2]
# ... etc
```

db/python/tables/sample.py
```diff
@@ -97,6 +97,78 @@ async def update_sample(
...
+
+        _query = """
+            UPDATE sample
+            SET author = :author,
+                meta = JSON_MERGE_PATCH(:meta, :merged_from)
+            WHERE id = :id
+        """
+        _query_seqs = f"""
+            UPDATE sample_sequencing
+            SET sample_id = {id_keep}
```
illusional, line 148: Have to use a placeholder for these queries and substitute using the second param of `connection.execute`.

db/python/tables/sample.py
```diff
@@ -97,6 +97,78 @@ async def update_sample(
...
+                lst = l1 + [l2]
+            elif isinstance(l2, list):
+                lst = [l1] + l2
+            return lst
+
+        def dict_merge(meta1, meta2):
+            d = dict(meta1)
+            d.update(meta2)
+            for key, value in meta2.items():
+                if key not in meta1.keys() or meta1[key] is None or value is None:
```
illusional, line 126:
```suggestion
                if key not in meta1 or meta1[key] is None or value is None:
```

api/routes/sample.py
```diff
@@ -205,6 +205,21 @@ async def update_sample(
     return result
 
 
+@router.patch('/{id_keep}/{id_merge}', operation_id='mergeSamples')
+async def merge_samples(
+    id_keep: str,
+    id_merge: str,
+    connection: Connection = get_projectless_db_connection,
+):
+    """Merge two samples"""
```
illusional, line 214:
```suggestion
    """
    Merge one sample into another, this function achieves the merge
    by rewriting all sample_ids of {id_merge} with {id_keep}. You must
    carefully consider if analysis objects need to be deleted, or other
    implications BEFORE running this method.
    """
```

illusional (changes requested):
A few changes, mostly related to query placeholders :)

db/python/tables/sample.py
```diff
@@ -97,6 +97,80 @@ async def update_sample(
...
+                d[key] = list_merge(meta1[key], value)
+
+            return d
+
+        values: Dict[str, Any] = {
+            'sample': {
+                'id': id_keep,
+                'author': author or self.author,
+                'meta': to_db_json(dict_merge(meta_map[id_keep], meta_map[id_merge])),
+                'merged_from': merged_from,
```
illusional, line 137:
```suggestion
        # this handles merging a sample that has already been merged
        meta_original = sample_keep.meta
        meta_original['merged_from'] = list_merge(meta_original.get('merged_from'), sid_merge)
        meta: Dict[str, any] = dict_merge(meta_orginal, sample_merge.meta)
        values: Dict[str, Any] = {
            'sample': {
                'id': id_keep,
                'author': author or self.author,
                'meta': to_db_json(meta),
```

db/python/tables/sample.py
```diff
@@ -97,6 +97,80 @@ async def update_sample(
...
+            },
+            'ids': {'id_keep': id_keep, 'id_merge': id_merge},
+        }
+
+        _query = """
+            UPDATE sample
+            SET author = :author,
+                meta = JSON_MERGE_PATCH(:meta, :merged_from)
```
illusional, line 145:
```suggestion
                meta = :meta
```

db/python/tables/sample.py
```diff
@@ -97,6 +97,80 @@ async def update_sample(
...
+        _del_sample = f"""
+            DELETE FROM sample
+            WHERE id = :id_merge
+        """
+
+        async with self.connection.transaction():
+            await self.connection.execute(_query, {**values['sample']})
+            await self.connection.execute(_query_seqs, {**values['ids']})
+            await self.connection.execute(_query_analyses, {**values['ids']})
+            await self.connection.execute(_del_sample, id_merge=id_merge)
```
illusional, line 167:
```suggestion
            await self.connection.execute(_del_sample, {'id_merge': id_merge})
```

db/python/tables/sample.py
```diff
@@ -97,6 +97,80 @@ async def update_sample(
...
 
+    async def merge_samples(
+        self,
+        id_keep: int = None,
+        id_merge: int = None,
+        author: str = None,
+    ):
+        """Merge two samples together"""
+        sid_merge = sample_id_format(id_merge)
+        _, samples = await self.get_samples_by(sample_ids=[id_keep, id_merge])
```
illusional, line 108: This was probably bad advice, looks like this returns a SampleModel which is casting the ID to a string, causing errors, I think your original code here was better, but wrapping it in a asyncio.gather so we don't have to wait for 2 sequential round trips:

```suggestion
                (_, sample_keep), (_, sample_merge) = await asyncio.gather(
            self.get_single_by_id(id_keep),
            self.get_single_by_id(id_merge),
        )
```

illusional (commented):
I've added a few more suggestions as had some issues when trying this locally. Could you give this one more shot locally, to make sure my suggestions do work (and I didn't half suggest broken changes), then this should be good to merge.

illusional (approved):
Amazing work on this PR, excited to see it being used soon!

PR #150, opened 2022-02-24
Modify get family to filter for sample and participant id

### LINKED ISSUE #145: Have an easy endpoint to access all the participants and/or samples in a given family
Get a list of families filtering on a subset of sample ids and/or participant ids.

violetbrina:
Closes #145

### DIFF
--- a/db/python/layers/family.py
+++ b/db/python/layers/family.py
@@ -250,12 +250,15 @@ class FamilyLayer(BaseLayer):
         all_participants = participant_ids if participant_ids else []
 
         # Find the participants from the given samples
-        _, samples = await self.stable.get_samples_by(
-            project_ids=[project], sample_ids=sample_ids
-        )
+        if sample_ids is not None and len(sample_ids) > 0:
+            _, samples = await self.stable.get_samples_by(
+                project_ids=[project], sample_ids=sample_ids
+            )
 
-        all_participants += [int(s.participant_id) for s in samples if s.participant_id]
-        all_participants = list(set(all_participants))
+            all_participants += [
+                int(s.participant_id) for s in samples if s.participant_id
+            ]
+            all_participants = list(set(all_participants))
 
         return await self.ftable.get_families(
             project=project, participant_ids=all_participants
--- a/db/python/layers/family.py
+++ b/db/python/layers/family.py
@@ -244,20 +244,21 @@ class FamilyLayer(BaseLayer):
         sample_ids: List[int] = None,
     ):
         """Get all families for a project"""
+        project = project or self.connection.project
 
         # Merge sample_id and participant_ids into a single list
         all_participants = participant_ids if participant_ids else []
 
         # Find the participants from the given samples
-        _, sid_participants = await self.stable.get_samples_by(
+        _, samples = await self.stable.get_samples_by(
             project_ids=[project], sample_ids=sample_ids
         )
 
-        all_participants += [int(s.participant_id) for s in sid_participants]
+        all_participants += [int(s.participant_id) for s in samples if s.participant_id]
         all_participants = list(set(all_participants))
 
         return await self.ftable.get_families(
-            project=project, participant_ids=participant_ids
+            project=project, participant_ids=all_participants
         )
 
     async def update_family(
--- a/db/python/layers/family.py
+++ b/db/python/layers/family.py
@@ -8,6 +8,7 @@ from db.python.tables.family import FamilyTable
 from db.python.tables.family_participant import FamilyParticipantTable
 from db.python.tables.participant import ParticipantTable
 from db.python.tables.project import ProjectId
+from db.python.tables.sample import SampleTable
 
 
 class PedRow:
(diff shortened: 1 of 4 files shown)


db/python/tables/family.py
```diff
@@ -29,16 +29,32 @@ async def get_projects_by_family_ids(self, family_ids: List[int]) -> Set[Project
...
-        )
+        where: List[str] = []
+
+        if participant_ids:
+            _query += """
+                JOIN family_participant
+                ON family.id = family_participant.family_id
+            """
+            pids = ', '.join([str(x) for x in participant_ids])
+            where.append(f'participant_id IN ({pids})')
```
illusional, line 49: We should use SQL placeholders here to mitigate risk of SQL injection attacks, it's not practical given the current structure of code, but this could be changed and forgotten this is constructing the values directly.

db/python/tables/family.py
```diff
@@ -29,16 +29,32 @@ async def get_projects_by_family_ids(self, family_ids: List[int]) -> Set[Project
...
+
+        if project:
+            where.append(f'project = {project or self.project}')
```
illusional, line 52: Same with how the previous uses `:project` with then the second argument to query `{'project': project or self.project}`

db/python/tables/family.py
violetbrina: Ah ok makes sense. I can convert it.

illusional: The SQL engine we're using is clever enough that it knows arrays, so you can just do something like

```
values_to_replace = {}
if participant_ids:
    where.append(f'participant_id IN :pids')
    values_to_replace['pids'] = participant_ids
# ....
connection.execute(_query, values)


```

db/python/tables/family.py
```diff
@@ -29,16 +29,39 @@ async def get_projects_by_family_ids(self, family_ids: List[int]) -> Set[Project
...
+            placeholders = []
+            for i, pid in enumerate(participant_ids):
+                values[str(i)] = pid
+                placeholders.append(f':{i}')
+
+            pids = ', '.join(placeholders)
+            where.append(f'participant_id IN ({pids})')
+
+        if project or self.project:
+            where.append('project = :project')
```
illusional, line 59: Kind of an unwritten rule, but if project isn't set, then you have to explicitly look up the projects that the participant / family IDs belong to and access check. So I'd advise to just always do this check and probably assert project exists.

illusional (changes requested):

db/python/tables/family.py
```diff
@@ -29,16 +29,33 @@ async def get_projects_by_family_ids(self, family_ids: List[int]) -> Set[Project
...
+            FROM family
+        """
 
-        rows = await self.connection.fetch_all(
-            _query, {'project': project or self.project}
-        )
+        values: Dict[str, Any] = {'project': project or self.project}
+        where: List[str] = []
+
+        if participant_ids:
```
illusional, line 44: This is returning multiple families for multiple individuals specified within the family. I suspect this will be cleaner to write two different queries for each case (no participants OR participants).

db/python/tables/family.py
```diff
@@ -29,16 +29,33 @@ async def get_projects_by_family_ids(self, family_ids: List[int]) -> Set[Project
...
+            where.append(f'participant_id IN (:pids)')
+            values['pids'] = participant_ids
+
+        if project or self.project:
+            where.append('project = :project')
+
+        if where:
+            _query += 'WHERE ' + ' AND '.join(where)
+
+        rows = await self.connection.fetch_all(_query, {**values})
```
illusional, line 58:
```suggestion
        rows = await self.connection.fetch_all(_query, values)
```

illusional (changes requested):
A couple of issues when specifying multiple samples, or multiple individuals

db/python/layers/family.py
```diff
@@ -232,12 +233,33 @@ class FamilyLayer(BaseLayer):
...
+        # Merge sample_id and participant_ids into a single list
+        all_participants = participant_ids if participant_ids else []
+
+        # Find the participants from the given samples
+        _, samples = await self.stable.get_samples_by(
+            project_ids=[project], sample_ids=sample_ids
+        )
+
+        all_participants += [int(s.participant_id) for s in samples if s.participant_id]
+        all_participants = list(set(all_participants))
```
illusional, line 258: When samples is false-y (empty list / unspecified), this returns all participant IDs within the project (that have samples) - regardless of the specified participants.

db/python/tables/family.py
```diff
@@ -29,17 +29,39 @@ async def get_projects_by_family_ids(self, family_ids: List[int]) -> Set[Project
...
+        if where:
+            _query += 'WHERE ' + ' AND '.join(where)
+
+        rows = await self.connection.fetch_all(_query, values)
+        seen = set()
+        families = []
+        for r in rows:
+            if r.id not in seen:
+                families.append(Family.from_db(dict(r)))
+                seen.add(r.id)
```
illusional, line 64: This works, but I think it would be cleaner to make the database do this by using a GROUP BY (on all values returned), eg:

```sql
SELECT f.id, f.external_id, f.description, f.coded_phenotype, f.project
FROM family f
INNER JOIN family_participant fp
   ON f.id = fp.family_id             
WHERE fp.participant_id IN (97, 100, 101) AND f.project = 3
GROUP BY f.id, f.external_id, f.description, f.coded_phenotype, f.project
```

I'd probably encourage writing separate queries for when the participants are specified, and when they're not. Gives you more flexibility to add the GROUP BY.

illusional (changes requested):

db/python/layers/family.py
vivbak: This is interesting, because it's kind of the initial expected behaviour. i.e., return all families, but it does kind of seem unintuitive. Would you imagine this actually being split into a separate endpoint? e.g. `get_all_families` and `get_families_by_participants`?

db/python/tables/family.py
violetbrina: Yeah I was thinking about switching to the native group by. But ended up with weird issues since if no pariticpants or samples or specified there is no group by and the aggregate was acting weird.

illusional (approved):

PR #151, opened 2022-02-24
Add validate option to generic parser

illusional:
Addresses "all uploaded files must be covered by the sample map" component of the seqr wizard.


vladsaveliev (approved):
Looks good!

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -49,6 +50,49 @@ def __init__(
...
+        Check that the files in the search_paths are completely covered by the sample_map
+        """
+        filenames = []
+        for sm in (sample_map if isinstance(sample_map, list) else [sample_map]):
+            for rows in sm.values():
+                if isinstance(rows, list):
+                    for r in rows:
+                        filenames.extend(r.get(self.reads_column, '').split(','))
+                else:
+                    filenames.extend(rows.get(self.reads_column, '').split(','))
```
vladsaveliev, line 77: A minor one, but it might be a bit more clear if it implemented the way as for `sample_map`:
```suggestion
                for r in (rows if if isinstance(rows, list) else [rows]):
                    filenames.extend(r.get(self.reads_column, '').split(','))
```

PR #152, opened 2022-02-28
Sample file map parser update - new format

### LINKED ISSUE #149: Update sample map parser to support modified format
Modified format includes that separates individual ID and sample ID (where they are different), and includes a sample_type (WGS / WES).

Column ID | Description | Notes
-- | -- | --
Individual ID | Alphanumeric individual ID (if different to the Sample ID) | The IDs are alphanumeric: the individual ID should uniquely identify a person.
Sample ID | Alphanumeric sample ID | A sample ID should be unique within a project. Note that an individual can have multiple samples.
Filenames | Comma-separated list of filenames for this sample. | If more than two files are provided, they will be grouped automatically
Type | One of the following: WGS, WES | Default=WGS, if left blankList whether the sequencing is WGS (Whole genome), or (WES) whole-exome sequencing. Note that if using comma-separated list of filenames, and a sample has both WES and WGS, you should include one row for each type.

Example:

Individual ID | Sample ID | File names | Type
-- | -- | -- | --
CPG_001_I | A0001 | A0001-R1.fastq.gz, A0001-R2.fastq.gz | WGS
CPG_001_I | A0001 | A0001_WES-R1.fastq.gz,A0001_WES-R2.fastq.gz | WES
CPG_001_I | A0002 | A002-R1.fastq.gz, A0002-R2.fastq.gz | WGS

violetbrina:
Closes #149


sample_metadata/parser/generic_parser.py
```diff
@@ -75,6 +77,19 @@
 SUPPORTED_VARIANT_TYPES = Literal['gvcf', 'vcf']
 
 
+class SequenceMetaGroup(BaseModel):
```
illusional, line 80: This doesn't need to inherit from pydantic.BaseModel, as we're not serialising this model it right?

sample_metadata/parser/generic_parser.py
```diff
@@ -113,6 +128,8 @@ def __init__(
         default_sequence_type='wgs',
```
illusional, line 128: I think this default_sequence_type if now outdated too right?

sample_metadata/parser/generic_parser.py
```diff
@@ -129,12 +146,14 @@ def __init__(
...
         self.default_sample_type: str = default_sample_type
+        self.default_analysis_type: str = default_analysis_type
+        self.default_analysis_status: str = default_analysis_status
 
         # gs specific
         self.default_bucket = None
 
         self.client = None
-        self.bucket_clients: Dict[str, Any] = {}
+        self.bucket_clients: SingleRow = {}
```
illusional, line 156: I don't think this type change makes semantic sense.

sample_metadata/parser/generic_parser.py
```diff
@@ -66,7 +67,8 @@
 )
 
 rmatch = re.compile(r'[_\.-][Rr]\d')
-GroupedRow = Union[List[Dict[str, Any]], Dict[str, Any]]
+SingleRow = Dict[str, Any]
+GroupedRow = Union[List[SingleRow], SingleRow]
```
illusional, line 71: Could we refactor this union out, as this script has got more complex, it just makes us double handle list checks - I think the parser should just _always_ make sure it's a list. 

```suggestion
GroupedRow = List[SingleRow]
```

sample_metadata/parser/sample_file_map_parser.py
```diff
@@ -4,17 +4,19 @@
...
+INDIVIDUAL_ID_COL_NAME = 'Individual ID'
+SAMPLE_ID_COL_NAME = 'Sample ID'
 READS_COL_NAME = 'Filenames'
+TYPE_COL_NAME = 'Type'
 
 __DOC = """
 The SampleFileMapParser is used for parsing files with format:
 
-- 'Individual ID'
+- 'Sample ID'
```
illusional, line 15: Could you update this docstring, looks like you've implemented it below though.

illusional (dismissed):
This is looking really great! I haven't tested the implementation yet, but the changes to the examples look pretty clean!

I have a few pieces of feedback:

General minor feedback:
- There is some inconsistency between the route names and function names, and the verbage between it, mostly there are some routes with `byX` , but IMO: `byX` implies the result it keyed by `X`.
- Remove trailing `/` in routes.
- It should be possible within the server to key by SequenceType (because that enum should be hashable). Potentially on the parsers you could alias `SequenceTypeStr = str` to make it clear when a dictionary is keyed by the literal value for the sequence type.

Structural:
- I really like the refactor for more concrete typed objects, I think it's a lot cleaner. Just make sure where you've replaced `Dict[str, Any]`, it makes conceptual sense.
- I'm not fully getting how sequences are split up by sequence type. To me, the default behaviour should be that the genericParser has:
  - a function called `split_rows_by_sequence_type` that groups by sequence type into a list of `SequenceMetaGroup`, ONE per sequence type, 
  - then `get_sequence_meta` should just take a `SequenceMetaGroup` (with no meta), and returns a `Dict[str, Any]` which gets set onto that `SequenceMetaGroup.meta`. 
  - If someone wants to, they could override `split_rows_by_sequence_type` to change that default behaviour.

violetbrina:
New addition of the participant_meta_map.
Had to modify all of the parsers to varying degrees in order for them not to break, this makes it a pretty massive PR. Let me know if anything is unclear or if there's something that should be changed.

sample_metadata/parser/generic_parser.py
```diff
@@ -310,64 +402,133 @@ async def process_group(
...
+            )
+        )
+
+        # Get all the samples and sequences to upsert first
+        samples_to_upsert = []
+        sequences_to_upsert = []
+        analyses_to_add = []
+
+        for sample_id, rows in sample_map.items():
+            cpg_id = existing_external_id_to_cpgid[sample_id]
```
illusional, line 498: They shouldn't always have a cpg_id right? I haven't run it, but I thought the direct access for `sample_id` would fail if the sample doesn't exist.

sample_metadata/parser/generic_parser.py
```diff
@@ -310,64 +402,133 @@ async def process_group(
...
+            sample, seqs, analyses = await self.process_sample_group(
+                rows, sample_id, cpg_id
+            )
+            samples_to_upsert.append(sample)
+            sequences_to_upsert.extend(seqs)
+            analyses_to_add.append(analyses)
+
+        # Construct participant to upsert
+        papi = ParticipantApi()
```
illusional, line 507: Could we share this ParticipantApi instance across the whole parser?

sample_metadata/parser/generic_parser.py
```diff
@@ -399,114 +582,279 @@ async def validate_rows(
...
+            )
+
+        if len(sample_map) != 0:
+            return await self.parse_manifest_by_samples(
+                sample_map, confirm=confirm, dry_run=dry_run
+            )
+
+        return await self.parse_manifest_by_participants(
+            participant_map, confirm=confirm, dry_run=dry_run
+        )
```
illusional, line 616: I think just keep this branched the whole function, you'll have a little bit of code duplication but there's no need to regroup to do the error check.

sample_metadata/parser/generic_parser.py
```diff
@@ -399,114 +582,279 @@ async def validate_rows(
...
+                raise SystemExit()
+        else:
+            logger.info(message)
+
+        # Batch update
+        result = papi.batch_upsert_participants(
+            proj, ParticipantUpsertBody(participants=all_participants)
+        )
+
+        # TODO: analyses adding is in batch by sample, not yet done for batch by participant
```
illusional, line 771: Can this commented code be removed now?

sample_metadata/parser/sample_file_map_parser.py
```diff
@@ -1,20 +1,30 @@
...
 
-- 'Individual ID'
+- ['Individual ID']
+- 'Sample ID'
 - 'Filenames'
+- ['Type']
 
 EG:
-    Individual ID   Filenames
+    Sample ID   Filenames
```
illusional, line 27: Do you mind adding a second example here with the Participant Id and Sequence type?

scripts/parse_generic_metadata.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # pylint: disable=too-many-instance-attributes,too-many-locals,unused-argument,no-self-use,wrong-import-order,unused-argument,too-many-arguments
-from typing import List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
```
illusional, line 3: This file is completely contained within the parser now right?
A quick way to remove the code but keep the functionality could be to call the main function within this file, eg:

```python
if __name__ == '__main__':
    from sample_metadata.parser.generic_metadata_parser import main
    main() # pylint: disable=no-value-for-parameter
```

illusional (changes requested):
This is looking amazing! I like how simple it will be to add extra mapping, and how much work it does behind the scenes, really awesome job on this PR.

I've added a few small comments, mostly about making API calls in effectively a loop (if we can avoid it), and sharing the SampleMetadata Api instances to avoid generating lots of access tokens.

.gitignore
```diff
@@ -6,7 +6,9 @@ __pycache__/
 .DS_Store
 .idea/
 db/mariadb-java-client-*.jar
-sample_metadata
+!sample_metadata/
+sample_metadata/*
+!sample_metadata/parser/
```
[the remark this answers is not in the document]
violetbrina: Sure thing!

db/python/tables/sequence.py
```diff
@@ -123,36 +126,115 @@ async def get_latest_sequence_id_for_sample_id(
...
+            ORDER by sq.id DESC
+            LIMIT 1
+        """
         result = await self.connection.fetch_one(_query, {'sample_id': sample_id})
         if not result:
             raise NotFoundError
 
         return result['project'], result['id']
 
+    async def get_sequence_id_from_sample_id_and_type(
```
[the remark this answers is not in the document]
violetbrina: Ah true. I can modify the name.

db/python/tables/sequence.py
```diff
@@ -123,36 +126,115 @@ async def get_latest_sequence_id_for_sample_id(
...
+            INNER JOIN sample s ON sq.sample_id = s.id
+            WHERE sample_id = :sample_id
+            ORDER by sq.id DESC
+        """
+        result = await self.connection.fetch_all(_query, {'sample_id': sample_id})
+        if not result:
+            raise NotFoundError
+
+        seq_map: Dict[str, int] = defaultdict(int)
+        projects: List[str] = []
```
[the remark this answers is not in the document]
violetbrina: Done! Converted to list of ints

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -44,17 +111,89 @@ def __init__(
...
         self.allow_extra_files_in_search_path = allow_extra_files_in_search_path
 
-    async def validate_rows(
-        self, sample_map: Dict[str, Union[dict, List[dict]]]
-    ):
+    def get_sample_id(self, row: SingleRow) -> Optional[str]:
+        """Get external sample ID from row"""
+        return row.get(self.sample_name_column, None)
+
+    def get_cpg_sample_id(self, row: SingleRow) -> Optional[str]:
```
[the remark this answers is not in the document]
violetbrina: Done!

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -275,6 +456,7 @@ async def get_sequence_meta(
         if gvcf_filenames:
             full_filenames.extend(self.file_path(f.strip()) for f in gvcf_filenames)
 
+        sample_id = self.get_cpg_sample_id(rows[0])
```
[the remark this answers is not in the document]
violetbrina: Done!

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -338,3 +517,125 @@ async def from_manifest_path(
...
+    nargs=2,
+    multiple=True,
+    help='Two arguments per listing, eg: --qc-meta-field "name-in-manifest" "name-in-analysis.meta"',
+)
+@click.option(
+    '--sample-meta-field',
+    multiple=True,
+    help='Single argument, key to pull out of row to put in sample.meta',
+)
+@click.option(
```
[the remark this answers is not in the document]
violetbrina: Done!

sample_metadata/parser/generic_parser.py
```diff
@@ -237,71 +286,114 @@ async def file_size(self, filename):
...
     @abstractmethod
     async def get_sequence_meta(
-        self, sample_id: str, row: GroupedRow
-    ) -> Dict[str, Any]:
-        """Get sequence-metadata from row"""
+        self, seq_group: SequenceMetaGroup
+    ) -> SequenceMetaGroup:
+        """Get sequence-metadata from row then set it in the SequenceMetaGroup"""
 
+    # @abstractmethod
```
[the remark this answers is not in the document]
violetbrina: Fixed, add return {} by default

sample_metadata/parser/generic_parser.py
violetbrina: Correct, but the process_sample_group was doing api calls to get the cpg_id. So I did one batch at the beginning and pass in both the sample_id (always present) and cpg_id (if it exists). Otherwise the cpg_id will just be None.

sample_metadata/parser/generic_parser.py
violetbrina: Sure!

sample_metadata/parser/generic_parser.py
violetbrina: Don't quite follow....

sample_metadata/parser/generic_parser.py
violetbrina: Sure!

sample_metadata/parser/sample_file_map_parser.py
violetbrina: Done!

scripts/parse_generic_metadata.py
violetbrina: Yeah I didn't want to delete it before full copying over then I forgot lol.

violetbrina:
Modified the pre-commit config as per this issue: [https://github.com/psf/black/issues/2964](https://github.com/psf/black/issues/2964)

illusional (changes requested):
This is looking good! I still had a problem when inserting samples with a KeyError, and I've flagged a few small minor things.

It would be awesome to add a test for this new participant upsert of the parser too.

db/project.xml
```diff
@@ -323,4 +323,16 @@
 		<sql>SET @@system_versioning_alter_history = 1;</sql>
 		<sql>ALTER TABLE family MODIFY COLUMN coded_phenotype TEXT;</sql>
 	</changeSet>
+	<changeSet author="sabrina.yan" id="2022-03-07_standardise-sequence-type">
+		<sql>SET @@system_versioning_alter_history = 1;</sql>
+		<sql>ALTER TABLE sample_sequencing MODIFY COLUMN type VARCHAR(30);</sql>
```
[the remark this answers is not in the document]
violetbrina: Ah true, order fixed

sample_metadata/parser/generic_parser.py
```diff
@@ -459,7 +466,7 @@ async def process_sample_group(
         )
 
     async def process_participant_group(
-        self, participant_id: int, sample_map: Dict[str, Any]
+        self, participant_id: str, sample_map: Dict[str, Any]
```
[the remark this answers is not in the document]
violetbrina: Ah good point. Yeah I can clarify.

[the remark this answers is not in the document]
violetbrina: I've renamed it to participant name. In the function the int internal id is called internal_id so it's clear.

sample_metadata/parser/sample_file_map_parser.py
```diff
@@ -23,13 +23,27 @@
...
-    Sample ID   Filenames
+e.g.
+    Sample ID       Filenames
     <sample-id>     <sample-id>.filename-R1.fastq.gz,<sample-id>.filename-R2.fastq.gz
     # OR
     <sample-id2>    <sample-id2>.filename-R1.fastq.gz
     <sample-id2>    <sample-id2>.filename-R2.fastq.gz
 
+Example with optional columns
+Note: Individual ID column must contain values in every row
```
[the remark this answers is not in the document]
violetbrina: Because it calls the participant upsert endpoint which has an entirely different structure. It's keyed on the participant name (external participant id). If we have samples without a participant we can use the batch sample upsert endpoint. 

I think it just makes it a bit cleaner. In a weird case where we have some samples and not others we would have two files. 

Otherwise, we can add a ticket to allow participant-less upsert of the samples without a participant. I'd add that as another feature request/ticket.

sample_metadata/parser/generic_parser.py
```diff
@@ -310,64 +409,133 @@ async def process_group(
...
+        analyses_to_add = []
+
+        for sample_id, rows in sample_map.items():
+            cpg_id = existing_external_id_to_cpgid[sample_id]
+            sample, seqs, analyses = await self.process_sample_group(
+                rows, sample_id, cpg_id
+            )
+            samples_to_upsert.append(sample)
+            sequences_to_upsert.extend(seqs)
+            analyses_to_add.append(analyses)
```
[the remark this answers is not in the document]
violetbrina: Ah good points. The cpg_id is optional, so I've changed it to a .get(sample_id, None) which won't get a key error thrown and the process_sample_group will still work.

sample_metadata/parser/generic_parser.py
```diff
@@ -399,116 +591,255 @@ async def validate_rows(
...
+                summary = self.upsert_summary(
+                    summary,
+                    [participant_to_upsert],
+                    samples_to_upsert,
+                    sequences_to_upsert,
                 )
-                current_batch_promises[external_sample_id] = promise
+
+        message = f"""\
+            {proj}: Processing samples: {', '.join(sample_map.keys())}
```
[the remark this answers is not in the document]
violetbrina: Changed to processing # participants

illusional (approved):
Amazing 🎉 
Love the tests!!

PR #159, opened 2022-02-28
Fix no-participants-to-add check

illusional:


db/python/layers/participant.py
```diff
@@ -210,13 +210,15 @@ async def fill_in_missing_participants(self):
             project=self.connection.project,
             allow_missing=True,
         )
-        if not unlinked_participants:
```
vivbak, line 213: Thanks for the fix! I was really unsure of what this was aiming to catch. Makes a lot of sense now.

vivbak (approved):
Thank you!! 🙏

PR #163, opened 2022-03-04
Seqr fixes

illusional:
- Fix implementation of get_sample_reads_map_for_seqr (was returning external sample ID, now correctly returns external individual ID)
- Add flip columns parameter to getExternalParticipantIdToInternalSampleID to make it more useful for importing into seqr.


vladsaveliev (approved):
Thanks!

PR #165, opened 2022-03-07
Update create-test-subset 

vivbak:
Pulls participant, family & pedigree information to test project.


vivbak:
Depends on https://github.com/populationgenomics/sample-metadata/pull/150

scripts/create_test_subset.py
```diff
@@ -187,15 +192,67 @@ def main(
...
+    tmp_family_tsv = 'tmp_families.tsv'
+    family_tsv_headers = ['Family ID', 'Description', 'Coded Phenotype', 'Display Name']
+    # Work-around as import_families takes a file.
+    with open(tmp_family_tsv, 'wt') as tmp_families:
+        tsv_writer = csv.writer(tmp_families, delimiter='\t')
+        tsv_writer.writerow(family_tsv_headers)
+        for family in families:
+            del family['id']
+            del family['project']
+            tsv_writer.writerow(list(family.values()))
```
illusional, line 233: I'd probably be more explicit with the order you write this in. Trusting the response from the SM server to be in the order with the correct keys is a bit dangerous. I'd probably prefer something that explicitly builds the keys:

```python
tsv_writer.writerow([
    family['external_id'],
    family['description'] or '',
# .... etc
```

vivbak: ah, good point -- thanks! Will fix :)

vladsaveliev (approved):
Looks good! Thanks Vivian :)

PR #167, opened 2022-03-08
Unittesting - another attempt

illusional:
- Each unittest class should add relevant data, and run tests against it
- After each unittest class, data should be removed
- ~~Attempt using the `testcontainers-python` module, but local testing is not possible because it's trying to consume port 3306 (https://github.com/testcontainers/testcontainers-python/blob/cfd6b12511d12d4762ac7f1102075b72b453975a/testcontainers/mysql.py#L43), which conflicts with my local mariadb installation.~~
- `testcontainers-python` works :)


test/manual_test_sample.py
```diff
@@ -0,0 +1,60 @@
...
+    sapi = SampleApi()
+    aapi = AnalysisApi()
+
+    def _print_samples():
+        samples = sapi.get_samples(
+            body_get_samples_by_criteria_api_v1_sample_post={
+                'project_ids': [PROJ],
+                'active': True,
+            }
+        )
```
MattWellie, line 19: is this making a real connection to the API?

illusional: This is, it's an older test I've renamed / pulled from Vlad's older PR - I'd like to rewrite this test, or come up with a new testing class that allows you to use the SM API, because for that it will have to:
- spin up a web server
- Probably ask the SM API to _NOT_ generate you an auth token because it relies on some auth'd user
- Ask the SM API to hit this local server instead, which might not necessarily run on port 8000 (the default `local` port).

violetbrina (approved):

test/manual_test_sample.py
MattWellie: When I was unit-testing a method that called the SM API I tried to spoof the responses using requests-mock, but struggled because the client required authentication in a way that I couldn't easily see how to mock. If you end up with a way to mock SMAPI/DB responses please let me know!

test/testbase.py
```diff
@@ -0,0 +1,173 @@
...
+                if am_i_in_test_environment:
+                    db_prefix = '../db'
+
+                con_string = db.get_connection_url()
+                con_string = 'mysql://' + con_string.split('://', maxsplit=1)[1]
+                lcon_string = f'jdbc:mariadb://{db.get_container_host_ip()}:{port_to_expose}/{db.MYSQL_DATABASE}'
+                # apply the liquibase schema
+                command = [
+                    'liquibase',
+                    *('--changeLogFile', db_prefix + '/project.xml'),
```
vladsaveliev, line 97: Just curious, why surrounding with parenthesis and adding with `*` here?

vladsaveliev (approved):

test/testbase.py
illusional: Stylistic, just formats to (keyword, arg) nicely

PR #175, opened 2022-03-15
Remove peddy dependency

vivbak:
As suggested https://github.com/populationgenomics/images/pull/8

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -108,6 +108,7 @@ def main(
         all_families = [family['id'] for family in fams]
         families = random.sample(all_families, families_n)
         pedigree = fapi.get_pedigree(project=project, internal_family_ids=families)
+        _print_fam_stats(pedigree)
         p_ids = [ped['individual_id'] for ped in pedigree]
         sample_ids = [
             sample
@@ -295,9 +296,10 @@ def _validate_opts(samples_n, families_n) -> Tuple[Optional[int], Optional[int]]
 
 
 def _print_fam_stats(families: List):
+    family_sizes = Counter([fam['family_id'] for fam in families])
     fam_by_size: typing.Counter[int] = Counter()
-    for fam in families:
-        fam_by_size[len(fam.samples)] += 1
+    for fam in family_sizes:
+        fam_by_size[family_sizes[fam]] += 1
     for fam_size in sorted(fam_by_size):
         if fam_size == 1:
             label = 'singles'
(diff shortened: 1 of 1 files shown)


scripts/create_test_subset.py
```diff
@@ -104,19 +103,17 @@ def main(
...
-        families = random.sample(families, families_n)
-        logger.info(f'After subsetting to {len(families)} families:')
-        _print_fam_stats(families)
-        sample_ids = []
-        for fam in families:
-            for s in fam.samples:
-                sample_ids.append(sample_id_by_participant_id[s.sample_id])
+        fams = fapi.get_families(project)
+        all_families = [family['id'] for family in fams]
+        families = random.sample(all_families, families_n)
```
illusional, line 109: Missing a `_print_fam_stats` call here

scripts/create_test_subset.py
```diff
@@ -104,19 +103,17 @@ def main(
...
+        pedigree = fapi.get_pedigree(project=project, internal_family_ids=families)
+        p_ids = [ped['individual_id'] for ped in pedigree]
+        sample_ids = [
+            sample
+            for (participant, sample) in sample_id_by_participant_id.items()
+            if participant in p_ids
+        ]
         samples = [s for s in all_samples if s['id'] in sample_ids]
```
illusional, line 117: calling `in` on a non-hashed type is an `O(n) operation, and in a loop like this is `O(n**2)`, you might consider making a set version of sample_ids to do a contains check.

illusional (commented):
I started reviewing this before I realised it was a draft, so early feedback.

scripts/create_test_subset.py
vivbak: Ah yes, would have to reimplement that function to work with the new input format. Is the print useful to have?

vivbak: Fixed in the latest :)

scripts/create_test_subset.py
vivbak: I think I've followed your suggestion in the latest push but let me know if I misinterpreted

scripts/create_test_subset.py
```diff
@@ -298,9 +297,10 @@ def _validate_opts(samples_n, families_n) -> Tuple[Optional[int], Optional[int]]
 
 
 def _print_fam_stats(families: List):
+    family_sizes = Counter([fam['family_id'] for fam in families])
     fam_by_size: typing.Counter[int] = Counter()
-    for fam in families:
-        fam_by_size[len(fam.samples)] += 1
+    for fam in family_sizes:
```
illusional, line 302:
```suggestion
    # determine number of singles, duos, trios, etc
    for fam in family_sizes:
```

illusional (approved):

PR #176, opened 2022-03-16
Fill participants during import ped

vivbak:
Previously, `create_test_subset` used the `fill_in_missing_participants` endpoint to create participants for the test project. This however, uses the external sample IDs as the external participant IDs and is invalid in the case where these two values do not match. Resulting in a [participant mismatch error ](https://batch.hail.populationgenomics.org.au/batches/13339/jobs/1)in the case of acute-care 

Assuming PED information will always be processed in the context of this script, it seems best to consistently generate the participants at the import ped stage.

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -161,50 +161,6 @@ def main(
     family_ids = transfer_families(project, target_project, participant_ids)
     external_participant_ids = transfer_ped(project, target_project, family_ids)
 
-    # # get_map_ipid_esid()
-
-    # # Acronyms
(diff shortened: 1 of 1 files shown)


vladsaveliev (approved):

vivbak:
hey hey @illusional @vladsaveliev, the work-around to fix mapping internal participant ids with external sample id's is a tad convoluted, so very happy to take on feedback if there are ways I can clean it up that I'm missing. TIA 🙏

vladsaveliev (approved):

PR #181, opened 2022-03-24
Fix Analysis Subset 

vivbak:
Previously, the internal sample IDs from the initial project were passed during the creation of new analysis objects in the test project. So new analyses were created for samples in the initial project rather than the test project. 

Further, when crams and gvcfs were copied over they maintained their initial internal IDs within the filename. This shouldn't create any issues (as paths are always passed), but this PR adjusts it for consistency.


vladsaveliev (approved):
Looks good!

PR #184, opened 2022-04-13
Convert relationships to pedigree

illusional:
Rough test - this is how I'd like to store relationships in SM in the future, so doubles as:

- Proof-of-concept that you can rebuild relationships like this
- Useful for dataset that gave us relationship information


illusional:
Hey @dancoates, this PR is super out of date so I've closed it, but just a heads-up if you were looking for improvements to the relationships model, this was my an initial idea.

dancoates:
Thanks @illusional! Will definitely keep this in mind if/when we revisit the relationships model

PR #186, opened 2022-04-19
Update prophecy parser to handle participants

vivbak:


vivbak:
Depends on https://github.com/populationgenomics/sample-metadata/pull/188

scripts/parse_prophecy.py
```diff
@@ -61,6 +56,26 @@ async def file_pointer_to_sample_map(
...
+            if line.strip() == '""':
+                break
+
+        return await super().file_pointer_to_participant_map(file_pointer, delimiter)
+
+    def get_participant_id(self, row) -> str:
+        """Get external participant ID from row"""
+        if not self.participant_column or self.participant_column not in row:
+            raise ValueError('Participant column does not exist')
+        return row[self.participant_column][:-4]
```
vladsaveliev, line 77: Why `[:-4]`?

scripts/parse_prophecy.py
```diff
@@ -97,18 +121,21 @@ async def main(
...
-    confirm=False,
+    confirm=True,
     dry_run=False,
 ):
     """Run script from CLI arguments"""
     parser = ProphecyParser(
         sample_metadata_project=sample_metadata_project,
         search_locations=search_locations,
-        sample_name_column='Sample/Name',
+        sample_name_column='External ID',
```
vladsaveliev, line 131: Remind me, what did we agree with regards to sample IDs? I recall that we discussed using `participantid` (e.g. `658`) instead of `Specimen ID`=`External ID` (e.g. `6580308`) as the DB participant IDs, and probably keep those `220208_FD02700719` as DB external sample IDs. But I didn't take notes and forgot what we ended up with :(

vladsaveliev: Ah, perhaps `External ID`(`6580308`) should match external sample IDs in the DB, and `658` should be the participant ID

scripts/parse_prophecy.py
vladsaveliev: Right, to strip off. I think it's safer to instead parse that other file that we got from them, where we have both `Specimen ID` and `participantid` columns. Just in case the format changes and the `[:-4]` holds no longer.

vivbak: Removed :)

scripts/parse_prophecy.py
vivbak: Yes, you're right! External ID = Specimen ID which is the Sample ID.  I've fixed it now. `658` would be the Participant ID.  Fixed now.

vladsaveliev (approved):
Awesome!

PR #187, opened 2022-04-20
Add participant ids to samples prior to upsert

vivbak:
Previously, when samples were upserted via `upsert_participant`, their corresponding participant id's were not correctly passed, and were instead set to `None`. This PR addresses this issue.


vivbak:
@illusional :)

illusional (approved):

PR #188, opened 2022-04-20
Add participant fields

vivbak:
Currently, participant fields (`sex`, `reported_gender` and `karyotype`) are not captured by the generic parser. This PR addresses that.


sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -186,6 +192,28 @@ def get_participant_id(self, row: SingleRow) -> Optional[str]:
...
+        if not self.reported_sex_column:
+            return None
+
+        reported_sex = row[0].get(self.reported_sex_column, None)
+
+        if reported_sex.lower() == 'female':
+            return 2
+        if reported_sex.lower() == 'male':
+            return 1
+        return 0
```
illusional, line 207: I think this should throw if it can't detect, not fall back to undefined - otherwise you might inadvertently lose data.

sample_metadata/parser/generic_parser.py
```diff
@@ -305,6 +305,18 @@ async def get_cpg_sample_id(self, row: SingleRow) -> Optional[str]:
...
+    def get_reported_sex(self, row: GroupedRow) -> Optional[int]:
+        """Get reported sex from grouped row"""
+
+    @abstractmethod
+    def get_reported_gender(self, row: GroupedRow) -> Optional[str]:
+        """Get reported gender from grouped row"""
+
+    @abstractmethod
+    def get_karyotype(self, row: GroupedRow) -> Optional[str]:
+        """Get karyotype from grouped row"""
```
illusional, line 318: Marking these as abstract requires the subclass to implement it, I think a default implementation of returning None should be fine. That will also stop the linting errors.

sample_metadata/parser/generic_parser.py
```diff
@@ -531,6 +548,14 @@ async def process_participant_group(
             'meta': collapsed_participant_meta.meta,
             'samples': samples_to_upsert,
         }
+
+        if reported_sex:
```
illusional, line 552: If reported_sex is `0` this will not work, maybe check for `is None`?

illusional (changes requested):

illusional:
I know it's annoying, but could you add a parser test to check that this works?

vivbak:
Thanks for calling out the lack of tests! Helped catch a few cases!

illusional (approved):
Nicely done! Tests LGTM!

illusional:
I've just merged dev in, so it can generate a coverage report - please don't merge until the codecov bot posts :)

PR #191, opened 2022-04-26
Fix failed backups 

vivbak:
Backups were failing with the error `error: failed to open the target stream for 'ib_logfile0'.`
This error was encountered previously and was handled by modifying permissions on the machine that runs `backup.py` i.e.
`sudo chmod g+r /var/lib/mysql/ibdata1`

However, the location of the `ib_logfile0`, has moved from `/var/lib/mysql/ibdata1`. Instead mariabackup outputs the file to the `targetdir` specified in the `mariabackup` call. 

In this PR, permissions to the `targetdir` are modified directly before `mariabackup` is run.


illusional (approved):
Thanks for fixing this!

PR #202, opened 2022-05-09
Fix Reads Metadata 

vivbak:
Previously, the location of the reads were not being appropriately parsed. This PR attempts to handle that.

### DIFF
--- a/sample_metadata/parser/generic_metadata_parser.py
+++ b/sample_metadata/parser/generic_metadata_parser.py
@@ -415,12 +415,11 @@ class GenericMetadataParser(GenericParser):
 
         return reduce(GenericMetadataParser.merge_dicts, dicts)
 
-    async def get_read_filenames(self, sample_id: str, row: GroupedRow) -> List[str]:
+    async def get_read_filenames(self, sample_id: str, row: SingleRow) -> List[str]:
         """Get paths to reads from a row"""
         read_filenames = []
-        for r in row if isinstance(row, list) else [row]:
-            if self.reads_column and self.reads_column in r:
-                read_filenames.extend(r[self.reads_column].split(','))
+        if self.reads_column and self.reads_column in row:
+            read_filenames.extend(row[self.reads_column].split(','))
 
         return read_filenames
 
@@ -487,6 +486,9 @@ class GenericMetadataParser(GenericParser):
         for r in rows:
             if self.reads_column and self.reads_column in r:
                 read_filenames.extend(r[self.reads_column].split(','))
+            if self.reads_column is None:
+                filenames = await self.get_read_filenames(sample_id=sample_id, row=r)
+                read_filenames.extend(filenames)
             if self.gvcf_column and self.gvcf_column in r:
                 gvcf_filenames.extend(r[self.gvcf_column].split(','))
 
--- a/sample_metadata/parser/generic_metadata_parser.py
+++ b/sample_metadata/parser/generic_metadata_parser.py
@@ -484,11 +484,8 @@ class GenericMetadataParser(GenericParser):
         read_filenames = []
         gvcf_filenames = []
         for r in rows:
-            if self.reads_column and self.reads_column in r:
-                read_filenames.extend(r[self.reads_column].split(','))
-            if self.reads_column is None:
-                filenames = await self.get_read_filenames(sample_id=sample_id, row=r)
-                read_filenames.extend(filenames)
+            filenames = await self.get_read_filenames(sample_id=sample_id, row=r)
+            read_filenames.extend(filenames)
             if self.gvcf_column and self.gvcf_column in r:
                 gvcf_filenames.extend(r[self.gvcf_column].split(','))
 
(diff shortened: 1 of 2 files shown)


vladsaveliev (approved):

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -487,6 +486,9 @@ async def get_sequence_meta(
         for r in rows:
             if self.reads_column and self.reads_column in r:
                 read_filenames.extend(r[self.reads_column].split(','))
+            if self.reads_column is None:
+                filenames = await self.get_read_filenames(sample_id=sample_id, row=r)
+                read_filenames.extend(filenames)
```
illusional, line 491: Can this just be called directly instead - this logic is better placed in that method anyway, then cleaner to override.

```suggestion
            reads = self.get_read_filenames(sample_id=sample_id, row=r)
            if reads:
                read_filenames.extend(read_filenames)
```

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -415,12 +415,11 @@ def prepare_dict_from_keys(key_parts: List[str], val):
 
         return reduce(GenericMetadataParser.merge_dicts, dicts)
 
-    async def get_read_filenames(self, sample_id: str, row: GroupedRow) -> List[str]:
+    async def get_read_filenames(self, sample_id: str, row: SingleRow) -> List[str]:
```
illusional, line 418: Was the GroupedRow / SingleRow definition wrong before hand? My thought would have been a GroupedRow, but as long as tests are passing it doesn't really matter

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -415,12 +415,11 @@ def prepare_dict_from_keys(key_parts: List[str], val):
...
         """Get paths to reads from a row"""
         read_filenames = []
-        for r in row if isinstance(row, list) else [row]:
-            if self.reads_column and self.reads_column in r:
-                read_filenames.extend(r[self.reads_column].split(','))
+        if self.reads_column and self.reads_column in row:
+            read_filenames.extend(row[self.reads_column].split(','))
 
         return read_filenames
```
illusional, line 424: Proposing to simplify this and do more processing

```python
        if not self.reads_column or not self.reads_column in row:
            return []
        # more post processing
        return [f.strip() for f in row[self.reads_column].split(',') if f.strip()]
```

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -485,8 +484,8 @@ async def get_sequence_meta(
         read_filenames = []
         gvcf_filenames = []
         for r in rows:
-            if self.reads_column and self.reads_column in r:
-                read_filenames.extend(r[self.reads_column].split(','))
+            filenames = await self.get_read_filenames(sample_id=sample_id, row=r)
+            read_filenames.extend(filenames)
```
illusional, line 488:
```suggestion
            read_filenames.extend(await self.get_read_filenames(sample_id=sample_id, row=r))
```

scripts/parse_prophecy.py
```diff
@@ -58,15 +58,15 @@ def fastq_file_name_to_sample_id(self, filename: str) -> str:
...
+    async def get_read_filenames(self, sample_id: str, row: SingleRow) -> List[str]:
         """
         We don't have fastq urls in a manifest, so overriding this method to take
         urls from a bucket listing.
         """
         return [
             path
             for filename, path in self.filename_map.items()
-            if self.fastq_file_name_to_sample_id(filename) == sample_id
+            if self.fastq_file_name_to_sample_id(filename) == row['Sample/Name']
```
illusional, line 69: This `row['Sample/Name']` should be defined as a constant somewhere - is it not the sample_id though?

illusional (changes requested):
Not fundamentally related to this change, but the ProphecyParser should define the sensible defaults in the __init__ method, not in the instantiation of the function: https://github.com/populationgenomics/sample-metadata/blob/3f8648eb099f6e4f0d5338d53c61825c62501a8e/scripts/parse_prophecy.py#L94-L104.

This should make a clearer pattern where the 'Sample/Name' constant is defined somewhere - or set on the class. Happy to chat more about this

scripts/parse_prophecy.py
```diff
@@ -15,33 +15,79 @@
...
+            Columns.PLATFORM,
+            Columns.CONCENTRATION,
+            Columns.VOLUME,
+        ]
+        return {k: COLUMN_MAP[k] for k in fields}
+
+    @staticmethod
+    def sample_meta_map():
+        """Columns that will be put into sample.meta"""
+        fields = [Columns.SEX]
```
illusional, line 67: Shouldn't Sex go on the participant?

illusional (approved):
Thank you, looks great! One small suggestion RE the sex field on samples

scripts/parse_prophecy.py
vivbak: Yes! Will fix.

vivbak: Actually, it's redundant. Sex is parsed anyway on the participant.

vivbak: Will scrap.

PR #212, opened 2022-05-19
Fixes incorrect regex parser + add test

illusional:


sample_metadata/parser/generic_parser.py
```diff
@@ -67,7 +67,7 @@
     + VCF_EXTENSIONS
 )
 
-rmatch = re.compile(r'[_\.-][Rr]?[12]')
+rmatch = re.compile(r'[_\.-][Rr]?[12]\.fastq(\.gz)?$')
```
vladsaveliev, line 70: Should we care about `.fq.gz`?

illusional: Addressed in latest commit @vladsaveliev, could you please confirm?

vladsaveliev: Awesome, thank you!

scripts/fix_fastqs.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Fix fastqs for all of MY projects
+    """
+    datasets = ProjectApi().get_my_projects()
+    for dataset in datasets:
+        await main(dataset)
+
+
+if __name__ == '__main__':
+    asyncio.new_event_loop().run_until_complete(for_all_projects())
```
vladsaveliev, line 79: Really enjoy your async code mastery. I should start using more of it too :)

vladsaveliev (approved):
Thank you for fixing it so quickly!

PR #214, opened 2022-05-20
Add ont + es-index analysis type

illusional:
+ size to crams

### DIFF
--- a/scripts/add_cram_size.py
+++ b/scripts/add_cram_size.py
@@ -1,0 +1,139 @@
+import logging
+import os
+import asyncio
+from itertools import groupby
+
+from google.cloud import storage
+from google.api_core.exceptions import NotFound
+
+from sample_metadata.apis import AnalysisApi
+from sample_metadata.model.analysis_query_model import AnalysisQueryModel
+from sample_metadata.model.analysis_update_model import AnalysisUpdateModel
+from sample_metadata.model.analysis_status import AnalysisStatus
+from sample_metadata.model.analysis_type import AnalysisType
+from sample_metadata.parser.generic_parser import chunk
+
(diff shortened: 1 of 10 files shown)


scripts/add_cram_size.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+
+    if not path.startswith("gs://"):
+        raise ValueError(f'path does not start with gs://: {path}')
+
+    return path[len("gs://") :].split("/", maxsplit=1)[0]
+
+
+async def process_all_projects():
+    pnames = [
```
violetbrina, line 36: probably missing a doc string

scripts/add_cram_size.py
```diff
@@ -0,0 +1,139 @@
...
+    ]
+    for pname in pnames:
+        await process_project(pname)
+
+    # projects = [{'name': 'ravenscroft-rdstudy'}] # await ProjectApi().get_all_projects_async()
+    # for pobj in projects:
+    #     pname = pobj['name']
+    #     if pname.endswith('test'):
+    #         continue
+    #     await process_project(pname)
```
violetbrina, line 61: Remove old code. That or maybe it's needed for later? In which case add a TODO

scripts/add_cram_size.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+async def process_project(project: str):
+    logger.info(f'{project} :: Begin processing')
```
violetbrina, line 65: Function doc string plz :)

violetbrina (changes requested):
Some linting errors that need to be cleaned up a bit. But functionally looks good!

violetbrina (approved):

PR #215, opened 2022-05-26
Fix sample grid for missing participants

illusional:
Failing because participant doesn't exist - meta doesn't exist which means the participant keys don't get filled - and except out.


vladsaveliev (approved):
Thanks!!

PR #218, opened 2022-05-27
Add parse ped script

vivbak:
A simple script to parse a ped file into SM. Created to by-pass permissions issues.

PR #222, opened 2022-05-30
Improve arbitrary sm script to support AnyPath

illusional:
@vivbak - a slightly different approach to #218 - but I like the idea of using AnyPath :)

You can run something like this with:

```shell
analysis-runner \
    --dataset $dataset --description "import $dataset pedigree" -o import-pedigree \
    python3 scripts/arbitrary_sm.py family import_pedigree \
    --json '{"project": "$dataset", "has_header": true, "file": "gs://cpg-$dataset-main-upload/pedigree.ped"}'
```


vivbak:
Nice!

vivbak (approved):

PR #224, opened 2022-05-31
Improve google identity token get

illusional:
Add check for internal google link + better Attribute error catch
Changes the configuration template which generates into useful code


violetbrina (approved):
Code looks good, more comments as to what the second try catch does would be nice

PR #236, opened 2022-06-20
Parser fixes

vladsaveliev:
A couple of fixes to the parser:

* `CloudHelper.populate_filename_map` referenced `self.filename_map` in the function that is used to initialise it, which led to an error;
* `CloudHelper._list_gcs_directory` didn't work with search prefixes ~deeper than the bucket root level~ on the bucket root level without trailing /, fixed it;
* In `GenericMetadataParser.check_files_covered_by_rows` the order of sets to determine missing/extra files seemed to be swapped. Also, not sure about the extension list there, should it import `generic_parser.FASTQ_EXTENSIONS`, `generic_parser.BAM_EXTENSIONS`, etc?

### DIFF
--- a/sample_metadata/parser/cloudhelper.py
+++ b/sample_metadata/parser/cloudhelper.py
@@ -99,7 +99,7 @@ class CloudHelper:
     async def file_size(self, filename) -> int:
         """Get size of file in bytes"""
         return AnyPath(self.file_path(filename)).stat().st_size
-    
+
     def populate_filename_map(
         self, search_locations: list[str]
     ) -> dict[str, str]:
@@ -116,8 +116,7 @@ class CloudHelper:
                 file_base = os.path.basename(file)
                 if file_base in fn_map:
                     logging.warning(
-                        f'File "{file}" from "{directory}" already exists in '
-                        f'directory map: {fn_map[file_base]}'
+                        f'File "{file}" from "{directory}" already exists in directory map: {fn_map[file_base]}'
                     )
                     continue
                 fn_map[file_base] = file
--- a/sample_metadata/parser/cloudhelper.py
+++ b/sample_metadata/parser/cloudhelper.py
@@ -108,7 +108,7 @@ class CloudHelper:
         so let's prepopulate that filename_map from the search_locations!
         """
 
-        fn_map = {}
+        fn_map: dict[str, str] = {}
         for directory in search_locations:
             directory_list = self.list_directory(directory)
             for file in directory_list:
--- a/sample_metadata/parser/cloudhelper.py
+++ b/sample_metadata/parser/cloudhelper.py
@@ -2,7 +2,7 @@
 import os
 import logging
 
-from cloudpathlib import AnyPath
+from cloudpathlib import AnyPath, GSPath
 from google.cloud import storage
 
 
@@ -19,7 +19,8 @@ class CloudHelper:
 
         self.search_paths = search_paths or []
 
-        self.filename_map: dict[str, str] = self.populate_filename_map(self.search_paths)
+        self.filename_map: dict[str, str] = {}
+        self.populate_filename_map(self.search_paths)
         # pylint: disable
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -303,15 +303,15 @@ async def check_files_covered_by_rows(
 
         filenames: List[str] = sum(await asyncio.gather(*filename_promises), [])
         fs = set(f.strip() for f in filenames if f and f.strip())
-        relevant_extensions = ('.cram', '.fastq.gz', '.bam')
+        relevant_extensions = ('.cram', '.crai', '.fq.gz', '.fastq.gz', '.bam', '.bai')
```
illusional, line 306: Oh you want to tell if an index is missing too? We don't require indexes for other ingested data, so happy for this change if you can parameterise it.

sample_metadata/parser/cloudhelper.py
```diff
@@ -155,7 +156,12 @@ async def get_gcs_blob(self, filename: str) -> storage.Blob:
 
     def _list_gcs_directory(self, gcs_path) -> list[str]:
 
-        bucket_name, remaining_path = gcs_path[5:].split('/', maxsplit=1)
-        bucket = self.get_gcs_bucket(bucket_name)
-        blobs = self.gcs_client.list_blobs(bucket, prefix=remaining_path, delimiter='/')
-        return [f'gs://{bucket_name}/{blob.name}' for blob in blobs]
+        path = GSPath(gcs_path)
```
illusional, line 159: I'm not really following the change here, is it mostly just the trailing `/` on the path? The only main change I see is the change in the `remaining_path`.

violetbrina (approved):

sample_metadata/parser/cloudhelper.py
```diff
@@ -108,20 +109,20 @@ def populate_filename_map(
         so let's prepopulate that filename_map from the search_locations!
         """
 
-        fn_map = {}
         for directory in search_locations:
```
illusional, line 112: I made this change because it would mean overriding this parser means you shouldn't have to know about state (and makes it much easier to mock) in tests. Curious about the errors you're seeing here though.

illusional (changes requested):

sample_metadata/parser/generic_metadata_parser.py
vladsaveliev: Ah, I guess we don't have to check them at all. I initially thought the error was because the index is not listed here, but it was because of the sets logic below. I will remove bai/crai.

sample_metadata/parser/cloudhelper.py
vladsaveliev: The line with `splis` just didn't work:
```
    bucket_name, remaining_path = gcs_path[5:].split('/', maxsplit=1)
ValueError: not enough values to unpack (expected 2, got 1)
```

sample_metadata/parser/cloudhelper.py
vladsaveliev: Line 119, 
```
logging.warning(
                        f'File "{file}" from "{directory}" already exists in directory map: {self.filename_map[file_base]}'
)
```
References `self.filename_map` which is not initialised. Alternatively we could make the method static.

illusional: Ah sorry I see, yes please if you could make it static that would be great.

sample_metadata/parser/cloudhelper.py
illusional: I see, cloudpathlib has to be installed for other reasons so this seems reasonable to me

sample_metadata/parser/cloudhelper.py
vladsaveliev: What about `self.list_directory(directory)` call? It would involve more refactoring.

vladsaveliev: Btw, feel free to commit to the PR!

vladsaveliev: I guess can make it pseudo-static, keeping the `self.list_directory` call

illusional (approved):

PR #238, opened 2022-06-21
Update analysis endpoints

illusional:
Previously `sync-seqr`, but moved that to a different PR as I fixed a lot of programmatic access as part of that. This PR addresses a few things:

- Removes `StreamingResponse` as a class, OpenApiGenerator isn't handling this, and the response is always None
- Speeds up the actual StreamingResponse (from many seconds to <100ms) by returning the whole file at once (fine because it's all in memory), rather than streaming it letter by letter
- More consistently uses exportType and guessing the delimiter of uploaded files
- Fixes queries on some endpoints (+ improves analysis-runner querying)


MattWellie:
@illusional with the export_type being removed from `{project}/individual-metadata-seqr`, is it now only capable of serving JSON unless the method is imported and called in code? 

Just asking as the majority of that method now seems inaccessible if the export type can't be altered

MattWellie (approved):
LGTM, happy to wait for a better informed reviewer

illusional:
Hey @MattWellie, github hides it by default, but export_type actually remains as a param, it's just not embedded in the URL path, but gets moved to a query param, so no change to the API, but old APIs might not work:

image

scripts/create_test_subset.py
```diff
@@ -255,12 +254,12 @@ def transfer_ped(initial_project, target_project, family_ids):
...
     ped_tsv = fapi.get_pedigree(
         initial_project,
-        response_type=ContentType('tsv'),
+        export_type='tsv',
         internal_family_ids=family_ids,
     )
     ped_json = fapi.get_pedigree(
         initial_project,
-        response_type=ContentType('json'),
+        export_type='tsv',
```
violetbrina, line 262: Should this be 'json'?

violetbrina (approved):

PR #242, opened 2022-06-26
Add reference assembly as requirement for crams

illusional:
Follow up steps to make production pipeline know to use this reference.

### DIFF
--- a/sample_metadata/parser/cloudhelper.py
+++ b/sample_metadata/parser/cloudhelper.py
@@ -8,6 +8,7 @@ from google.cloud import storage
 
 class CloudHelper:
     """General CloudHelper for parsers"""
+
     LOCAL_PREFIX = '/'
     GCS_PREFIX = 'gs://'
     AZ_PREFIX = 'az://'
@@ -19,7 +20,9 @@ class CloudHelper:
 
         self.search_paths = search_paths or []
 
-        self.filename_map: dict[str, str] = self.populate_filename_map(self.search_paths)
+        self.filename_map: dict[str, str] = self.populate_filename_map(
+            self.search_paths
+        )
         # pylint: disable
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


sample_metadata/parser/cloudhelper.py
```diff
@@ -89,7 +92,7 @@ def list_directory(self, directory_name) -> list[str]:
     async def file_contents(self, filename) -> str | None:
         """Get contents of file (decoded as utf8)"""
         path = self.file_path(filename)
-        with AnyPath(path).open() as f:
+        with AnyPath(path).open(encoding='utf-8-sig') as f:
```
lgruen, line 95: Interesting -- why `utf-8-sig` here instead of `utf-8`? According to https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-without-bom, a BOM isn't generally recommended.

sample_metadata/parser/generic_metadata_parser.py
```diff
@@ -555,6 +566,7 @@ async def get_sequence_meta(
         read_filenames: List[str] = []
         gvcf_filenames: List[str] = []
         read_checksums: List[str] = []
+        reference_fastas: set[str] = set()
```
lgruen, line 569: Should this be called reference assembly as well, for consistency? (Or the other way around.)

sample_metadata/parser/cloudhelper.py
illusional: Specifically the OHMR4 template has a BOM in it - and I don't think it breaks the other way. I don't have a good way to detect it when reading and it causes issues with forming the key map.

sample_metadata/parser/generic_metadata_parser.py
illusional: Yes, absolutely :)

violetbrina (approved):

PR #252, opened 2022-07-01
Handle case where sample doesn't exist in metamist

vivbak:
At present, update_sequence_from_external_id_and_type does not handle the case where the sample does not yet exist in metamist. 

FYI @violetbrina

### DIFF
--- a/db/python/layers/sequence.py
+++ b/db/python/layers/sequence.py
@@ -316,7 +316,12 @@ class SampleSequenceLayer(BaseLayer):
         return sequence_id
 
     async def update_sequence_from_external_id_and_type(
-        self, external_sample_id, sequence_type, status, meta
+        self,
+        external_sample_id,
+        sequence_type,
+        status,
+        meta,
+        sample_type,
     ):
         """Update a sequence from the external_id and sequence_type"""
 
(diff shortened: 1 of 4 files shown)


vivbak:
Ah sorry pushed the wrong file somehow

illusional (approved):

db/python/layers/sequence.py
```diff
@@ -316,22 +316,44 @@ async def update_sequence_from_sample_and_type(
         return sequence_id
 
     async def update_sequence_from_external_id_and_type(
```
illusional, line 318: Sorry, another nitpick, can you rename this to `upsert_sequence...`

illusional (approved):

db/python/layers/sequence.py
vivbak: Done!

test/test_sequence.py
```diff
@@ -149,9 +149,44 @@ async def test_update_sequence_from_external_id_and_type(self):
...
+        statuses = [
+            'received',
+            'sent-to-sequencing',
+            'completed-sequencing',
+            'completed-qc',
+            'failed-qc',
+            'uploaded',
+            'unknown',
+        ]
+        status = random.choice(statuses)
```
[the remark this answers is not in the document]
vivbak: Switched it

PR #256, opened 2022-07-14
Clean Up Script 

vivbak:
Script to clean up sequencing data/metadata from cloud storage and metamist.


vivbak:
Hey hey @illusional, could you take a look and let me know if this is the general approach you were thinking?  Will clean it up if so. Otherwise happy to take this in another direction. TIA 🙏

scripts/clean_up.py
```diff
@@ -0,0 +1,121 @@
...
+def clean_up_metamist(
+    sequence_ids_internal: list[str],
+    clean_up_location: str,
+):
+    """Deleting sequencing entires for specified sequences in metamist
+    This will delete all of the reads for the latest sequence."""
+    all_locations: list[CloudPath] = []
+    sequence_ids: list[int] = []
+    for sequence_id in sequence_ids_internal:
+        current_sequence = seqapi.get_sequence_by_id(sequence_id)
```
illusional, line 31: Should get all sequences at once by IDs outside the loop, otherwise this will be very slow.

scripts/clean_up.py
```diff
@@ -0,0 +1,121 @@
...
+        reads = current_sequence['meta']['reads']
```
illusional, line 32: Nitpick: sequences are not guaranteed to have a `'reads'` value on the meta.

scripts/clean_up.py
```diff
@@ -0,0 +1,121 @@
...
+        current_read_locations: list[CloudPath] = []
+        # Skip delete operations if there are no reads for the given sequence
+        delete_reads = bool(reads)
+        for read in reads:
```
illusional, line 36: Reads can be both crams/bams and fastqs and they have a slightly different structure.

scripts/clean_up.py
```diff
@@ -0,0 +1,121 @@
...
+            for strand in read:
+                cloud_path = strand['location']
+                read_blob = CloudPath(cloud_path)
+                blob_location = read_blob.anchor + read_blob.drive
+                # Validates the file that we want to delete, is in the bucket specified.
+                if read_blob.exists() and blob_location == clean_up_location:
```
illusional, line 42: Could this more simply be: `cloud_path.startswith(clean_up_location)`?

scripts/clean_up.py
```diff
@@ -0,0 +1,121 @@
...
+                    current_read_locations.append(read_blob)
+                else:
+                    logging.error(
+                        f'{read_blob.name} was not found in {clean_up_location}. Skipping delete of {sequence_id} from metamist'
+                    )
+                    # Skip delete operations, something might be wrong with the reads meta locations.
+                    delete_reads = False
+                    continue
+        # Clear reads metadata, including location, basename, class, checksum & size
+        if delete_reads:
```
illusional, line 52: You'll want to clear the buckets before the sequence update, to ensure we don't fail between that process.

illusional (commented):
Looks on the right track to me! I've left a few initial comments:

- Getting all values from server up-front to avoid HTTP calls in loops
- I really like getting a full list of locations to clear, could we add a confirm dialogue (or maybe a dry run feature) to check out what files are going to be deleted before they do (and have some --force flag to skip when running on batch)
- We should clear the files, then update metamist to avoid accidentally falling out of sync
- Nice work with the "clear all the sequences or none" if any location falls outside the 'location to delete from' :)

scripts/clean_up.py
vivbak: Done!

scripts/clean_up.py
vivbak: switched to a get :)

scripts/clean_up.py
vivbak: Added a check for now so we only support fastqs. Can expand this in future if there is a use case.

scripts/clean_up.py
vivbak: have moved things around accordingly

scripts/clean_up.py
vivbak: good point!

illusional (approved):

PR #265, opened 2022-08-06
Ingest thousand-genomes/hgdp metadata

vladsavelyev:
Process 1KG, HGDP, SGDP population labels, sex, pedigree, ingest into `thousand-genomes` and `hgdp` datasets.


lgruen (changes requested):
Should we schedule a walkthrough with the population analysis team for this?

PR #273, opened 2022-08-26
Per sample size endpoint

violetbrina:
Endpoint to return the per project, per sample size for each sequence type over a set date range.


README.md
```diff
@@ -240,25 +243,28 @@ Some handy links:
...
+  - We generate the sample_metadata installable Python API based on this schema.
 
 #### Generating the installable API
 
 The installable API is automatically generated through the `package.yml` GitHub action and uploaded to PyPI.
 
-
 To generate the python api you'll need to install openapi generator v5.x.x
 
+Follow [this tutorial](https://openapi-generator.tech/docs/installation/) for
```
illusional, line 259: Can you list the latest 5.x.x release here, and move instructions to download that release to here?

api/routes/analysis.py
```diff
@@ -309,3 +320,67 @@ async def get_sample_reads_map_for_seqr(
         media_type='text/tab-separated-values',
         headers={'Content-Disposition': f'filename={basefn}.tsv'},
     )
+
+
+@router.post('/sample-file-sizes', operation_id='getSampleFileSizes')
```
illusional, line 325: I think the `project_names` will never be large enough to exceed that max query param length, could we make this a get request?

api/routes/analysis.py
```diff
@@ -309,3 +320,67 @@ async def get_sample_reads_map_for_seqr(
...
+) -> list[ProjectSizeModel]:
+    """
+    Get the per sample file size by type over the given projects and date range
+    """
+    atable = AnalysisLayer(connection)
+    stable = SampleLayer(connection)
+
+    # Check access to projects
+    project_ids = None
+    if project_names:
```
illusional, line 340: Should project_names a required field, and can we raise a 400 if no projects are specified?

api/routes/analysis.py
```diff
@@ -309,3 +320,67 @@ async def get_sample_reads_map_for_seqr(
...
+        pt = ProjectPermissionsTable(connection=connection.connection)
+        project_ids = await pt.get_project_ids_from_names_and_user(
+            connection.author, project_names, readonly=True
+        )
+
+    prj_name_map = dict(zip(project_ids, project_names))
+
+    # Try to convert dates
+    start, end = None, None
+    if start_date or end_date:
```
illusional, line 350: You don't need this first `if` conditional. Could we move the date parsing logic so somewhere central?

api/routes/analysis.py
```diff
@@ -309,3 +320,67 @@ async def get_sample_reads_map_for_seqr(
...
+
+            if end_date:
+                end = datetime.strptime(end_date, '%Y-%m-%d').date()
+        except Exception as excep:
+            raise ValueError(
+                f'Start or end date could not be converted: {start_date} {end_date}'
+            ) from excep
+
+    # Get samples from pids
+    samples = await stable.get_samples_by(project_ids=project_ids)
```
illusional, line 363: This logic should be moved to the AnalysisLayer so it's potentially reusable.

illusional: If the only result is just getting active sample_ids (as internal ints), we should add a function to do that to avoid `int -> Sample.id: str -> int`.

api/routes/sample.py
```diff
@@ -233,3 +233,15 @@ async def get_history_of_sample(
     result = await st.get_history_of_sample(internal_id)
 
     return result
+
+
+@router.post('/created', operation_id='getSampleCreateDate')
+async def get_sample_create_date(
```
illusional, line 239:
```suggestion
async def get_samples_create_date(
```

api/routes/sample.py
```diff
@@ -233,3 +233,15 @@ async def get_history_of_sample(
     result = await st.get_history_of_sample(internal_id)
 
     return result
+
+
+@router.post('/created', operation_id='getSampleCreateDate')
```
illusional, line 238:
```suggestion
@router.post('/created', operation_id='getSamplesCreateDate')
```

illusional: Could we expand the url path to less likely to clash later?

db/python/layers/analysis.py
```diff
@@ -132,6 +138,80 @@ async def query_analysis(
...
+
+        # Manual filtering to find the most recent analysis cram of each sequence type
+        # for each sample
+        for cram in crams:
+            sids = cram.sample_ids
+            seqtype = cram.meta.get('sequence_type')
+            seqtype = seqtype if seqtype else cram.meta.get('sequencing_type')
+            size = cram.meta.get('size')
+
+            if len(sids) == 1 and not isinstance(seqtype, list) and seqtype and size:
```
illusional, line 187: Should raise an exception if this isn't the case (except maybe seqtype`).

db/python/layers/analysis.py
```diff
@@ -132,6 +138,80 @@ async def query_analysis(
...
+                crams_by_sid[sid][seqtype].append(cram)
+
+        # Format output
+        result: dict[int, list] = defaultdict(list)
+        for sid in use_samples:
+            sample_crams: dict[SequenceType, int] = defaultdict(int)
+            for seqtype in crams_by_sid[sid]:
+                sequence_crams = sorted(
+                    crams_by_sid[sid][seqtype],
+                    key=lambda x: datetime.fromisoformat((x.timestamp_completed)),
```
illusional, line 199:
```suggestion
                    key=lambda x: datetime.fromisoformat(x.timestamp_completed),
```

db/python/layers/sample.py
```diff
@@ -150,6 +150,10 @@ async def get_sample_with_missing_participants_by_internal_id(
         m = await self.st.get_sample_with_missing_participants_by_internal_id(project)
         return dict(m)
 
+    async def get_sample_create_date(self, sample_ids: List[int]):
+        """Get a map of {internal_sample_id: date_created} for list of sample_ids"""
+        return await self.st.get_sample_create_date(sample_ids)
```
illusional, line 155: Will need to do access checks to see if user has access to sample_ids.

illusional (changes requested):
This is looking good! A few comments + can you add tests to cover each of the new functionality?

README.md
violetbrina: Sure thing

api/routes/analysis.py
violetbrina: Unless there's a way to do this that I don't know, when you have an array input FastApi automatically makes it part of the request_body. It then throws an error that you cannot have a request_body in a get request.

api/routes/analysis.py
violetbrina: Sure thing!

api/routes/sample.py
violetbrina: Sure

db/python/layers/sample.py
violetbrina: Done

api/routes/analysis.py
violetbrina: Ah figured it out

db/python/layers/analysis.py
violetbrina: But aren't there analyses that have multiple sample ids in the db? If that's the case I was simply going to ignore them quietly since there is no straight forward way to determine the file sizes.

db/python/tables/sample.py
```diff
@@ -402,6 +417,12 @@ async def get_sample_with_missing_participants_by_internal_id(
         )
         return {row['id']: row['external_id'] for row in rows}
 
+    async def get_samples_create_date(self, sample_ids: list[int]):
+        """Get a map of {internal_sample_id: date_created} for list of sample_ids"""
+        _query = 'SELECT id, min(row_start) FROM sample FOR SYSTEM_TIME ALL WHERE id in :sids GROUP BY id'
+        rows = await self.connection.fetch_all(_query, {'sids': sample_ids})
+        return {r[0]: str(r[1].date()) for r in rows}
```
illusional, line 424: Should this just return the date object without casting it back to string, the usage in the next casts this back to a date.

db/python/tables/sample.py
```diff
@@ -402,6 +417,12 @@ async def get_sample_with_missing_participants_by_internal_id(
         )
         return {row['id']: row['external_id'] for row in rows}
 
+    async def get_samples_create_date(self, sample_ids: list[int]):
+        """Get a map of {internal_sample_id: date_created} for list of sample_ids"""
+        _query = 'SELECT id, min(row_start) FROM sample FOR SYSTEM_TIME ALL WHERE id in :sids GROUP BY id'
```
illusional, line 422: Just prompting to add this so empty list doesn't cause an exception.

```suggestion
        if len(sample_ids) == 0:
            return []
        _query = 'SELECT id, min(row_start) FROM sample FOR SYSTEM_TIME ALL WHERE id in :sids GROUP BY id'
```

regenerate_api.py
```diff
@@ -88,7 +88,7 @@ def check_openapi_version():
     """
     Check compatible OpenAPI version
     """
-    command = [*OPENAPI_COMMAND, '--version']
+    command = [*OPENAPI_COMMAND, 'version']
```
illusional, line 91: This will cause issues in the deploy because it's using an older version with `--version` - can we support both using a loop or update the deploy instructions.

test/test_analysis.py
```diff
@@ -13,27 +17,28 @@ class TestAnalysis(DbIsolatedTest):
...
+            sample_ids=[self.sample_id],
+            meta={'sequence_type': 'genome', 'size': 1024},
+        )
+
     @run_test_as_sync
     async def test_add_cram(self):
         """
         Test adding an analysis of type CRAM
         """
-        await self.al.insert_analysis(
```
illusional, line 31: Hmm, I think if this behaviour changes and the test fails it will be tricky to narrow down, I'd be more in favour of moving the insert sample into each of these tests (as well as keeping the `insert_analysis` into each test.

test/test_analysis.py
```diff
@@ -45,3 +50,54 @@ async def test_add_cram(self):
...
+                                'size': {SequenceType.GENOME: 1024},
+                            }
+                        ],
+                    }
+                ],
+            }
+        ]
+
+        # Check output is formatted correctly
+        self.assertDictEqual(expected[0], result[0])
```
illusional, line 80:
```suggestion
        self.assertEqual(1, len(result))
        self.assertDictEqual(expected[0], result[0])
```

test/test_analysis.py
```diff
@@ -45,3 +50,54 @@ async def test_add_cram(self):
 
         self.assertEqual(1, analysis_samples[0]['sample_id'])
         self.assertEqual(analyses[0]['id'], analysis_samples[0]['analysis_id'])
+
+    @run_test_as_sync
+    async def test_get_sample_file_sizes(self):
```
illusional, line 55: Could you add tests to cover the additional cases:

- Multiple CRAMs of the same sequencing type
- CRAMs from multiple samples

test/test_analysis.py
```diff
@@ -45,3 +50,54 @@ async def test_add_cram(self):
...
+        self.assertDictEqual(expected[0], result[0])
+
+        # Assert that if we select a date range outside of sample creation date
+        # that is doesn't show up in the map
+        yesterday = date.today() - timedelta(days=1)
+        result = await self.al.get_sample_file_sizes(
+            project_ids=[1], end_date=yesterday
+        )
+
+        self.assertEqual([], [])
```
illusional, line 103:
```suggestion
```

illusional (changes requested):
Looking really great! I have a few nitpicks + commentary of testing

api/routes/sample.py
```diff
@@ -233,3 +233,23 @@ async def get_history_of_sample(
     result = await st.get_history_of_sample(internal_id)
 
     return result
+
+
+@router.get('/dateofcreation', operation_id='getSamplesCreateDate')
```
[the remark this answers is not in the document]
violetbrina: Done

api/routes/sample.py
```diff
@@ -233,3 +233,23 @@ async def get_history_of_sample(
...
+async def get_samples_create_date(
+    sample_ids: List[str] = Query(None),
```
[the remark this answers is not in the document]
violetbrina: Done

api/routes/sample.py
```diff
@@ -233,3 +233,23 @@ async def get_history_of_sample(
...
+    pt = ProjectPermissionsTable(connection.connection)
+    st = SampleLayer(connection)
+
+    # Check access permissions
+    sample_ids_raw = sample_id_transform_to_raw_list(sample_ids) if sample_ids else None
+    pjcts = await st.get_project_ids_for_sample_ids(sample_ids_raw)
+    await pt.check_access_to_project_ids(connection.author, pjcts, readonly=True)
+
+    # Convert to raw ids and query the start dates for all of them
+    result = await st.get_samples_create_date(sample_ids_raw)
```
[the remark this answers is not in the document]
violetbrina: Done

api/utils/dates.py
```diff
@@ -0,0 +1,12 @@
...
+
+def strtodate(d: str | None) -> date | None:
+    """Convert date string to date, allow for None"""
+    if d:
+        try:
+            return datetime.strptime(d, '%Y-%m-%d').date()
+        except Exception as excep:
+            raise ValueError(f'Date could not be converted: {d}') from excep
+
+    return None
```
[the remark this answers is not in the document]
violetbrina: Done

api/utils/dates.py
```diff
@@ -0,0 +1,12 @@
+from datetime import datetime, date
+
+
+def strtodate(d: str | None) -> date | None:
```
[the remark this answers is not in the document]
violetbrina: Done

db/python/tables/sample.py
violetbrina: Done

db/python/tables/sample.py
violetbrina: True...

regenerate_api.py
violetbrina: Instructions updated, both supported.

test/test_analysis.py
violetbrina: Done

test/test_analysis.py
violetbrina: Done

regenerate_api.py
```diff
@@ -88,8 +88,18 @@ def check_openapi_version():
...
-    command = [*OPENAPI_COMMAND, '--version']
-    out = subprocess.check_output(command).decode().split('\n', maxsplit=1)[0].strip()
+    cmds = ['--version', 'version']
+
+    for cmd in cmds:
+        command = [*OPENAPI_COMMAND, cmd]
+
+        try:
+            out = subprocess.check_output(command)
+        except Exception:  # pylint: disable=broad-except
```
illusional, line 98:
```suggestion
        except subprocess.CalledProcessError:
```

regenerate_api.py
```diff
@@ -25,7 +25,7 @@
 def check_if_server_is_accessible() -> bool:
     """Check if request to 'SCHEMA_URL' returns OK"""
     try:
-        return requests.get(SCHEMA_URL).ok
+        return requests.get(SCHEMA_URL, timeout=None).ok
```
illusional, line 28:
```suggestion
        return requests.get(SCHEMA_URL, timeout=30).ok
```

db/python/tables/sample.py
```diff
@@ -402,6 +417,14 @@ async def get_sample_with_missing_participants_by_internal_id(
         )
         return {row['id']: row['external_id'] for row in rows}
 
+    async def get_samples_create_date(self, sample_ids: list[int]):
```
illusional, line 420: Nitpick type annotation:

Might need to import datetime for this annotation to work though
```suggestion
    async def get_samples_create_date(self, sample_ids: list[int]) -> dict[int, date]:
```

db/python/layers/analysis.py
illusional: But you're filtering for crams only - there shouldn't be any crams with more than one sample ID right? I guess technically possible, but If that happens, sounds like we've made a mistake. For now could this script collect all the cases it encounters, and logs them once after the for loop.

If possible to turn this into an early return, eg:

```python
if len(sids) > 1:
    affected_analyses.append(cram['id')
    continue
# other stuff
```

illusional (approved):
Thanks for putting this together, it looks great!! I've left a few nitpick comments, could you address before merging but no need for re-review.

PR #274, opened 2022-08-29
Simplify fetching of ID tokens

lgruen:


lgruen:
Unfortunately doesn't look like it's that simple after all: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1661916709755859

PR #281, opened 2022-09-11
Add sql script to delete metamist project

illusional:
Currently has to be run manually on the server.


scripts/deleteproject.sql
```diff
@@ -0,0 +1,14 @@
...
+SELECT @project := id FROM project WHERE name = '<dataset>';
+
+DELETE FROM participant_phenotypes where participant_id IN (SELECT id FROM participant WHERE project = @project);
+DELETE FROM family_participant WHERE family_id IN (SELECT id FROM family where project = @project);
+DELETE FROM family WHERE project = @project;
+DELETE FROM sample_sequencing WHERE sample_id in (SELECT id FROM sample WHERE project = @project);
+DELETE FROM analysis_sample WHERE sample_id in (SELECT id FROM sample WHERE project = @project);
+DELETE FROM analysis_sample WHERE analysis_id in (SELECT id FROM analysis WHERE project = @project);
+DELETE FROM participant WHERE project = @project;
+DELETE FROM sample WHERE project = @project;
```
vivbak, line 12: I know this is still in a draft state, but in case it's helpful, I think these lines should be swapped because of foreign key stuff. 
```suggestion
DELETE FROM sample WHERE project = @project;
DELETE FROM participant WHERE project = @project;
```

PR #283, opened 2022-09-16
Add functionality to sync seqr project within metamist UI

illusional:
UI:

image

I'm hoping to keep the `sync_seqr.py` script for a bit longer as a backup, and remove if this is all going well :)
@danielreti, could you review the React side please 🙏


danielreti (approved):
Nice loading ducks update! React looks good

db/python/layers/seqr.py
```diff
@@ -0,0 +1,525 @@
...
+from db.python.layers.participant import ParticipantLayer
+from db.python.layers.sequence import SampleSequenceLayer
+from db.python.tables.project import ProjectPermissionsTable
+from db.python.utils import ProjectId
+from models.enums import SequenceType, AnalysisStatus, AnalysisType
+from models.models.sample import sample_id_format
+
+# literally the most temporary thing ever, but for complete
+# automation need to have sample inclusion / exclusion
+SAMPLES_TO_IGNORE = {22735, 22739}
```
lgruen, line 28: Could we use Metamist's concept of "disabled" samples for now?

api/server.py
```diff
@@ -115,11 +115,30 @@ async def exception_handler(_: Request, e: Exception):
...
+    response = JSONResponse(
         status_code=code,
         content=base_params,
     )
 
+    # https://github.com/tiangolo/fastapi/issues/457#issuecomment-851547205
+    # FastAPI doesn't run middleware on exception, but if we make a non-GET/INFO
+    # request, then we lose CORS and hence lose the exception in the body of the
+    # response. Grab it manually, and explicitly allow origin if so.
+    middlewares = [m for m in app.user_middleware if isinstance(m, CORSMiddleware)]
```
lgruen, line 127: I'm not entirely sure why this is necessary / which case this is addressing. It's only internal infrastructure, so we don't need to be super strict in terms of security, but might maybe still be worth asking someone like Tommy for a review of this part?

lgruen (approved):

db/python/layers/seqr.py
illusional: We don't really have it, it's really poor and inconsistently implemented. Would be good to overhaul as part of upcoming refactors :)

api/server.py
illusional: We could always set it to always? You're sort of prone to scripts on external sites hitting metamist API without the user knowing. It's been working well so far, just unfortunate that the exception handler doesn't run the remaining middleware and hits the CORS middleware (and hence blocks the content body).

PR #286, opened 2022-09-27
Move script from seqr-private to metamist

illusional:
Moved from https://github.com/populationgenomics/seqr-private/pull/18 back here.


lgruen (approved):

lgruen:
Should this be mentioned somewhere in the top-level README?

Issue #139, opened 2022-02-16
Multiple external IDs for samples and participants


illusional:
We interact with samples that potentially have more than one identifier that it's referred by (potentially by multiple collaborators).

I think it's probably enough to add a `sample_external_id` table with `project, sample_id, external_id, name, meta`.

The real challenge to this card is determining all of the external places we use an external_id, and working out the process going forward. Do we need to have the concept of a primary external_id, or can that be avoided with enough refactoring - does it even make sense? 

What consequences does this have, does it make mapping back to one external ID more difficult, eg: 
- because you need to know the name of a set of external IDs
- because you might not always know external IDs for all samples
- Do people want to do analysis only on a subset of data for which a specific set of external IDs are available?
- \+ probably more.

2023-10-02 notes:

- leave the externalId field there, but have it populated as the "primary" external ID.
- Need some mechanism for setting the primary external ID on these entities
- Need some schema change in `project.xml`

jmarshall:
Draft PR is now #659.

PR #288, opened 2022-10-04
Support for multiple external ids for a sequence + sequence API revisit

illusional:
BREAKING / backwards incompatible changes.

- Will require DB migration
- Will require updates to pipelines packages.
- Removes all `latest_sequence*` endpoints
- Still to update / add tests
- External ID key names are case insensitive, and are NOT required for a sequence.

Still to complete audit of removed endpoints.

This is sort of the model for how #139 will be applied to participants + samples, eg:

```jsonc
{
  "id": 1,
  "external_ids": {
    "kccg": "KCCGX01",
    "vcgs": "VG_01"
  },
  // [... other fields]
}
```

Keen to get initial feedback @vivbak

Actioned:

- [ ] cpg-utils: https://github.com/populationgenomics/cpg-utils/pull/98
- [ ] joint-calling: https://github.com/populationgenomics/joint-calling/pull/92
- [ ] ~hail-elasticsearch-pipelines~ (not actioning, has been moved to [vladsavelyev/hail-elasticsearch-pipelines](https://github.com/vladsavelyev/hail-elasticsearch-pipelines) and archived)


vivbak:
Looks good @illusional! 
The only question floating about my head is whether this (all sequences having unique external identifiers) is a valid assumption. I suppose they would need to, given they are all stored as uniquely named files. 

I'm also trying to think about the whether asking sequencing providers to hit that cloud function on a sequence level vs a sample level changes anything majorly for them. Do you think it's worth touching base to check?

illusional:
This doesn't enforce that all sequences MUST have an external ID. Just offers it as an option to sequence type being the rough identifier. Potentially, this will be more important for the SequenceGroup proposal - currently it says "update_sample" but we really mean "update sequence of type for sample" - but if it's available, we'd love you to tell us information about the sequence (as that's really what we're wanting).

We give providers the sample, they give us sequencing - we don't really want information about the sample, just status about how the sequencing is going.

api/routes/sequence.py
```diff
@@ -64,59 +63,62 @@ async def update_sequence(
...
-    sequence_id = await sequence_layer.upsert_sequence_from_external_id_and_type(
-        external_sample_id=external_sample_id,
-        sequence_type=sequence_type,
-        status=sequence.status,
-        meta=sequence.meta,
-        sample_type=sample_type,
-    )
-
-    return sequence_id
+# @router.patch(
```
violetbrina, line 75: Did you want to leave this commented out or just delete it?

api/routes/sequence.py
```diff
@@ -190,82 +206,81 @@ async def get_sequences_by_internal_sample_ids(
...
+# async def get_latest_sequence_ids_from_sample_ids(
+#     sample_ids: List[str], connection: Connection = get_projectless_db_connection
+# ) -> Dict[str, int]:
+#     """Get sequence ids from internal sample ids"""
+#     sequence_layer = SampleSequenceLayer(connection)
+#     sample_ids_raw = sample_id_transform_to_raw_list(sample_ids)
+#     sequence_id_map = await sequence_layer.get_latest_sequence_ids_for_sample_ids(
+#         sample_ids_raw
+#     )
+#     return {sample_id_format(k): v for k, v in sequence_id_map.items()}
```
violetbrina, line 286: Again, keep or remove?

test/testbase.py
```diff
@@ -31,6 +31,17 @@
 nest_asyncio.apply()
 
 
+for lname in (
+    'asyncio',
+    'urllib3',
+    'docker',
+    # 'databases',
```
violetbrina, line 38: Remove?

test/testbase.py
```diff
@@ -78,7 +89,7 @@ async def setup():
             try:
                 os.environ['SM_ALLOWALLACCESS'] = '1'
                 set_full_access(True)
-                db = MySqlContainer('mariadb:latest')
+                db = MySqlContainer('mariadb:10.8.3')
```
violetbrina, line 92: Should we be hard coding the mariadb version here? Or is there somewhere else we can specify it for metamist that this function can use?

violetbrina (approved):
Looks good to me. Mostly wondering about the large commented out sections. Are they to be re-implemented later or simply removed? Otherwise looks good to me.

test/testbase.py
illusional: It's a good point, I'm really not sure. Maybe @vivbak might have an opinion, our mariadb instance is very decoupled from infrastructure code. Our current server version is `10.5.15-MariaDB-1:10.5.15+maria~buster`, so I guess we should really just test against the production code. 

Raised an issue here: https://github.com/populationgenomics/sample-metadata/issues/293

update_sample_status/main.py
```diff
@@ -43,26 +51,60 @@ def update_airtable(project_config: dict, sample: str, status: str):
...
+    samapi = SampleApi()
     sequence_type = SequenceType('genome')
     sample_type = SampleType('blood')
     seq_meta = {'batch': batch}
+    external_sequence_id: Optional[str] = None
 
-    try:
-        sequence_model = SequenceUpdateModel(
-            status=SequenceStatus(status), meta=seq_meta
+    sid_map = samapi.get_sample_id_map_by_internal([external_sample_id])
```
vivbak, line 63: This won't work with an external_sample_id. Did you mean 
```suggestion
    sid_map = samapi.get_sample_id_map_by_external(
        project=project, request_body=[external_sample_id]
    )
```
But I think this will also error out at this stage instead of progressing gracefully if the sample doesn't exist. Perhaps a try-except or something?

update_sample_status/main.py
```diff
@@ -43,26 +51,60 @@ def update_airtable(project_config: dict, sample: str, status: str):
...
+                type=sample_type,
+            ),
         )
-        seqapi.upsert_sequence_from_external_sample_and_type(
-            external_sample_id=sample,
-            sequence_type=sequence_type,
-            sequence_update_model=sequence_model,
+
+        # should create here if it doesn't exist
+        return abort(
```
vivbak, line 76: I'm not sure that aborting is what we want to do at this stage? Previously with this function, if the sample didn't exist it would create it and proceed with creating the sequence. Is there a reason we abort and exit here?

update_sample_status/main.py
```diff
@@ -74,6 +116,36 @@ def update_sm(project: str, sample: str, status: str, batch: str):
...
+    GET the latest relevant sequence ID or None if not found (meaning insert)
+    """
+    if external_sequence_id:
+        try:
+            sequence = seqapi.get_sequence_by_external_id(
+                project=project, external_id=external_sequence_id
+            )
+            return sequence['id']
+        except Exception as exception:  # pylint: disable=broad-except
+            # TODO: Check that it's a 404
```
vivbak, line 133: Was this TODO intentionally left?

vivbak (changes requested):
Sorry for the delayed response @illusional! It's looking really good! Just have some q. re the changes to that update sample status script

update_sample_status/main.py
```diff
@@ -43,26 +51,60 @@ def update_airtable(project_config: dict, sample: str, status: str):
...
-    external sample id"""
+def update_sm(project: str, external_sample_id: str, status: str, batch: str):
+    """Update the status of a sequence given it's corresponding external sample id"""
 
-    seqapi = SequenceApi()
+    samapi = SampleApi()
     sequence_type = SequenceType('genome')
     sample_type = SampleType('blood')
     seq_meta = {'batch': batch}
+    external_sequence_id: Optional[str] = None
```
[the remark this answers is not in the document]
illusional: (Reverted update_sample_status per our conversation)

vivbak (approved):

update_sample_status/main.py
illusional: (Reverted update_sample_status per our conversation)

api/routes/sequence.py
illusional: Removed :)

test/testbase.py
illusional: Uncommented :)

PR #291, opened 2022-10-11
New UI for Metamist

danielreti:
Draft for upcoming UI changes 

Completed Changes
- [x] Added Search to Nav Bar
- [x] Redesigned Nav Bar
- [x] Added links to project overview page
- [x] Created Sample page with info + pedigree
- [x] Included Project-level statistics
- [x] Fixed routing + implemented hotlink URLs
- [x] Add new interfaces to API

Upcoming
- [ ] Fix linting
- [ ] Add participant and family pages
- [ ] Improve pedigrees
- [ ] Add CRAM/file/seqr info (requires more data + endpoints)


web/src/Search.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const handleSearchChange = React.useCallback((e, data) => {
+        clearTimeout(timeoutRef.current as NodeJS.Timeout);
+        dispatch({ type: ActionKind.Start, query: data.value });
+
+        timeoutRef.current = setTimeout(async () => {
+            if (data.value.length === 0) {
+                dispatch({ type: ActionKind.Clear });
+                return;
+            }
+            new WebApi().searchByKeyword(data.value).then((resp) => {
```
illusional, line 79: Need to add a `catch` to this

web/src/Search.tsx
```diff
@@ -0,0 +1,121 @@
...
+    loading: boolean;
+    results: {}[];
+    value: string;
+};
+
+enum ActionKind {
+    Clear = "CLEAN_QUERY",
+    Start = "START_SEARCH",
+    Finish = "FINISH_SEARCH",
+    Update = "UPDATE_SELECTION",
```
illusional, line 16: Handle error case?

web/src/Search.tsx
```diff
@@ -0,0 +1,121 @@
...
+            input={{ fluid: true }}
+            loading={loading}
+            placeholder="Search..."
+            onResultSelect={(e, data) => {
+                dispatch({
+                    type: ActionKind.Update,
+                    selection: data.result.title,
+                });
+                navigate(
+                    `/project/${data.result.data.project}/${data.result.type}/${data.result.title}`,
```
illusional, line 108: Could we add a `searchResultToRoute` function? Make a switch handle it to catch unimplemented cases and just give a bit more flexibility to route handling.

```suggestion
                    `/project/${data.result.data.project}/${data.result.type}/${data.result.data.id}`,
```

web/src/mystyle.css
```diff
@@ -0,0 +1,54 @@
+.react-grid-layout {
```
illusional, line 1: Could we give this file a better name?

web/src/project/ProjectSummary.tsx
```diff
@@ -1,183 +1,364 @@
...
+    );
 
-    if (error) {
-        table = <p><em>An error occurred when fetching samples: {error}</em></p>
-    }
-    else if (!summary) {
-        if (globalContext.project) { table = <p>Loading...</p> }
-        else { table = <p><em>Please select a project</em></p> }
-    }
+    const renderLink = (s: any, k: any, external_id: string) => (
```
illusional, line 105: In previous projects I've added something like this to standardise SampleLinks, and then only have to change one place to change all the sample URLs. Could we add a collection of link to some standard place to make URL-ing standard?

```
export const SampleLink = ({id, projectName}) => <Link to `/project/{projectName}/sample/${id}`
```

web/src/project/ProjectSummary.tsx
```diff
@@ -1,183 +1,364 @@
...
+        <Link to={`/project/${projectName}/sample/${external_id}`}>
```
illusional, line 106: This should be the internal sample ID (`CPGxxx`)

web/src/urls.tsx
```diff
@@ -1,55 +1,56 @@
...
+                    element={<SwaggerUI url="/openapi.json" />}
+                />
+
+                <Route path="project/" element={<ProjectSummary />} />
+                <Route
+                    path="project/:projectName"
+                    element={<ProjectSummary />}
+                />
+                <Route
+                    path="project/:projectName/:page/:pageSize"
```
illusional, line 30: IMO, pageSize should remain a query param

web/src/urls.tsx
```diff
@@ -1,55 +1,56 @@
...
+                    path="project/:projectName/sample/:sampleName"
+                    element={<DetailedInfoPage />}
+                />
+                <Route
+                    path="project/:projectName/family/:familyName"
+                    element={<DetailedInfoPage />}
+                />
+                <Route
+                    path="project/:projectName/participant/:participantName"
+                    element={<DetailedInfoPage />}
```
illusional, line 51: All of these should be the internal identifier for each, all of these will have multiple external IDs (in the future) which could potentially change.

web/src/project/ProjectSummary.tsx
illusional:
```typscript
<SampleLink id="CPGLCL28">GRKMYTH11J1</SampleLink>
<SampleLink id="CPGLCL28" />
```

web/src/Search.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const navigate = useNavigate();
+    const [{ loading, results, value }, dispatch] = React.useReducer(
+        exampleReducer,
+        initialState
+    );
+
+    const timeoutRef = React.useRef<NodeJS.Timeout | null>(null);
+
+    const handleSearchChange = React.useCallback((e, data) => {
+        clearTimeout(timeoutRef.current as NodeJS.Timeout);
```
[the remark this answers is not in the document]
danielreti: Added

web/src/Search.tsx
```diff
@@ -0,0 +1,121 @@
...
+    results?: {}[];
+};
+
+const initialState: State = {
+    loading: false,
+    results: [],
+    value: "",
+};
+
+const exampleReducer = (state: State, action: Action): State => {
```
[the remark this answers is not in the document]
danielreti: Done

web/src/Search.tsx
```diff
@@ -0,0 +1,121 @@
...
+        case ActionKind.Clear:
+            return initialState;
+        case ActionKind.Start:
+            return { ...state, loading: true, value: action.query! };
+        case ActionKind.Finish:
+            return { ...state, loading: false, results: action.results! };
+        case ActionKind.Update:
+            return { ...state, value: action.selection! };
+
+        default:
```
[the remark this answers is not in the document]
danielreti: Replaced default with an unreachable function to catch any missing cases - is this what you meant?
```
const assertUnreachable = (x: ActionKind): never => {
    throw new Error(`Unexpected value: ${x}`);
};

    switch (action.type) {
        case "CLEAN_QUERY":
            return initialState;
        case "START_SEARCH":
            return { ...state, loading: true, value: action.query! };
        case "FINISH_SEARCH":
            return { ...state, loading: false, results: action.results! };
        case "UPDATE_SELECTION":
            return { ...state, value: action.selection! };
        case "QUERY_TOO_SHORT":
            return { ...state, loading: false };
    }
    return assertUnreachable(action.type);
```

web/src/Search.tsx
danielreti: Added a console.log  - how do you want these errors handled?
Also what are some examples of these errors?

web/src/Search.tsx
danielreti: TODO once I understand better

web/src/project/ProjectSummary.tsx
danielreti: Fixed

web/src/urls.tsx
danielreti: Noted and fixed for sample page - other pages to do. But they don't have internal and external IDs right? (do they?)

web/src/Search.tsx
danielreti: Added a catch and an error dropdown
Trigger by adding a raise ValueError in the python search

web/src/Search.tsx
danielreti: As above, added an error case

web/package.json
```diff
@@ -14,13 +14,16 @@
...
     "react-dom": "^17.0.2",
+    "react-grid-layout": "^1.3.4",
     "react-markdown": "^7.1.0",
     "react-router-dom": "^6.0.1",
     "react-scripts": "^4.0.3",
     "react-syntax-highlighter": "^15.4.4",
     "reactstrap": "^9.0.1",
     "remark-gfm": "^3.0.1",
     "remark-toc": "^8.0.1",
+    "semantic-ui-css": "^2.5.0",
```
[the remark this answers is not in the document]
danielreti: Switched all to semantic and removed `reactstrap`

web/src/Search.tsx
danielreti: Added a switch with alerts to handle unimplemented cases

web/src/project/ProjectSummary.tsx
danielreti: Great suggestion - added

web/src/urls.tsx
danielreti: Fixed

web/src/DetailedInfoPage.tsx
```diff
@@ -0,0 +1,499 @@
...
+    const [sampleInfo, setSampleInfo] = React.useState<Sample>();
+    const [pedigree, setPedigree] = React.useState<PedigreeEntry[]>();
+    const [participants, setParticipants] =
+        React.useState<ParticipantModel[]>();
+    const [selectedExternalID, setSelectedExternalID] =
+        React.useState<string>();
+    const [idNameMap, setIdNameMap] = React.useState<[string, string][]>();
+    const [sequenceInfo, setSequenceInfo] = React.useState<SampleSequencing>();
+
+    const getSamplesFromProject = React.useCallback(async () => {
```
illusional, line 52: I think this is fine for first iteration, but we should move all this logic to the web layer, and have a "get family display info for web" sort of thing.

illusional (approved):

web/src/DetailedInfoPage.tsx
danielreti: Yeah definitely, would be great to have fewer (or 1) API call which gets the all relevant info

PR #294, opened 2022-10-15
OpenAPI Optional types

illusional:
Hack the openapi generator for Pydantic to allow support for nullable fields in the openapi spec. Ignore one case (nullable schema) to avoid hitting an openapigenerator 5.x.x bug (we're pinned to 5.x.x).

This change is currently no-op, as we don't specify `response_model` for any types.


violetbrina (approved):

illusional:
I've validated this better, and it does work only with openapi-generator `5.3.0` - it does NOT work in 6.2.0 as `python-prior`, I keep seeing `external_id: str | None` type turn into `external_id: (ExternalId,)` with no definition for ExternalId. I can't track this down in their generator code, and not worth trying to upgrade at this stage.

It does have one limitation that function parameters don't work well to be optional, just set it to a sensible non nullable default. I'm following up in another PR to add more types to endpoints.

PR #295, opened 2022-10-16
Update test subset file

vivbak:
Fixes including: 
- Handles projects that don't have pedigree information, but require participants to be transferred 
- Fixes pedigree export type 
- Handles case where sample might exist already in the test project, but analysis would still need to be transferred 
- Updates API call 
- Creates a new function to get random families that is a little less random. It aims to evenly distribute the families that are chosen. 
For example, if there are 2 families of size 3, 2 families of size 4 and 2 families of size 5, the and the function has been asked to sample 3 families, it will pull 1 family of each size.

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -310,7 +310,7 @@ def transfer_ped(initial_project, target_project, family_ids):
 
 
 def transfer_participants(initial_project, target_project, participant_ids):
-    """Transfers a set list of participants between projects"""
+    """Transfers relevant participants between projects"""
 
     current_participants = papi.get_participants(
         initial_project,
@@ -325,8 +325,10 @@ def transfer_participants(initial_project, target_project, participant_ids):
     participants_to_transfer = []
     for participant in current_participants:
         if participant['external_id'] not in target_project_epids:
+            # Participants with id field will be updated & those without will be inserted
             del participant['id']
         transfer_participant = {k: v for k, v in participant.items() if v is not None}
+        # Participants are being created before the samples are, so this will be empty for now.
         transfer_participant['samples'] = []
         participants_to_transfer.append(transfer_participant)
 
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -106,7 +106,7 @@ def main(
         if resp.lower() != 'y':
             raise SystemExit()
 
-    random.seed(40)  # for reproducibility
+    random.seed(42)  # for reproducibility
 
     pid_sid = papi.get_external_participant_id_to_internal_sample_id(project)
     sample_id_by_participant_id = dict(pid_sid)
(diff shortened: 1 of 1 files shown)


scripts/create_test_subset.py
```diff
@@ -394,6 +452,54 @@ def _print_fam_stats(families: List):
...
+            returned_families.extend(random.sample(fams, n_pull))
+
+    else:
+        # we can't evenly distribute, so we'll just pull randomly
+        returned_families = random.sample(families_within_threshold, families_n)
+
+    return returned_families
+
+
 def _copy_files_in_dict(d, dataset: str, sid_replacement: Optional[Tuple] = None):
```
vladsavelyev, line 503: Missing return type hint

scripts/create_test_subset.py
```diff
@@ -280,6 +309,35 @@ def transfer_ped(initial_project, target_project, family_ids):
     return external_participant_ids
 
 
+def transfer_participants(initial_project, target_project, participant_ids):
```
vladsavelyev, line 312: Please add type hints. (I know I'm guilty of sometimes leaving them out too 😬)

scripts/create_test_subset.py
```diff
@@ -105,9 +114,18 @@ def main(
     if families_n is not None:
         fams = fapi.get_families(project)
         all_families = [family['id'] for family in fams]
-        families = random.sample(all_families, families_n)
-        pedigree = fapi.get_pedigree(project=project, internal_family_ids=families)
+        full_pedigree = fapi.get_pedigree(
+            project=project, internal_family_ids=all_families
+        )
+        external_family_ids = _get_random_families(full_pedigree, families_n)
```
vladsavelyev, line 120: You mentioned you ended up selecting families manually. Perhaps we could use an option to provide family IDs as input to the script?

scripts/create_test_subset.py
```diff
@@ -394,6 +452,54 @@ def _print_fam_stats(families: List):
...
 
 
+def _get_random_families(
+    families: List,
+    families_n: int,
+    include_single_person_families: Optional[bool] = False,
+):
+    """Obtains a subset of families, that are a little less random.
+    By default single-person families are discarded.
+    The function aims to evenly distribute the families chosen by size.
```
vladsavelyev, line 462: Nice!

scripts/create_test_subset.py
vladsavelyev: Or with `_get_random_families` the selection ends up better?

vladsavelyev (approved):
I like the families selection function!

scripts/create_test_subset.py
vivbak: Yeah so _get_random_families just does what I was doing with that manual selection :)

PR #298, opened 2022-10-18
Fix Get Sequences End Point 

vivbak:
Because all of the existing data in metamist do not have multiple external ids, the current get_sequences_by function fails. This PR handles sequences with/without multiple external ids.


vladsavelyev (approved):
Approving, even though I don't really understand what's `sample_sequencing_eid` for

Thanks for the fix!! Let's test it.

PR #299, opened 2022-10-18
Bump version: 5.0.0 → 5.1.0

vivbak:


vladsavelyev (approved):
Probably doesn't matter and I'm not really into semver, but for bug fixes we usually increment the "patch" component (so it would make it 5.0.1)

PR #305, opened 2022-10-25
Add specific fams or samples

vivbak:


scripts/create_test_subset.py
```diff
@@ -74,18 +74,41 @@
     default=False,
     help='Skip transferring pedigree/family information',
 )
+@click.option(
+    '--add-fam',
```
illusional, line 78:
```suggestion
    '--add-family',
```

scripts/create_test_subset.py
```diff
@@ -111,13 +134,25 @@ def main(
...
+                    project,
+                    _additional_samples,
+                )
+            )
         fams = fapi.get_families(project)
-        all_families = [family['id'] for family in fams]
+        families_to_subset = [
+            family['id']
+            for family in fams
+            if family['external_id'] not in _additional_fams
```
illusional, line 149: Not a big deal for this script with likely small amount of additional family IDs, but if doing existence checks, good to make sure `_additional_families` is a set for O(1) lookups instead of O(n).

scripts/create_test_subset.py
```diff
@@ -137,7 +172,19 @@ def main(
 
     else:
         assert samples_n
```
illusional, line 174: Not related to this change, but better to check and raise with a specific message. Also `asserts` can be skipped when running with Python in some modes (w/ `-O`).

scripts/create_test_subset.py
```diff
@@ -137,7 +172,19 @@ def main(
...
-        samples = random.sample(all_samples, samples_n)
+        if _additional_fams:
+            _additional_samples.extend(
+                get_samples_for_families(project, _additional_fams)
+            )
+
+        samples_to_subset = [
+            sample for sample in all_samples if sample['id'] not in _additional_samples
```
illusional, line 181: Same O(n) lookup comment with `_additional_samples`, but more so as _additional_samples could be significantly larger.

scripts/create_test_subset.py
```diff
@@ -390,6 +437,63 @@ def get_map_ipid_esid(
...
+    full_pedigree = fapi.get_pedigree(
+        project=project,
+        replace_with_participant_external_ids=False,
+        replace_with_family_external_ids=True,
+    )
+
+    ipids = [
+        family['individual_id']
+        for family in full_pedigree
+        if family['family_id'] in additional_fams
```
illusional, line 453: This will fail if `additional_fams` is `None`

scripts/create_test_subset.py
```diff
@@ -390,6 +437,63 @@ def get_map_ipid_esid(
...
+
+    pids = [sample['participant_id'] for sample in sample_objects]
+    full_pedigree = fapi.get_pedigree(
+        project=project,
+        replace_with_participant_external_ids=False,
+        replace_with_family_external_ids=True,
+    )
+
+    fams = [
+        fam['family_id'] for fam in full_pedigree if str(fam['individual_id']) in pids
```
illusional, line 491: I think this might result in duplicate family IDs as a family Id would appear multiple times in a pedigree. Good to `set` and cast to list if needed as a list (otherwise a Set in this usage makes sense IMO).

scripts/create_test_subset.py
```diff
@@ -74,18 +74,41 @@
...
+    help="""Additional samples to include.
+    This is in addition to the number of families specified in
+    --families and the number of samples specified in -n""",
+)
 def main(
     project: str,
     samples_n: Optional[int],
     families_n: Optional[int],
     skip_ped: Optional[bool] = True,
+    additional_fams: Optional[Tuple[str]] = None,
```
illusional, line 101: Renaming to `additional_families` is worth the extra chars :)

illusional (changes requested):
Sorry for the slow review, structure looks good to me. One request for rename + using sets when checking for existence in loops improves performances.

vivbak:
I'm so sorry that I forgot about this one @illusional! Hopefully have addressed everything :)

scripts/create_test_subset.py
```diff
@@ -146,8 +181,21 @@ def main(
...
-        samples = random.sample(all_samples, samples_n)
+        if _additional_families:
+            _additional_samples.extend(
+                get_samples_for_families(project, _additional_families)
+            )
+
+        samples_to_subset = [
+            sample
+            for sample in all_samples
+            if sample['id'] not in set(_additional_samples)
```
illusional, line 192: Don't cast to a set inside the loop, can you bring this (and the other examples) out of the loop please.

scripts/create_test_subset.py
```diff
@@ -84,18 +84,41 @@
...
+    additional_samples: Optional[Tuple[str]] = None,
 ):
     """
     Script creates a test subset for a given project.
     A new project with a prefix -test is created, and for any files in sample/meta,
     sequence/meta, or analysis/output a copy in the -test namespace is created.
     """
     samples_n, families_n = _validate_opts(samples_n, families_n)
+    _additional_families: List[str] = list(additional_families)
+    _additional_samples: List[str] = list(additional_samples)
```
illusional, line 121:
```suggestion
    _additional_families: list[str] = list(additional_families)
    _additional_samples: list[str] = list(additional_samples)
```

scripts/create_test_subset.py
```diff
@@ -11,7 +11,7 @@
 This example will populate acute-care-test with the metamist data for 4 families.
 """
 
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, Set
```
illusional, line 14: You can use inbuilt types now, eg:

```
dict[str, str]
set[int]
list[float]
str | None
```

Not super big deal, but good to refactor these out as we go

illusional (approved):
Please action minor changes for performance before merging

PR #316, opened 2022-10-27
Metamist Fixes V

illusional:
- Adds overflow scroll to project grid
- Removes response_model from endpoints to stop generated Python API from serializing models (addressed in #294)


vladsavelyev (approved):

danielreti (commented):
LGTM

PR #318, opened 2022-10-30
Upgrade dependencies

illusional:
Would love to upgrade to Python 3.11, but it's not available in pyenv yet and I'm not super keen to install outside it to test.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,14 @@
 # requirements are for running the server
 cpg-utils
-requests==2.25.1
+requests==2.28.1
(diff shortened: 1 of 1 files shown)


violetbrina (approved):

lgruen:
> Would love to upgrade to Python 3.11, but it's not available in pyenv yet and I'm not super keen to install outside it to test.

@illusional If you only want it for testing, it should be in conda already:

```sh
conda create -c conda-forge -n py311 python=3.11
```

PR #319, opened 2022-10-30
Update project admin to nest seqr project types

illusional:
We use one seqr project per sequence type, change the UI to support this functionality dynamically based on SequenceType. Use `rowSpan` to keep things appearing in the same table.

Requires migration on release to rename existing projects: 

seqr_guid -> seqr-project-genome

```sql
UPDATE project 
SET meta = JSON_REMOVE(JSON_SET(meta, "$.seqr-project-genome", JSON_VALUE(meta, "$.seqr_guid")), "$.seqr_guid") 
WHERE JSON_EXISTS(meta, "$.seqr_guid")
FROM project;
```

Preview: 

image

### DIFF
--- a/web/src/admin/ProjectsAdmin.tsx
+++ b/web/src/admin/ProjectsAdmin.tsx
@@ -1,7 +1,12 @@
 import * as React from "react";
-import { ProjectApi, ProjectRow } from "../sm-api";
+import { ProjectApi, ProjectRow, SequenceType } from "../sm-api";
 
-import { Message, Button, Checkbox, Input } from "semantic-ui-react";
+import { Message, Button, Checkbox, Input, InputProps } from "semantic-ui-react";
+
+interface ControlledInputProps extends InputProps {
+    project: ProjectRow;
+    metaKey: string;
+}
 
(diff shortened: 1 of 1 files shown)


web/src/admin/ProjectsAdmin.tsx
```diff
@@ -59,7 +62,7 @@ const ProjectsAdmin = (props: any) => {
                 // label={metaKey}
                 defaultValue={projectMeta[metaKey]}
                 // onChange={(e) => setProjectMetaState({ [project.name!]: { meta: { ...projStateMeta, [metaKey]: e.target.value } } })}
-                onBlur={(e) => {
+                onBlur={(e: any) => {
```
danielreti, line 65: Try to avoid `any` types. I think it should be `e: React.FocusEvent<HTMLInputElement>`

danielreti (approved):
1 small comment to fix then can merge :)

web/src/admin/ProjectsAdmin.tsx
illusional: Amazing, thank you!

PR #320, opened 2022-10-31
Handle Multiple Sequences 

vivbak:


vivbak:
Pending solution here https://github.com/populationgenomics/sample-metadata/issues/321

update_sample_status/main.py
```diff
@@ -106,6 +142,6 @@ def update_sample_status(request):  # pylint: disable=R1710
 
     update_airtable(project_config, sample, status)
 
-    update_sm(project, sample, status, batch)
+    update_sm(project, sample, status, batch, eseqid)
```
illusional, line 145: Is `eseqid` a map of `{name_of_provider: external_id}`, eg: `{"kccg": "EX01"}`?
That's what metamist expects, and what's annotated in `update_sm`, but the variable names make me less confident.

illusional (approved):
Change looks good except for external_sequence_id format.

update_sample_status/main.py
vivbak: Yeah the function kind of assumes that eseqid will be provided as a dict in the request. But I think it's probably better to just take eseqid as a string in the request and create the dict within this script.

vivbak: Will amend, thanks :)

vivbak:
Current test run broke on 
`exceptionMessage: "'SequenceApi' object has no attribute 'get_sequence_by_external_id'"` 
So I believe this is pending the changes here https://github.com/populationgenomics/sample-metadata/pull/325

vivbak:
Ah look at that, I actually made an issue for this https://github.com/populationgenomics/sample-metadata/issues/284

PR #325, opened 2022-10-31
Annotate endpoints with response_model

illusional:
Gives better types suggestions, and is stacked on #294.
FYI @vladsavelyev.


illusional:
Closing this, because it's unviable with current openapi + openapi-generator combo

PR #326, opened 2022-10-31
Fix sequence eid project

### LINKED ISSUE #321: Create New Sequence Fails | Project Missing
Calling something like this 
`new_sequence = NewSequence(**sequence_info)
        seqapi.create_new_sequence(new_sequence=new_sequence)`
Results in  
`Exception has occurred: IntegrityError (note: full exception trace is shown but execution is paused at: create_sequence) (1048, "Column 'project' cannot be null")`

illusional:
\+ re-enable tests

Closes #321


vivbak (approved):
🙏 thanks very much this is working well now!

Issue #328, opened 2022-11-02
Batch Summary Statistics 


vivbak:
**Background:** 
We often need to answer questions like 'how many samples have we processed?'. At present the UI presents a useful summary of how many sequences/crams there are. However, this isn't very granular at the moment, and it can be challenging to understand whether the numbers listed are in line with what is _expected_ to be there. 

For example, we had a case where we ingested 13 batches and then processed all of the samples. At some point, several samples needed to be sent back to be re-sequenced, creating 4 additional batches to be processed. Because these samples were not new, when viewing the summary stats, it is easy to assume that they've already been ingested (because the values for sequences appears correct). If however, we were able to easily identify which batches or how many batches were ingested, then it would've been easier to spot that the re-sequenced samples need to be ingested (because 4 batches were missing). 

**Proposed User Story:** 
As a user I want to be able to view stats per batch in the project summary stage, so I can quickly verify how many samples have been ingested/processed. 

**Request:** 
It would be useful to have: 
- How many batches have been ingested/processed for this project. 
- A breakdown of the current project summary stats per project. 

**Not essential but cool:**
- Being able to filter and sum according to batches.

PR #332, opened 2022-11-07
Batch Level Statistics

danielreti:
~~Would love some help on this, working on #328 .
In particular, I cannot fiddle with batch level Seqr and CRAM stats as I don't have any good test data for it, so can't validate any SQL queries I changed. (Not sure if Seqr stats are even currently working?)
Batch level sequence counts seem ok though.~~

For #328


danielreti:
Mostly just want to sanity check the changes I made to the API and SQL queries

danielreti:
And someone please help me set up my linter correctly because this fails every time and I have no idea why (@illusional )

db/python/layers/web.py
```diff
@@ -372,13 +387,28 @@ async def get_project_summary(
...
+
+        for seq in sorted(seen_seq_types):
+            cram_seqr_stats[seq] = {
                 'Sequences': str(seq_number_by_seq_type.get(seq, 0)),
                 'Crams': str(cram_number_by_seq_type.get(seq, 0)),
                 'Seqr': str(seqr_stats_by_seq_type.get(seq, 0)),
             }
+            seq_batch = seq_number_by_seq_type_and_batch.get(seq, 0)
+            sequence_stats[seq] = {}
+            for batch in sorted(
```
illusional, line 408: The sorting doesn't really mean anything here as it's dictionary, if we need sorting we should make it a list.

db/python/tables/sequence.py
```diff
@@ -279,21 +279,44 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
 
+    async def get_sequence_type_numbers_by_batch_for_project(self, project: ProjectId):
+        """
+        This groups by batch.
+        """
+
+        _query = """
+        SELECT batch, type, count(*) AS n
+        FROM (
+            SELECT IFNULL(JSON_EXTRACT(sq.meta, '$.batch'), 'null') as batch, sq.type
```
illusional, line 305: Is there a reason to cast it to `"null`"`, as opposed to leaving it as None, and flow it through as the Python `None` and then JSON `null`? Or if you can't have `null` keys, casting it to something like `"no-batch"` at that point?

db/python/tables/sequence.py
```diff
@@ -279,21 +279,44 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+            WHERE s.project = :project
+        ) as p
+        GROUP BY batch, type
+            """
+        rows = await self.connection.fetch_all(_query, {'project': project})
+        batch_result: Dict[str, Dict[str, str]] = {}
+        for batch, seqType, count in rows:
+            batch = batch.strip('\"')
+            batch_result.setdefault(seqType, {}).update(dict({batch: str(count)}))
+        print(batch_result)
```
illusional, line 317:
```suggestion
```

db/python/tables/sequence.py
```diff
@@ -279,21 +279,44 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+            INNER JOIN sample s ON s.id = sq.sample_id
+            WHERE s.project = :project
+        ) as p
+        GROUP BY batch, type
+            """
+        rows = await self.connection.fetch_all(_query, {'project': project})
+        batch_result: Dict[str, Dict[str, str]] = {}
+        for batch, seqType, count in rows:
+            batch = batch.strip('\"')
+            batch_result.setdefault(seqType, {}).update(dict({batch: str(count)}))
```
illusional, line 316: Consider making batch_result a defaultdictionary:

```suggestion
        batch_result: Dict[str, Dict[str, str]] = defaultdict(dict)
        for batch, seqType, count in rows:
            batch = batch if batch else "no-batch"
            batch = batch.strip('\"')
            batch_result[seqType][batch] = str(count)
```

web/src/project/ProjectSummary.tsx
```diff
@@ -310,8 +319,39 @@ export const ProjectSummary = () => {
     const titleCase = (s: string) => {
         return s[0].toUpperCase() + s.slice(1).toLowerCase();
     };
+    console.log(summary?.sequence_stats);
```
illusional: Should the batch table only show if there are other keys than "no-batch" - seems sort of irrelevant for like perth-neuro or something.

db/python/layers/web.py
```diff
@@ -64,6 +64,8 @@ class ProjectSummary:
     # stats
     total_samples: int
     total_participants: int
+    total_sequences: int
+    cram_seqr_stats: dict[str, dict[str, str]]
     sequence_stats: dict[str, dict[str, str]]
```
illusional, line 69: This should be renamed to `batch_sequence_stats`?

test/test_web.py
```diff
@@ -159,13 +161,15 @@ async def test_project_summary(self):
...
-            sequence_stats={
+            total_sequences=1,
+            cram_seqr_stats={
                 'genome': {
                     'Sequences': '1',
                     'Crams': '0',
                     'Seqr': '0',
                 }
             },
+            sequence_stats={'genome': {'null': '1'}},
```
illusional, line 172: Would you consider modifying this test to check if the batch stats come through correctly? I should have done the same with the `seqr`, because it obviously doesn't work for seqr.

illusional (changes requested):
Awesome work on the back-end changes!! I've left a few notes:

- Keeping `null` through where we can, and only casting at the end
- Renaming `sequence_stats` to `batch_sequence_stats`
- Updating the test to test these changes

db/python/layers/web.py
danielreti: seen_batches and seen_seq_types are sets which are basically lists so sorting should work?

db/python/tables/sequence.py
danielreti: SQL handles a None batch field and an omitted batch field as 2 distinct groupings and I wanted to combine them at the SQL level as it's easier to do here than in Python.

db/python/tables/sequence.py
danielreti: Oops

db/python/tables/sequence.py
danielreti: Fixed

web/src/project/ProjectSummary.tsx
danielreti: Can do - would it make more sense to return an empty object from the backend if there are no other keys or should we check on the front end for other keys?

test/test_web.py
danielreti: How detailed do you want me to go?

danielreti (commented):
Responded

web/src/project/ProjectSummary.tsx
illusional: Yeah I think empty object if there are no other keys, make it easy for you. Total sequences are available in other places.

db/python/tables/sequence.py
illusional: Makes sense, can you use the SQL NULL directly instead of string?

db/python/layers/web.py
illusional: The actual sorting makes sense, but you put the sorted result on a dictionary. In the end, I think Python orders the JSON string in the same order as inserted, and Typescript's `Object.keys` probably preserves the "order seen", but this ordering a map, dict, object is undefined.

test/test_web.py
illusional: Not super detailed, just set batch on one should basically test all of it (except the future case of no known batches results empty here.

db/python/tables/sequence.py
illusional, line 316: Less consequential and feel free to ignore because it probably works better for display, but food for thought. Does it make more sense to order `batch` then `seqtype`. I think conceptually I'd want to know what's inside a batch, rather than genomes had transfers in different batches.

db/python/tables/sequence.py
danielreti: It's weird, the SQL NULL comes about when the field is omitted, but the Python None type is a 'null' string once in SQL, unless I made a mistake in the data_generation.py step (very possible). So I converted the SQL NULL to 'null' so it merges with the Python 'null'. I couldn't find a way to make them both turn into NULL. Thoughts?

test/test_web.py
danielreti: Fixed

db/python/layers/web.py
danielreti: Removed sorting for now, can sort on client later if required

db/python/tables/sequence.py
danielreti: Resolved - merge SQL null and Python None type at SQL query, then relabel as 'no-batch' at Python level

web/src/project/ProjectSummary.tsx
danielreti: Returns {} if 'no-batch' is the only key

db/python/tables/sequence.py
danielreti: This was a good spot. Removed the toggleable nested batch table and added a second batch table (only rendered if there is batch data), which shows the breakdown of each batch by seqType and a total count for each batch.

danielreti (commented):
Fixed discussed issues

api/routes/web.py
```diff
@@ -34,7 +34,9 @@ class ProjectSummaryResponse(BaseModel):
     # high level stats
     total_participants: int
     total_samples: int
-    sequence_stats: dict[str, dict[str, str]]
+    total_sequences: int
+    cram_seqr_stats: dict[str, dict[str, str]]
```
illusional, line 38: Could be worth a comment here just saying _by sequence_type then batch_

db/python/layers/web.py
```diff
@@ -372,20 +387,33 @@ async def get_project_summary(
...
         for seq in seen_seq_types:
-            sequence_stats[seq] = {
+            cram_seqr_stats[seq] = {
                 'Sequences': str(seq_number_by_seq_type.get(seq, 0)),
                 'Crams': str(cram_number_by_seq_type.get(seq, 0)),
                 'Seqr': str(seqr_stats_by_seq_type.get(seq, 0)),
             }
 
+        for batch in seen_batches:
+            sequence_stats[batch] = {}
```
illusional, line 403: This is fine, but also a good use for `defaultdict`

db/python/tables/sequence.py
```diff
@@ -279,21 +279,45 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+
+        _query = """
+            SELECT batch, type, count(*) AS n
+            FROM (
+                SELECT IFNULL(JSON_EXTRACT(sq.meta, '$.batch'), 'null') as batch, sq.type
+                FROM sample_sequencing sq
+                INNER JOIN sample s ON s.id = sq.sample_id
+                WHERE s.project = :project
+            ) as p
+            GROUP BY batch, type
```
illusional, line 310: You can actually use the selected fields inside the group by to avoid the double query, eg:

```suggestion
            SELECT IFNULL(JSON_EXTRACT(sq.meta, '$.batch'), 'null') as batch, sq.type, COUNT(*) AS n
            FROM sample_sequencing sq
            INNER JOIN sample s ON s.id = sq.sample_id
            WHERE s.project = :project
```

db/python/tables/sequence.py
```diff
@@ -279,21 +279,45 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+        """
+        rows = await self.connection.fetch_all(_query, {'project': project})
+        batch_result: Dict[str, Dict[str, str]] = defaultdict(dict)
```
illusional, line 313: We can use lowercase `dict` instead of from the typing library
```suggestion
        batch_result: dict[str, dict[str, str]] = defaultdict(dict)
```

db/python/tables/sequence.py
```diff
@@ -279,21 +279,45 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+            GROUP BY type
         """
 
         rows = await self.connection.fetch_all(_query, {'project': project})
 
         return {r['type']: r['n'] for r in rows}
 
+    async def get_sequence_type_numbers_by_batch_for_project(self, project: ProjectId):
+        """
+        This groups by batch. Python NONE types and SQL NULL types are combined into a null string by IFNULL
```
illusional, line 299:
```suggestion
        Grouped by the meta.batch field on a sequence
```

db/python/tables/sequence.py
```diff
@@ -279,21 +279,45 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+        """
+
+        _query = """
```
illusional, line 302:
```suggestion
        # During the query, cast to the string null with IFNULL, as the GROUP BY
        # treats a SQL NULL and JSON NULL (selected from the meta.batch) differently.
        _query = """
```

db/python/tables/sequence.py
```diff
@@ -279,21 +279,45 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+                FROM sample_sequencing sq
+                INNER JOIN sample s ON s.id = sq.sample_id
+                WHERE s.project = :project
+            ) as p
+            GROUP BY batch, type
+        """
+        rows = await self.connection.fetch_all(_query, {'project': project})
+        batch_result: Dict[str, Dict[str, str]] = defaultdict(dict)
+        for batch, seqType, count in rows:
+            batch = batch.strip('\"') if batch != 'null' else 'no-batch'
```
illusional, line 315: I have a feeling that randomly a batch might be a number (or non-string) in the JSON,
and it will be a tricky error to chase down.

```suggestion
            batch = str(batch).strip('\"') if batch != 'null' else 'no-batch'
```

db/python/tables/sequence.py
```diff
@@ -279,21 +279,45 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+                SELECT IFNULL(JSON_EXTRACT(sq.meta, '$.batch'), 'null') as batch, sq.type
+                FROM sample_sequencing sq
+                INNER JOIN sample s ON s.id = sq.sample_id
+                WHERE s.project = :project
+            ) as p
+            GROUP BY batch, type
+        """
+        rows = await self.connection.fetch_all(_query, {'project': project})
+        batch_result: Dict[str, Dict[str, str]] = defaultdict(dict)
+        for batch, seqType, count in rows:
```
illusional, line 314: Could consider using the field indexes, rather than the order from the query, but not a big deal.

db/python/tables/sequence.py
```diff
@@ -279,21 +279,45 @@ async def get_sequence_type_numbers_for_project(self, project: ProjectId):
...
+            batch = batch.strip('\"') if batch != 'null' else 'no-batch'
+            batch_result[batch][seqType] = str(count)
+        if list(batch_result.keys()) == ['no-batch']:
```
illusional, line 317: My gut is this is a little clunky, and could do something like:
```suggestion
        if len(batch_result) == 1 and 'no-batch' in batch_result:
            # if there are no batches, ignore the no-batch option
```

illusional (changes requested):
Looks really great, super happy with how this looks and awesome to have a test cover this new pathway.
I've requested a few small code-changes, mostly non-functional :)

api/routes/web.py
danielreti: Added comments - cram_seqr_stats is by seqType then cram/seqr/seq counts, batch_sequence_stats is by batch then seqType.

db/python/layers/web.py
danielreti: I figured we didn't need a defaultdict for this but actually I'll make the code more Pythonic with list comprehension. Then no need to create an empty dict first anyway -
```dict([(seq, seq_number_by_seq_type_and_batch[batch].get(seq, 0)) for seq in seen_seq_types])```

db/python/tables/sequence.py
danielreti: When do I use the typing library?

db/python/tables/sequence.py
danielreti: I don't think this matters as we can reorder on the client if/when we want to anyway

db/python/tables/sequence.py
danielreti: Good catch!

db/python/tables/sequence.py
danielreti: I had this initially but thought it was too verbose haha - fixed :)

danielreti (commented):
Applied suggestions :)

illusional (approved):
Awesome, looks great!

PR #340, opened 2022-11-09
Postgres conversion

illusional:
Context: https://centrepopgen.slack.com/archives/C03FZL2EF24/p1667796642812949

Test convert from mariadb to postgres

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -39,7 +39,7 @@ jobs:
 
           # liquibase connector
           pushd db/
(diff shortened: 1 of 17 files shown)


illusional:
Closing with no intention to continue, this was a great test - and it was awesome to see Hasura be able to pick up our schema and allow for very quick graphql queries - but it was missing too much business logic which we encode here.

It also relied on the temporal table plugin, which isn't supported by the hosted versions (issue: https://issuetracker.google.com/issues/68986162).

There isn't a lot of novelty in this PR, so no point keeping it as a draft.

PR #341, opened 2022-11-11
Remove Zip

vivbak:
Caused mismatch between samples and sequencing info.

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -156,7 +156,7 @@ def main(
     except exceptions.ApiException:
         seq_info_by_s_id = {}
     else:
-        seq_info_by_s_id = dict(zip(sample_ids, seq_infos))
+        seq_info_by_s_id = {seq['sample_id']: seq for seq in seq_infos}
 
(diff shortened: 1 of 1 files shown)


vivbak:
FYI @KatalinaBobowik, this should fix the issue. 

@illusional @vladsavelyev This would have impacted any -test project that was set up with this script. It's not too much of a big deal, but we should probably clear any -test projects that have been set up with this script (as well as corresponding buckets).

scripts/create_test_subset.py
```diff
@@ -156,7 +156,7 @@ def main(
     except exceptions.ApiException:
         seq_info_by_s_id = {}
     else:
-        seq_info_by_s_id = dict(zip(sample_ids, seq_infos))
+        seq_info_by_s_id = {seq['sample_id']: seq for seq in seq_infos}
```
illusional, line 159: Oh damn, this burned production-pipelines early on too.

illusional (approved):

scripts/create_test_subset.py
vivbak: Yeah, the second I saw it, I remembered the last incident. I tried to search the code base to see if we do this anywhere else and didn't notice anything.

PR #342, opened 2022-11-11
Small UI fixes

danielreti:
Fixed missing title on sample page
Don't try to render null fields on sample page
Don't render next page button when only 1 page
Added missing Sample interface
Don't throw error when a sample has no associated sequences
Remove excessive nesting; instead return early (general code cleanup)

### DIFF
--- a/db/python/layers/sequence.py
+++ b/db/python/layers/sequence.py
@@ -87,7 +87,7 @@ class SampleSequenceLayer(BaseLayer):
         projects, sequences = await self.seqt.get_sequences_by(sample_ids=sample_ids)
 
         if (
-            check_project_ids and projects and sequences
+            check_project_ids and sequences
         ):  # Only do this check if a sample actually has associated sequences
             await self.ptable.check_access_to_project_ids(
                 self.author, projects, readonly=True
--- a/db/python/layers/sequence.py
+++ b/db/python/layers/sequence.py
@@ -86,7 +86,9 @@ class SampleSequenceLayer(BaseLayer):
         """
         projects, sequences = await self.seqt.get_sequences_by(sample_ids=sample_ids)
 
-        if check_project_ids:
+        if (
+            check_project_ids and projects and sequences
+        ):  # Only do this check if a sample actually has associated sequences
             await self.ptable.check_access_to_project_ids(
                 self.author, projects, readonly=True
             )
--- a/web/src/DetailedInfoPage.tsx
+++ b/web/src/DetailedInfoPage.tsx
@@ -60,7 +60,7 @@ enum SampleType {
 }
 
 interface Sample {
-    id: number | string;
+    id: string;
     external_id: string;
     participant_id?: number;
     active?: boolean;
--- a/web/src/DetailedInfoPage.tsx
+++ b/web/src/DetailedInfoPage.tsx
@@ -323,7 +323,6 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
     const safeValue = (value: any): string => {
         if (!value) return value;
         if (Array.isArray(value)) {
-            debugger;
             return value.map(safeValue).join(", ");
         }
         if (typeof value === "number") {
@@ -333,12 +332,10 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
             return value;
         }
         if (value && typeof value === "object" && !Array.isArray(value)) {
-            debugger;
             if (!!value.location && !!value.size) {
                 return `${value.location} (${formatBytes(value.size)})`;
             }
         }
-        debugger;
         return JSON.stringify(value);
     };
 
@@ -427,17 +424,14 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
                     .filter(([key, value]) =>
                         sampleFieldsToDisplay.includes(key)
                     )
-                    .map(([key, value]) =>
-                        value ? (
-                            <div key={`${key}-${value}`}>
-                                <b>{key}:</b> {value.toString()}
-                            </div>
-                        ) : (
-                            <React.Fragment
-                                key={`${key}-${value}`}
-                            ></React.Fragment>
-                        )
-                    )}
+                    .map(([key, value]) => (
+                        <div key={`${key}-${value}`}>
+                            <b>{key}:</b>{" "}
+                            {value != null
+                                ? value.toString()
+                                : "null/undefined"}
+                        </div>
+                    ))}
             </>
         );
     };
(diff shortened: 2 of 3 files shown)


db/python/layers/sequence.py
```diff
@@ -86,7 +86,9 @@ async def get_sequences_for_sample_ids(
         """
         projects, sequences = await self.seqt.get_sequences_by(sample_ids=sample_ids)
 
-        if check_project_ids:
+        if (
+            check_project_ids and projects and sequences
```
illusional, line 90: This should fail if there are sequences but no projects
```suggestion
            check_project_ids and sequences
```

web/src/DetailedInfoPage.tsx
```diff
@@ -37,21 +37,37 @@ interface SequenceMeta {
...
+    status: string;
+}
+
+enum SampleType {
+    BLOOD = "blood",
+    SALIVA = "saliva",
+}
+
+interface Sample {
+    id: number | string;
```
illusional, line 63: On web side this will only ever be a string:

```suggestion
    id: string;
```

web/src/DetailedInfoPage.tsx
```diff
@@ -306,28 +321,26 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
     }, [projectName, samples, sampleName]);
 
     const safeValue = (value: any): string => {
-        if (!value) return value
+        if (!value) return value;
         if (Array.isArray(value)) {
-            debugger
-            return value.map(safeValue).join(", ")
+            debugger;
```
illusional, line 326:
```suggestion
```

web/src/DetailedInfoPage.tsx
```diff
@@ -306,28 +321,26 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
...
-                return `${value.location} (${formatBytes(value.size)})`
+                return `${value.location} (${formatBytes(value.size)})`;
             }
-
         }
-        debugger
-        return JSON.stringify(value)
-
-    }
+        debugger;
```
illusional, line 341:
```suggestion
```

web/src/DetailedInfoPage.tsx
```diff
@@ -372,6 +391,133 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
...
+                    }}
+                >
+                    Sample Information
+                </h4>
+                {Object.entries(sampleInfo)
+                    .filter(([key, value]) =>
+                        sampleFieldsToDisplay.includes(key)
+                    )
+                    .map(([key, value]) =>
+                        value ? (
```
illusional, line 431: I think we should display keys with falsey, especially this will probably catch falsey like `false, 0, etc`.

web/src/DetailedInfoPage.tsx
```diff
@@ -372,6 +391,133 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
...
+                            <React.Fragment
+                                key={`${key}-${value}`}
+                            ></React.Fragment>
+                        )
+                    )}
+            </>
+        );
+    };
+
+    const renderPedigreeSection = () => {
```
illusional, line 445: Can do this later, but would be good to make this it's own separate component importable in other places.

web/src/DetailedInfoPage.tsx
```diff
@@ -372,6 +391,133 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
...
+                                    item.sex
+                                )}
+                        </React.Fragment>
+                    ))}
+            </>
+        );
+    };
+
+    const renderTitle = () => {
+        return (
```
illusional, line 481: Why not just set this to a component?
```suggestion
    const title = (
```

web/src/DetailedInfoPage.tsx
```diff
@@ -37,21 +37,37 @@ interface SequenceMeta {
...
+    external_id: string;
+    participant_id?: number;
+    active?: boolean;
+    meta?: { [key: string]: any };
+    type?: SampleType;
+    project?: number;
+    author?: string;
 }
 
 const sampleFieldsToDisplay = ["active", "type", "participant_id"];
```
illusional, line 73: I think it would be worth turning `DetailedInfoPage` into a state component, then we can use a [React error boundary](https://reactjs.org/docs/error-boundaries.html) to show better error messages and not have the page crash.

Even just display the JSON response if we have one to the user, as that might be useful if there are display issues.

illusional (changes requested):

db/python/layers/sequence.py
danielreti: Fixed

web/src/DetailedInfoPage.tsx
danielreti: Not sure what you mean by a state component. 
Some of the changes in this PR should fix a lot of crashes.

web/src/DetailedInfoPage.tsx
danielreti: Good point - this was meant to catch instances where a field is null (or undefined), but false or 0 are both valid. 
Replaced with `{value != null ? value.toString() : "null/undefined"}`

web/src/DetailedInfoPage.tsx
danielreti: Might do in future - haven't done yet as might be swapping to a separate package at some point

web/src/DetailedInfoPage.tsx
danielreti: Should be, fixed

danielreti (commented):
Fixed most suggestions - 1 to discuss more

web/src/DetailedInfoPage.tsx
illusional: Sorry I meant a class component, then you can implement these two methods to bound any further exceptions:

```typescript
class DetailedInfoPage extends React.Component {

  static getDerivedStateFromError(error) {
    // Update state so the next render will show the fallback UI.
    return { hasError: true };
  }

  componentDidCatch(error, errorInfo) {
    // You can also log the error to an error reporting service
    logErrorToMyService(error, errorInfo);
  }

  render() {
    if (this.state.hasError) {
      // You can render any custom fallback UI
      return <h1>Something went wrong.</h1>;
    }

    return this.props.children; 
  }

File: web/src/DetailedInfoPage.tsx [hunk 9]
```diff
@@ -427,17 +424,14 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
...
-                                key={`${key}-${value}`}
-                            ></React.Fragment>
-                        )
-                    )}
+                    .map(([key, value]) => (
+                        <div key={`${key}-${value}`}>
+                            <b>{key}:</b>{" "}
+                            {value != null
+                                ? value.toString()
+                                : "null/undefined"}
```
illusional, line 432: Could use typescript optional coalescing to simplify this, and I think other people might find `null` confusing.

```suggestion
                            {value?.toString() ?? <em>no value</em>}
```

web/src/DetailedInfoPage.tsx
```diff
@@ -526,7 +516,7 @@ export const DetailedInfoPage: React.FunctionComponent<{}> = () => {
             {projectName && !isLoading && !error && (
                 <>
                     <div className="detailedInfo">
-                        <div>{renderTitle()}</div>
+                        <div>{renderTitle}</div>
                         <br />
                         <div>{renderPedigreeSection()}</div>
```
illusional, line 521: Similar treatment for all these functions because they're basically just constants.

illusional (commented):
Happy to chat on Monday about the _class_ component and error boundaries :)

web/src/DetailedInfoPage.tsx
danielreti: Oooh yes smart, will fix this after we discuss the other thing

web/src/DetailedInfoPage.tsx
danielreti: The renderTitle was a simple one, the others have if statements and early returns which was the whole reason I made them functions. I thought it was preferable to making them variables and either having to do `let var` and then updating them or doing a bunch of nested checks. Is it preferable for them to not be functions?

web/src/DetailedInfoPage.tsx
danielreti: I was under the impression class components should never be used these days. But admittedly I don't know much about them. 
There are some npm packages which seem to mimic the behaviour (https://www.npmjs.com/package/react-use-error-boundary) if they are of interest. 
Happy to discuss on Monday :) (Think it's wise to ask @daniaki for some expertise as he knows more React than I do)

illusional (approved):
Looks great! 🔒

PR #353, opened 2022-11-21
More UI fixes

### LINKED ISSUE #352: Summary Page Error
Page doesn't reset when a project is selected and then the user clicks 'Explore'
Also, when project is unselected, no stats should be visible

### LINKED ISSUE #366: Displaying Ext. Seq. Id's  in UI
At present External ID's, listed under Sequence Information on the sample page appear as` [object Object]` rather than the key value pairs.

danielreti:
Fixes #352 
Fixes Seqr Links
Fixes #366

### DIFF
--- a/web/src/project/ProjectSummary.tsx
+++ b/web/src/project/ProjectSummary.tsx
@@ -75,7 +75,10 @@ export const ProjectSummary = () => {
 
     const getProjectSummary = React.useCallback(
         async (token: any) => {
-            if (!projectName) return;
+            if (!projectName) {
+                setSummary(undefined);
+                return;
+            }
             let sanitisedToken = !!token ? token : undefined;
             setError(undefined);
             setIsLoading(true);
--- a/web/src/project/ProjectSummary.tsx
+++ b/web/src/project/ProjectSummary.tsx
@@ -458,7 +458,7 @@ export const ProjectSummary = () => {
     };
 
     const seqrLinks = () => {
-        if (!summary?.seqr_links.length) {
+        if (!summary?.seqr_links || _.isEmpty(summary?.seqr_links)) {
             return <></>;
         }
         return (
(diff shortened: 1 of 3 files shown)


web/src/project/ProjectSummary.tsx
```diff
@@ -455,7 +458,7 @@ export const ProjectSummary = () => {
     };
 
     const seqrLinks = () => {
-        if (!summary?.seqr_links.length) {
+        if (!summary?.seqr_links || _.isEmpty(summary?.seqr_links)) {
```
illusional, line 461: Can you pass the seqr_links directly to isEmpty, removing the first check, eg:

```suggestion
        if (_.isEmpty(summary?.seqr_links)) {
```

Or use optional coalesce:

```suggestion
        if ((summary?.seqr_links?.length ?? 0) == 0) {
```

illusional (approved):

web/src/project/ProjectSummary.tsx
danielreti: I didn't think you could check object size with .length, that's why it broke in the first place. So now I check if the object doesn't exist or if it's empty, then don't render anything.

illusional: But you can do that with the optional chain operator, eg: `summary?.length` will be falsey if empty or unset.

danielreti: Sorry my point is that objects don't have a length so {}.length is not a thing. I mistakenly added in .length in a previous PR and just wanted to fix it. So for objects, I want to check is it not null/undefined and is it non empty. I thought that best done with is !object || _.isEmpty(object)? Could use Object.keys(object).length too

danielreti: How about `if (!Object.keys(summary.seqr_links ?? {}).length) return <></>` . This catches undefined, null, nonexistent key and empty objects

PR #354, opened 2022-11-23
Sequence technology

illusional:
New types, it would be good to annotate the analysis entries with the technology, then allow for filtering based on that. (To migrate the `analysis.meta.sequencing_type` to these new values).

```python
class SequenceType(Enum):
    """Type of sequencing"""

    GENOME = 'genome'
    EXOME = 'exome'
    TRANSCRIPTOME = 'transcriptome'
    CHIP = 'chip'
    MT = 'mtseq'


class SequenceTechnology(Enum):
    """Sequencing technologies"""

    SHORT_READ = 'short-read'
    LONG_READ = 'long-read'
    SINGLE_CELL_RNA_SEQ = 'single-cell-rna-seq'
    BULK_RNA_SEQ = 'bulk-rna-seq'
```


db/python/tables/sequence.py
```diff
@@ -219,6 +223,7 @@ async def get_sequence_ids_for_sample_id(
         Get the sequence IDs from internal sample_id and sequence type
         Map them to be keyed on sequence type
         """
+        # TODO: how should we address this, because we ultimately want to split on the technology I think
```
vivbak, line 226: Do you think a nested dict would be too clunky?

vivbak (commented):
LGTM! But note I didn't pull and test manually.

db/python/tables/sequence.py
illusional: I think sequence groups would solve this completely, so I'll keep the behaviour here the same (no breaking changes), and then more seriously action sequence groups.

vivbak (approved):

PR #361, opened 2022-11-30
Handle External Sequence IDs

vivbak:


sample_metadata/parser/generic_parser.py
```diff
@@ -212,6 +212,40 @@ def wrapper(*args, **kwargs):
...
+    external_sequence_ids: list[str] = []
+    for participant in participant_map:
+        for sample in participant_map[participant]:
+            for sequence in participant_map[participant][sample]:
+                external_sequence_ids.append((sequence.get('Sequence ID')))
+
+    return external_sequence_ids
+
+
+def get_existing_sequences(
```
illusional, line 226: Ditto about not being defined on the class

illusional (changes requested):
Looking good, minor feedback

illusional (approved):

PR #365, opened 2022-12-06
Fixes to handle ext_seq_ids 

vivbak:
2 work-arounds here 
1. Because the fastq's have been moved to archival and the read information from metamist has been cleared, populating external_sequence id's is tricky. This PR includes a quick script that lets us backfill from a csv as a workaround. Note: This is handling an edge case that occurred as a result of a schema change happening after we already cleared the sequences for one of our projects. The script isn't very elegant, but we shouldn't need to use it more than once. 
2. Our existing-cohort csv's don't actually contain a sequence_id field at present (future ones should), but we need to be able to process the remaining batches from prophecy ASAP.


vivbak:
FYI @KatalinaBobowik

scripts/back_populate_sequences_from_file.py
```diff
@@ -0,0 +1,61 @@
...
+
+    with open(dest_file) as fh:
+        rd = csv.DictReader(fh, delimiter=',')
+        for row in rd:
+            external_id = row['sample_id']
+            external_sequence_id = row['sequence_id']
+            sample_map = sapi.get_sample_id_map_by_external(
+                project, request_body=[external_id]
+            )
+            sequence = seqapi.get_sequences_by_sample_ids([sample_map[external_id]])
```
illusional, line 47: I'm not going to block this, but do get into the habit of not making REST calls inside loops.

illusional (approved):

scripts/back_populate_sequences_from_file.py
vivbak: yeah I know its awful sorry

PR #371, opened 2022-12-12
Got basic GraphQL working in frontend

danielreti:


web/src/index.tsx
```diff
@@ -1,19 +1,36 @@
...
+import {
+    ApolloProvider,
+    ApolloClient,
+    createHttpLink,
+    InMemoryCache,
+} from "@apollo/client";
 
+// 2
+const httpLink = createHttpLink({
+    uri: "http://localhost:8000/graphql",
```
illusional, line 17: Can this just be `/graphql`:

```suggestion
    uri: "/graphql",
```

web/src/graphQL.tsx
```diff
@@ -0,0 +1,23 @@
...
+    }
+`;
+export const GraphQL: React.FunctionComponent<{}> = () => {
+    const { loading, error, data } = useQuery(GET_PROJECTS);
+    if (loading) return <>"Loading..."</>;
+    if (error) return <>`Error! ${error.message}`</>;
+    console.log(data.myProjects);
+    return (
+        <>
+            {data.myProjects.map((item: { id: string }) => (
```
illusional, line 18: Awesome! Do you get any type-safety here? Or would you basically have to create an interface for each endpoint?

web/src/index.tsx
danielreti: Fixed

web/src/graphQL.tsx
danielreti: Added :)

danielreti:
This became a little messy I think but it seems to be on the right track now. 

Before deploying locally, you need to run `strawberry export-schema api.graphql.schema:schema > web/src/schema.graphql` in the root directory to generate a GraphQL schema. Then running `npm run compile` before `npm start` creates the typed gql interfaces. It also creates some new files which I added to gitignore along with the schema (which should regenerate every deploy I think). We'll need to add this flow to the deploy actions too. 

Happy to discuss and make changes as obviously this is not my area and I'm not sure of the best way to do things. Also would love a double check on a few things that changed when I updated things and I'm not sure why, like the jsx compiler options in `tsconfig.json` changing from `'react'` to `'react-jsx'` and the fact that TypeScript now seems to prevent compilation when there are errors (which I like) whereas previously it would still compile and then show errors in the console (and break the page if they were breaking errors).

illusional:
Can you make `npm compile` also run `strawberry export` (and probably also `npm build` run the compile, which also transitively gets the schema). That way you don't have to change the github action when compiling.

web/src/graphQL.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { gql } from "../src/__generated__/gql";
+
+const GET_PROJECTS = gql(`
+    query getProjects {
+        myProjects {
+            id
+        }
+    }
+`);
+export const GraphQL: React.FunctionComponent<{}> = () => {
```
illusional, line 13: Is this a test component, or a ProjectSelector?

web/src/index.tsx
```diff
@@ -1,19 +1,36 @@
-import React from 'react';
-import ReactDOM from 'react-dom';
-import './index.css';
-import App from './App';
-import reportWebVitals from './reportWebVitals';
+import React from "react";
```
illusional, line 1: Can you standardise quotes + formatting for the projects in one PR, this reformatting clutters the PR changes when reviewing

illusional (changes requested):

danielreti:
Problem is you can only run the `strawberry export` in the root directory not `web` as it doesn't support relative imports for some reason?? (You get this error: `TypeError: the 'package' argument is required to perform a relative import`) Otherwise I would have just added it to the `npm build` already

web/src/graphQL.tsx
danielreti: Just a test component. It's still a draft PR, just wanted to get something working. Once all the schema setup (+linting etc) have been setup, I'll delete this component and start making bigger changes.

illusional:
You should be able to change directory within a compile, eg ([source](https://stackoverflow.com/a/54390944/2860731)):

```
(cd /path/to/your/app && npm start && cd -)
```

PR #373, opened 2022-12-12
Add web / analysis-runner analysis-types

### LINKED ISSUE #358: Add html report analysis type
A common workflow for tools in the RD pipeline will be to produce a standalone HTML report per sample, family or project. Current examples of this include:
- The AIP - project level 
- The [MultiQC](https://github.com/populationgenomics/production-pipelines/blob/fafbe20d39a897b7e5fc570c0fbf9f3a6c693c8f/cpg_workflows/stages/cram_qc.py#L243) stage - project level
- The [STRipy](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/stripy.py) stage - sample level

Currently, these reports are written to the project web bucket but we have no generic way to make them discoverable/browseable by either our analyst team or our external collaborators.

## Long-term (ideal) goal
- We track report generation in metamist 
- We add generic functionality into seqr to display something like a "reports" section where report URLs can be listed alongside an informative description of the report
- Web bucket and seqr project permissions are harmonised so all accounts with access to a seqr project have read access to the web bucket.

## Immediate goal we would love in the short term
- Add a report analysis type into metamist
- Allow stages to register reports including the url and report name (eg stripy)
- Add a view to the metamist UI listing the links to available reports to make it easy for an analyst to browse to a specific report for a specific family.

illusional:
Closes #358.


danielreti (approved):
LGTM :)

violetbrina (approved):

PR #383, opened 2023-01-17
Lint B028 + new markdownlint concerns

illusional:
- I ran `precommit autoupdate`
- Then fixed all suggestions


danielreti (approved):
LGTM - just note that tests (and linting) are failing

PR #386, opened 2023-02-01
Add alternate implementation for file_exists

illusional:
- get_gcs_blob uses list_blobs and returns None when a blob with path name doesn't exist.

Implementation:

```python
    async def get_gcs_blob(self, filename: str) -> storage.Blob:
        """Convenience function for getting blob from fully qualified GCS path"""
        if not filename.startswith(self.GCS_PREFIX):
            raise ValueError('No blob available')

        bucket_name, path = filename[5:].split('/', maxsplit=1)
        bucket = self.get_gcs_bucket(bucket_name)

        # the next few lines are equiv to `bucket.get_blob(path)`
        # but without requiring storage.objects.get permission
        blobs = list(self.gcs_client.list_blobs(bucket, prefix=path))
        # first where r.name == path (or None)
        return next((r for r in blobs if r.name == path), None)
```

### DIFF
--- a/sample_metadata/parser/cloudhelper.py
+++ b/sample_metadata/parser/cloudhelper.py
@@ -99,6 +99,8 @@ class CloudHelper:
         """Determines whether a file exists"""
         path = self.file_path(filename)
         if path.startswith('gs://'):
+            # add a specific case for GCS as the AnyPath implementation calls
+            # bucket.get_blob which triggers a read (which humans don't have)
             blob = self.get_gcs_blob(path)
             return blob is not None
 
--- a/sample_metadata/parser/cloudhelper.py
+++ b/sample_metadata/parser/cloudhelper.py
@@ -99,7 +99,7 @@ class CloudHelper:
         """Determines whether a file exists"""
         path = self.file_path(filename)
         if path.startswith('gs://'):
-            blob = self.get_gcs_blob(filename)
+            blob = self.get_gcs_blob(path)
             return blob is not None
 
         return AnyPath(path).exists()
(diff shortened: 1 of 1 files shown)


sample_metadata/parser/cloudhelper.py
```diff
@@ -97,7 +97,12 @@ async def file_contents(self, filename) -> str | None:
 
     async def file_exists(self, filename: str) -> bool:
         """Determines whether a file exists"""
-        return AnyPath(self.file_path(filename)).exists()
+        path = self.file_path(filename)
+        if path.startswith('gs://'):
+            blob = self.get_gcs_blob(filename)
+            return blob is not None
```
lgruen, line 103: Couldn't you use https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.blob.Blob#google_cloud_storage_blob_Blob_exists instead?

As far as I remember, you can instantiate a blob directly (without calling `bucket.get_blob`, which you mention is problematic in `get_gcs_blob` below). That shouldn't require `get` permissions, I believe. Maybe that's also an opportunity to simplify `get_gcs_blob`, as long as callers don't assume that it includes an existence check.

lgruen (changes requested):

sample_metadata/parser/cloudhelper.py
illusional: I've simplified `get_gcs_blob` to instantiate it directly.

sample_metadata/parser/cloudhelper.py
```diff
@@ -97,7 +97,12 @@ async def file_contents(self, filename) -> str | None:
 
     async def file_exists(self, filename: str) -> bool:
         """Determines whether a file exists"""
-        return AnyPath(self.file_path(filename)).exists()
+        path = self.file_path(filename)
+        if path.startswith('gs://'):
```
lgruen, line 101: Maybe add a comment here why we have a special case for GCS?

lgruen (approved):

sample_metadata/parser/cloudhelper.py
illusional: Good idea, added a comment :)

PR #387, opened 2023-02-01
Add reference option to  in sample map parser

cassimons:
Allow a path to a reference genome to be set in the sample map parser script. This is needed for ingest of cram files.

### DIFF
--- a/scripts/parse_sample_file_map.py
+++ b/scripts/parse_sample_file_map.py
@@ -46,6 +46,10 @@ logger.setLevel(logging.INFO)
     help='Search path to search for files within',
 )
 @click.option(
+    '--ref',
+    help='default_reference_assembly_location',
+)
+@click.option(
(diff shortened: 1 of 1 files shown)


scripts/parse_sample_file_map.py
```diff
@@ -80,6 +85,7 @@ async def main(
         default_sequence_type=default_sequence_type,
         search_locations=search_path,
         allow_extra_files_in_search_path=allow_extra_files_in_search_path,
+        default_reference_assembly_location=ref
```
illusional, line 88: Linter wants a comma here

```suggestion
        default_reference_assembly_location=ref,
```

illusional (approved):

scripts/parse_sample_file_map.py
cassimons: Cheers. I need to sort out my black settings.

PR #390, opened 2023-02-01
Fix group by import

illusional:
Realistically, it's not practical to run from root scope as a module to ensure the package imports work correctly (which is why tests work), so easier to just redefine it inside the sample_metadata package, then it's perfectly scoped, and works on independent installations.


cassimons (approved):

illusional:
I'm going to merge and skip the linting because it's unrelated - and will follow-up in a different PR

PR #378, opened 2022-12-13
Sequencing groups

illusional:
A bigger change, outlined here: https://docs.google.com/document/d/1lS9JbUSK73ZLSvZmvwpOjH7daDYpoWyMx71vuzOq6XM/edit#

Review process:
- @vivbak: parsers
- @danielreti: graphql + interaction
- @violetbrina: backend changes

PR #393, opened 2023-02-02
Add default_sequence_technology to parser

cassimons:
Sequences are currently being ingested with a null technology field, resulting in a [type error](https://batch.hail.populationgenomics.org.au/batches/420919/jobs/1) when read by the api. This PR sets the default_sequence_technology in sample map parsers to 'short-read'.

This does not address why records could be written to the db with a null technology field. @illusional I assume we should add not null constraint be added to the db, or handling of null in the API?


illusional (approved):
Sounds good to me, I've addressed technology nullable concerns in the #378 PR.

PR #395, opened 2023-02-16
First step in addressing sequence technology

EddieLF:


illusional (approved):
Looks good (as long as linting + tests pass)

PR #399, opened 2023-02-21
Refactor and graphql

danielreti:
Switch to graphql for sample page, family page and project selector
Added pedigrees and family view
Large refactor following these doc suggestions - https://docs.google.com/document/d/1A9MWYjO8VTsYsAmG8DIb0ANWjAQBVVk3FEOODyA4V1Q/edit#

@illusional if you could look over the query side and any backend changes I made
@daniaki could you look at the refactor (components, file structure, error boundaries etc)
Could we also do a review session to refactor `TangledTree.tsx` together

Marking as draft for now as I anticipate some further changes before we merge


api/graphql/schema.py
```diff
@@ -122,7 +122,7 @@ async def load_families_for_participants(
 ) -> list[list['Family']]:
     flayer = FamilyLayer(connection)
     fam_map = await flayer.get_families_by_participants(participant_ids=participant_ids)
-    return [fam_map.get(p) for p in participant_ids]
+    return [fam_map.get(p) for p in participant_ids if p in fam_map] or [[]]
```
illusional, line 125: The dataloaders loaders here want you to always return the correct amount of elements, so a filter on a map here can break the loader, I presume the default should just be an empty list from the .get.

```suggestion
    return [fam_map.get(p, []) for p in participant_ids]
```

api/routes/family.py
```diff
@@ -132,7 +132,12 @@ async def get_pedigree(
     )
 
 
-@router.get('/{project}/', operation_id='getFamilies', tags=['seqr'])
+@router.get(
+    '/{project}/',
+    operation_id='getFamilies',
+    response_model=List[Family],
```
illusional, line 138: Response model here breaks the Python API, and it should be covered by the GraphQL API right (if not we should add it).
```suggestion
```

db/backup/README.md
```diff
@@ -4,31 +4,31 @@ This document defines the procedures defined for database recovery, backup and v
...
-   1. [Recover Sample Metadata Database](#Recover-Sample-Metadata-Database)
-      1. [Rebuild the VM](#Rebuild-VM)
-      2. [Install MariaDB 10.5](#Install-MariaDB-10.5)
-2. [Current Systems](#Current-Systems)
-   1. [Daily Backup](#Daily-Backup)
-   2. [Monitoring and Alerts](#Monitoring-and-Alerts)
-3. [Validation](#Validation)
-   1. [Validation Procedure](#Validation-Procedure)
-   2. [Validation Plan](#Validation-Plan)
+1. [Recovery Procedures](#recovery-procedures)
```
illusional, line 7: Might be worth merging from `dev`, because pretty sure this is fixed there.

db/backup/recovery_test.py
```diff
@@ -99,7 +99,7 @@ def test_row_number(self, table):
         )
         prod_count = get_results(
             self.prod_conn,
-            f"SELECT COUNT(*) FROM {table} FOR SYSTEM_TIME AS OF TIMESTAMP'{self.timestamp}'",
+            f'SELECT COUNT(*) FROM {table} FOR SYSTEM_TIME AS OF TIMESTAMP{self.timestamp!r}',
```
illusional, line 102: Same with these, this is fixed in `dev`.

models/models/search.py
```diff
@@ -33,9 +33,16 @@ class SampleSearchResponseData(SearchResponseData):
     sample_external_ids: list[str]
 
 
+class ErrorSearchResponseData(SearchResponseData):
```
illusional, line 36: Why subclass SearchResponesData, does the error belong to a project?
I would've expected just a straight:
```suggestion
class ErrorResponse(SMBase):
```

models/models/search.py
```diff
@@ -33,9 +33,16 @@ class SampleSearchResponseData(SearchResponseData):
...
+    error: str
+
+
 class SearchResponse(SMBase):
     """Response class for each search result"""
 
     type: SearchResponseType
     title: str
     data: SampleSearchResponseData | ParticipantSearchResponseData | FamilySearchResponseData
+    error: ErrorSearchResponseData | None
```
illusional, line 48: I think providing a default here means you don't have to specify it when constructing (additionally allowing it to be None):

```suggestion
    error: ErrorSearchResponseData | None = None
```

web/package.json
```diff
@@ -1,67 +1,100 @@
...
-      "react-app",
-      "react-app/jest"
-    ]
-  },
-  "browserslist": {
-    "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
+    "name": "sample-metadata",
```
illusional, line 2: Does the linting formatter cover this file, just noticed this frequently gets flicked between 2 and 4 spaces - would be good to try to reduce unrelated changes in the future.

web/src/pages/admin/ProjectsAdmin.tsx
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable */
```
illusional, line 1: Weird that git reports this as a delete and create rather than a move

illusional (changes requested):
Provided some suggestions on server side. 
Although generally supportive of the refactor with the PR (I know I approved that), this PR reports moves as delete + create, which makes it difficult to review meaningful changes to the components. 

Not sure what you can do about it now, but generally try to move the file, commit, then make more modifications to allow Git to recognise it as a move + updates.

You also need to merge from `dev`, I'd recommend doing this often, as you have lots of conflicts which will be fiddly to merge.

web/package.json
danielreti: Defs should be 4 spaces - haven't noticed the flicking but will keep an eye on it

web/src/pages/admin/ProjectsAdmin.tsx
danielreti: Yeah no idea why

danielreti (commented):
Responded
Removed Pedigree component into separate PR

web/src/shared/utilities/sanitiseValue.tsx
```diff
@@ -0,0 +1,10 @@
+const sanitiseValue = (value: unknown): string => {
```
daniaki, line 1: This is a plain typescript (`ts`) file. The `tsx` file extension should be reserved for files that render JSX.

web/src/shared/components/SeqInfo.tsx
```diff
@@ -0,0 +1,154 @@
+import * as React from 'react'
```
daniaki, line 1: There's a lot going on in this component, and the render logic can be hard to follow at times. I think it would be great to do a pair programming session where we can refactor and break it down into simpler components together. We can discuss some design strategies and their trade-offs. What do you think? @danielreti

web/src/pages/project/ProjectSummary.tsx
```diff
@@ -0,0 +1,174 @@
...
+    const getProjectSummary = React.useCallback(
+        async (token: any) => {
+            if (!projectName) {
+                setSummary(undefined)
+                return
+            }
+            const sanitisedToken = token || undefined
+            setError(undefined)
+            setIsLoading(true)
+            new WebApi()
```
daniaki, line 56: Just a comment - I find it much cleaner and idiomatic to use the `await` syntax inside `async` functions instead of `then/catch` promises. So this promise becomes:

```ts
try {
    const response = await new WebApi().getProjectSummary(projectName, pageLimit, sanitisedToken)
    setSummary(resp.data)
} catch (er) {
    setError(er.message)
} finally {
    setIsLoading(false)
}
```

web/src/pages/sample/SampleView.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    </h4>
+                    <SampleInfo
+                        sample={Object.fromEntries(
+                            Object.entries(data.sample).filter(([key]) =>
+                                sampleFieldsToDisplay.includes(key)
+                            )
+                        )}
+                    />
+                </div>
+                <SeqPanel isOpen sequences={data.sample.sequences} />
```
daniaki, line 97: Just curious - is `data.sample.sequences` guaranteed by the API to always be present? Might cause access errors if not. Also applies for other areas in the code with chained property access.

daniaki (changes requested):
Looks good overall, just a few minor comments.

web/src/shared/components/SeqInfo.tsx
danielreti: Let's do this together in Melbourne

web/src/pages/project/ProjectSummary.tsx
danielreti: Fixed, although this will migrate to GraphQL soon

web/src/pages/sample/SampleView.tsx
danielreti: Yeah it should always be present, could be an empty array though. These are all validated by the typescript graphql codegen

danielreti:
Responded to everything + merged dev again

illusional (approved):
Looks good to me!

web/src/pages/project/ProjectGrid.tsx
```diff
@@ -0,0 +1,138 @@
...
+    error,
+}) => {
+    if (!projectName) {
+        return (
+            <p>
+                <em>Please select a project</em>
+            </p>
+        )
+    }
+    if (error) {
```
daniaki, line 29: As a suggestion, I think that rendering the error and loading states should be a responsibility of the parent component that is fetching the data. Otherwise it seems like the data fetching abstraction is leaking into this child component which should only be responsible for rendering the summary data.

web/src/pages/project/ProjectGrid.tsx
```diff
@@ -0,0 +1,138 @@
...
+    projectName?: string
+    error?: string
+}
+
+const ProjectGrid: React.FunctionComponent<ProjectGridProps> = ({
+    summary,
+    projectName,
+    error,
+}) => {
+    if (!projectName) {
```
daniaki, line 22: I would change condition to `!projectName || Object.keys(summary ?? {}).length === 0` if following the comment below, to check for an empty summary data object.

web/src/pages/project/ProjectGrid.tsx
```diff
@@ -0,0 +1,138 @@
...
+        return (
+            <p>
+                <em>Please select a project</em>
+            </p>
+        )
+    }
+    if (error) {
+        return (
+            <p>
```
daniaki, line 31: Should all errors be a `MuckError` for consistency?

daniaki (changes requested):
Nice work on the change. Another minor nitpick.

danielreti:
Added suggested changes (roughly). Move non rending logic to parent component, including check for loading, errors and non-empty summary objects.

daniaki (approved):
LGTM!

PR #401, opened 2023-02-21
regex fix for parsing fastq with trailing digits

EddieLF:


cassimons (commented):
Looks good to me, but I will leave it to @illusional to approve.

sample_metadata/parser/generic_parser.py
```diff
@@ -73,7 +73,7 @@
 
 # construct rmatch string to capture all fastq patterns
 rmatch_str = (
-    r'[_\.-][Rr]?[12]('
+    r'[_\.-][Rr]?[12](_\d+)?('
```
illusional, line 76: Can you add some examples in the doctest down here with the new format?

https://github.com/populationgenomics/sample-metadata/blob/299bc966b16e9f57db80f0f8cbafe2635b721d6c/sample_metadata/parser/generic_parser.py#L1145

illusional (changes requested):

sample_metadata/parser/generic_parser.py
illusional: I suspect this won't correctly group on the second group (the 001 after the R\d)

EddieLF: @illusional not sure how to see if the doctest passed here but I've added an example with _R1/R2_001.fastq.gz, is this sufficient? Cheers

cassimons: That format is typically used to represent a sharded fastq - ie we want all files with the prefix `21R2112345-20210326-A00123_S70_L001_R1_*` to be merged in the final alignment.

cassimons: Put another way, they should end up in the same read group in the bam.

EddieLF: I have been practicing with the fastq parser to try and get this to work. So the current function creates pairs of R1 / R2 files based on their common filename prefix

With this change, do we want to create pairs of reads in both directions with a common prefix and a common suffix? 
i.e group `R1_001` with `R2_001` etc.
```
[
['L001_R1_001.fastq.gz', 'L001_R2_001.fastq.gz'],  
['L001_R1_002.fastq.gz', 'L001_R2_002.fastq.gz']
]
```

Or should the pairs of reads be the same direction but with different suffix?
i.e group `R1_001` with `R1_002`
```
[
['L001_R1_001.fastq.gz', 'L001_R1_002.fastq.gz'],  
['L001_R2_001.fastq.gz', 'L001_R2_002.fastq.gz']
]
```

Or is it neither of the above and i'm misunderstanding?

cassimons: As far as I know option 1 is correct, these are just normal fastq pairs. Unless I am missing something @illusional?

EddieLF:
@illusional @cassimons let me know what you think of this approach [c6728c1](https://github.com/populationgenomics/sample-metadata/pull/401/commits/c6728c10e5c56d0085112af7ee14fb8e98253a4b)

First I changed the regex to have a non-capturing group for the optional trailing digits. I also modified the fastq parsing method to call a new function `group_fastqs_by_common_filename_components`, which uses `group_by` to group the files based on:

- a common prefix (the filename component before R1/R2 or 1/2), and 
- a common suffix, if applicable (the component after R1/R1 or 1/2 - for filenames like `L001_R1_001.fq.gz`).

I think this succeeds for all the cases included in the doctests and those from Michael's example above, but I'm sure it could be improved.

sample_metadata/parser/generic_parser.py
```diff
@@ -1267,6 +1284,40 @@ def guess_delimiter_from_filename(filename: str):
...
+        'S70_L001_R2_001.fastq.gz': ('S70_L001_', 'R2_001.fastq.gz'),
+        'S70_L001_R1.fastq.gz': ('S70_L001_', 'R1.fastq.gz'),    
+        'S70_L001_R2.fastq.gz': ('S70_L001_', 'R1.fastq.gz')
+        }
+    Sample output: {
+        'S70_L001_001.fastq.gz': ['S70_L001_R1_001.fastq.gz', 'S70_L001_R2_001.fastq.gz'],
+        'S70_L001_fastq.gz': ['S70_L001_R1.fastq.gz', 'S70_L001_R2.fastq.gz']
+        }
+    """
+    def key_selector(kv: Tuple[str, str]) -> Tuple[str, str]:
```
illusional, line 1304: In Python3.9, you can directly use the type.

```suggestion
    def key_selector(kv: tuple[str, str]) -> tuple[str, str]:
```

sample_metadata/parser/generic_parser.py
```diff
@@ -1267,6 +1284,40 @@ def guess_delimiter_from_filename(filename: str):
...
+        k, v = kv
```
illusional, line 1305: Can you use more descriptive variable names here?

illusional (changes requested):
Yep, this looks great to me. Requesting changes for clarifying variable names + linting, if you install pre-commit `pip install pre-commit`, running pre-commit locally `pre-commit run --all` will format and lint the files.

EddieLF:
@illusional thanks for the feedback. Can this branch still be merged with the unit test failing? Its for an unchanged file so hopefully its fine.

illusional (approved):

PR #402, opened 2023-02-21
Analysis cleaner

EddieLF:
This is coming late and unfinished, its a bit of a mess right now. 

I'm going to complete the outstanding ingestion tasks today and return to this later today. If either of you @MattWellie @illusional have time to take a look and provide any feedback that's much appreciated


sample_metadata/parser/generic_parser.py
```diff
@@ -73,7 +73,7 @@
 
 # construct rmatch string to capture all fastq patterns
 rmatch_str = (
-    r'[_\.-][Rr]?[12]('
+    r'[_\.-][Rr]?[12](_\d+)?('
```
illusional, line 76: It looks like this change got included in this PR, not a big deal, but may cause a conflict when the original PR is merged

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+
+    all_analyses = defaultdict(str)
+    # map each analysis id to its output path
+    for analysis in analyses:
+        all_analyses[analysis['id']]=analysis['output']
+    
+    # map each output path to the analysis entries which point to it
+    reversed_analyses_dict = {}
+    for k, v in all_analyses.items():
+        reversed_analyses_dict.setdefault(v, set()).add(k)
```
illusional, line 151: You can use a defaultdict here to achieve the same thing, would recommend more descriptive variable names here based on the above:

```suggestion
    reversed_analyses_dict = defaultdict(set):
    for analysis_id, analysis_output_path in all_analyses.items():
        reversed_analyses_dict[analysis_output_path].add(analysis_id)
```

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+
+    duplicate_analyses = [vals for vals in reversed_analyses_dict.values() if len(vals)>1]
+    
+    # keep the first entry pointing to a path, duplicates can be deleted
```
illusional, line 155: For the lols, can you report the output with the highest number of duplicate entries?

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # bad_mappings = [] # Use logger for this
+
+    for analysis in analyses:
+        project_id = analysis['project']
+        project_name = project_id_map[project_id]
+        dataset_name = get_path_components_from_path(analysis['output'])['dataset']
+        if dataset_name == project_name:
+            continue
+        else:
```
illusional, line 175: Don't need the `else` here as the if: continue early returns (love early returns)

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+    for p in paths:
+        bucket_name = p.removeprefix('gs://').split('/', maxsplit=1)[0]
+        paths_by_bucket[bucket_name].append(p)
+
+
+    valid_paths = {}
+    invalid_paths = {}
+    # Check if each subdirectory/path in a bucket is valid or not
+    for bucket_name, paths in paths_by_bucket.items():
+        vp, ivp = validate_paths_for_bucket(bucket_name, paths)
```
illusional, line 197: The linter probably wants longer, less ambiguous variable names here

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+            continue
+        
+        files_by_subdir[subdir] = get_paths_for_subdir(client, bucket_name, subdir) 
+
+    # paths = all paths found in Metamist for a given bucket name from analysis['output']
+    # files_by_subdir = all paths found in gcloud storage for a given bucket grouped by bucket subdir
+    files_set = set([e for sublist in files_by_subdir.values for e in sublist])
+
+    # Has analysis entry but file moved or deleted:
+    analysis_entries_missing_files = path_set.difference(files_set)
```
illusional, line 227: Nipick, but can you make these variable names use `path` instead of `entries` just to be clear about what the type is, eg: `analysis_paths_missing_files`

illusional (commented):
Looking like a great base! Let me know if you want clarity on any comments :)

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+
+    >>> get_bucket_name_path_from_path('gs://cpg-dataset-main/subfolder/subfolder2/my.cram')
+    {bucket_name:'cpg-dataset-main', dataset:'dataset', subdir:'subfolder/subfolder2'}
+    """
+    if path.startswith('gs://'):
+        short_path = path.removeprefix('gs://').split('/', maxsplit=1)
+        bucket_name = short_path[0]
+
+        dataset_with_bucket_type = bucket_name.removeprefix('cpg-')
+        bucket_type = dataset_with_bucket_type.split('-')[-1]
```
MattWellie, line 128: we have single and double-suffixed buckets: [see](https://github.com/populationgenomics/team-docs/tree/main/storage_policies)

It would be good to slot in some test cases for `-upload`, `-analysis`, `-web` variants

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
+from collections import defaultdict
+
+from sample_metadata.apis import ProjectApi, AnalysisApi
+from sample_metadata.model.analysis_query_model import AnalysisQueryModel
+from sample_metadata.model.analysis_status import AnalysisStatus
+from sample_metadata.model.analysis_type import AnalysisType
+
+import logging
```
MattWellie, line 8: You're importing logging, but also including print statements and `warnings` - could this all be done using logging instead?

You can also set logging to show the line number the printed statement is coming from, which can help when debugging ([e.g.](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/reanalysis/interpretation_runner.py#L451))

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+    if path.startswith('gs://'):
+        short_path = path.removeprefix('gs://').split('/', maxsplit=1)
+        bucket_name = short_path[0]
+
+        dataset_with_bucket_type = bucket_name.removeprefix('cpg-')
+        bucket_type = dataset_with_bucket_type.split('-')[-1]
+        dataset = dataset_with_bucket_type.removesuffix(bucket_type)[:-1]
+
+        file_name = short_path[1].split('/')[-1]
+        subdir = short_path[1].removesuffix(file_name)[:-1]
```
MattWellie, line 132:
```suggestion
        file_name, subdir = short_path[1].rsplit('/', 1)
```

as well as split, you can do an rsplit, with an optional `maxsplit` argument - it performs the same way, but breaks from right to left instead

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+
+    else:
+        raise warnings.warn(f'Analysis path {path} not bucket storage.')
```
MattWellie, line 135: You can put this condition first, `if not path.startswith('gs://')...`

That removes the need to indent the rest of the function lines

scripts/cleanup_analysis_entries.py
MattWellie, line 151:
```suggestion
    all_analyses = defaultdict(set)
    
    # map each analysis id to its output path
    for analysis in analyses:
        all_analyses[analysis['output']].add(analysis['output'])
```

with a default dict there's no need to collect then reverse the ordering

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+        reversed_analyses_dict.setdefault(v, set()).add(k)
+
+    duplicate_analyses = [vals for vals in reversed_analyses_dict.values() if len(vals)>1]
+    
+    # keep the first entry pointing to a path, duplicates can be deleted
+    duplicate_analyses_ids = []
+    for duplicate_ids in duplicate_analyses:
+        duplicate_analyses_ids.append(sorted(duplicate_ids)[1:])
+
+    ids_to_delete = [e for sublist in duplicate_analyses_ids for e in sublist]
```
MattWellie, line 160:
```suggestion
    ids_to_delete = {}
    for duplicate_ids in duplicate_analyses:
        ids_to_delete.update(sorted(duplicate_ids)[1:])
```

Alternatively the `itertools.chain.from_iterable` method is a neat way to avoid this double layered list comp

```
from itertools import chain
ids_to_delete = chain.from_iterable(duplicate_analyses_ids)
```

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # paths = all paths found in Metamist for a given bucket name from analysis['output']
+    # files_by_subdir = all paths found in gcloud storage for a given bucket grouped by bucket subdir
+    files_set = set([e for sublist in files_by_subdir.values for e in sublist])
+
+    # Has analysis entry but file moved or deleted:
+    analysis_entries_missing_files = path_set.difference(files_set)
+
+    # valid entries are all analysis output paths that intersect with all gcp output paths
+    valid_entries = path_set.intersect(files_set)
```
MattWellie, line 230:
```suggestion
    valid_entries = path_set.intersection(files_set)
```

I might be wrong about this...

scripts/cleanup_analysis_entries.py
MattWellie: Alternatively if you process each dataset's analysis entries separately you can just keep hold of the dataset name, so you don't need to worry about how to identify

```
gs://cpg-acute-care-main-analysis/
vs.
gs://cpg-acute-care-main/
```

MattWellie (commented):
Sorry it took ages to get around to this!

EddieLF:
Thanks a lot for the feedback both @illusional @MattWellie. 

The script now seems to be working mostly as it should, with room for improvements. I think the outputs could be better than just printing/logging. Also not really sure how to implement deletion of objects from Metamist and honestly a bit scared to try :^)

A few key points: 
 - 20,977 analysis entries found with get_seqr_projects()
 - 16,467 entries are duplicates
 - = 4,510 non-duplicate entries
 - Most duplicated path is in acute-care with 38 total entries pointing to it
 
 - 2,658 "valid" entries that actually lead to a file in the gcp cloud
 - (I feel like something is going wrong here in validate_paths, since 4510-2658=1852 which is not the count of invalid paths)

illusional:
Hey @EddieLF, are how many are entries for which the files don't exist anymore, and are any samples left without attached crams?

scripts/cleanup_analysis_entries.py
```diff
@@ -0,0 +1,378 @@
...
+
+    split_bucket_name = dataset_with_bucket_type.split('-')
+    for bucket_component in split_bucket_name:
+        if bucket_component in BUCKET_TYPES:
+            bt_index = split_bucket_name.index(bucket_component)
+
+    # bucket_type = '-'.join(split_bucket_name[bt_index:])
+    dataset = '-'.join(split_bucket_name[:bt_index])
+
+    subdir, _ = short_path[1].rsplit('/', 1)
```
MattWellie, line 75: If you were feeling brave this would be an ideal regex situation...

```
import re

pattern = re.compile(
    r'gs://cpg-(?P<bucket>[a-zA-Z0-9\-]+)(?=-main|-test|-release|-archive)[a-zA-Z0-9\-]+/(?P<suffix>.+)/(?P<file>.*)'
)

test = 'gs://cpg-acute-care-main-web/reanalysis/folder/subfolder/aip_index.html'
mat = pattern.match(test)
print(mat.groups())

> ('acute-care', 'reanalysis/folder/subfolder', 'aip_index.html')
```

link to a RegEx101 sandbox: https://regex101.com/r/DQzojH/4

illusional: I like the regex, I wouldn't match on specific bucket categories, just match anything after `main/test` as category if you wanted to return that.

EddieLF:
> Hey @EddieLF, are how many are entries for which the files don't exist anymore, and are any samples left without attached crams?

@illusional of the 4,510 detected non-duplicate paths, only 2,875 are actually not duplicates once their cram output paths have been parsed by the json part. I will need to change the script to remove the duplicates after this step too. 

Of these 2,875 entries, 217 have paths which lead to files that don't exist anymore. I also picked up 3 crams in the subdirectories searched which have no analysis entry pointing to them. 
Not sure what you mean about samples without attached crams, are you suggesting to use the sample api to do a similar audit of sample entries?

scripts/cleanup_analysis_entries.py
MattWellie: I was struggling to find a bulletproof way to make sure that the dataset name was separated from the bucket suffix without a lookahead match, given that the datasets also have an arbitrary number of hyphens 😓

EddieLF: @MattWellie thanks heaps for the base to work from. This is what I came up with [regex101](https://regex101.com/r/94WYPd/1), it puts the bucket extension after main/test into its own group and then the code will handle if that group is none.

MattWellie: I've tweaked that a bit further [here](https://regex101.com/r/94WYPd/2), but it's not clear to me from current usage of that method whether it's important to capture the extension separately from the rest of the bucket name (or if that's even valid to do, given that `cpg-dataset-main` and `cpg-dataset-main-web` are different `buckets`).

I've made a tweak on the original one [here](https://regex101.com/r/DQzojH/6) which separately captures the dataset, full bucket, suffix, and file (if required).

EddieLF:
@illusional just pushed a change which looks at the samples that don't have an attached cram, I'm seeing just 25 total samples where the cram from the analysis output is missing. 

Thoughts on dumping some of this output into CSVs? We could have one for duplicate analysis paths, one for invalid analysis paths, one for duplicate samples, one for samples without attached crams, etc. 

Would having those flat files created by running the code be useful for updating the Metamist entries?

PR #404, opened 2023-02-23
Added TangledTree component

danielreti:
Separated TangledTree into its own stack PR to separate from refactor PR. 
This component is fully functional but very messy and I think could do with a big refactor before merging.


danielreti:
I can write an outline of the logic behind generating the pedigrees if that is of use

illusional (commented):
\+1 to writing some docs for this to explain how this comes together, bonus points for examples and tests.
Merging from `dev` will get rid of that unittest failure.

danielreti:
https://docs.google.com/document/d/1k5fYHsFleUiIDaDZxROk2BthLQfuTTXelwQVvzQ5Crg/edit?usp=sharing @illusional @daniaki

web/src/shared/components/pedigree/TangledTree.tsx
```diff
@@ -11,19 +15,672 @@ interface PedigreeEntry {
...
+            }, {})
+        )
+    }, [data])
+
+    return (
+        <>
+            {(!data || !data.length) && (
+                <MuckError message={`Ah Muck, there's no data for this pedigree!`} />
+            )}
+            {data && !!data.length && !tree && (
```
daniaki, line 672: I think it would be better if you handle the loading and error states outside of this component, and let this responsibility be on the parent component, for example:

```es6
# In parent component

if (error) return <MuckError message={error.message} />
if (loading) return <LoadingDucks />

const pedigreeData = data?.family.project.pedigree
if (!pedigreeData) {
  return <div>Pedigree data not available</div>
} else {
  return <TangledTree data={pedigreeData} click={onClick} />
}
```

Then I would delete most of the rendering logic here and simply have:

```es6
# In this component

if (tree && keyedData) {
  return renderChart(tree, keyedData, onClick)
} else {
  return <div>No pedigree to show</div>
}
```

web/src/shared/components/pedigree/TangledTree.tsx
```diff
@@ -11,19 +15,672 @@ interface PedigreeEntry {
...
+    const [keyedData, setKeyedData] = React.useState<{
+        [name: string]: PedigreeEntry
+    }>()
+    const onClick =
+        click ??
+        function defaultClick() {
+            return undefined
+        }
+
+    React.useEffect(() => {
```
daniaki, line 657: You don't need a `useEffect` or `useState` hooks. Try using derived properties instead, which will be re-computed any time that the props change (like you have done with `onClick`). This will clean up your code considerably.

```es6
const tree: NodeList[] = data ? formatData(data) : []
const keyedData: {[name: string]: PedigreeEntry} = data ? data.reduce(..., {}) : {}
```

daniaki (changes requested):

dancoates:
Closing as this was merged in #771

PR #411, opened 2023-02-28
Additional swagger update fields for samples and participants

EddieLF:
Hi @vivbak  & @illusional,

These requested changes are so that I can use swagger to update a sample's external id, such as for the case below where I want to remove the trailing whitespace from the external_id field.
```
{
  "id": "CPG12345",
  "external_id": "D01-2345 ",
  "participant_id": 123
}
````

The other change is to allow updating a participants family id, for participants who were mistakenly linked to the wrong family_id in the initially provided pedigree data. 

Will these changes allow for these fields to be updated with swagger after the next deployment? 
Thanks :)

edit: Noting the pre-commit checks failing, I'll try and fix them now and push a new commit. Gonna remember to run pre-commit from now on!!

### CODE REVIEW

File: api/routes/participant.py [hunk 1]
```diff
@@ -179,6 +179,7 @@ async def update_participant(
         'success': await participant_layer.update_single_participant(
             participant_id=participant_id,
             external_id=participant.external_id,
+            family_id=participant.family_id,
```
illusional, line 182: This isn't plumbed up like this, and unfortunately not how the families are modelled. I'd probably recommend removing the familyID update and we can chat about how to update families separately.

illusional (changes requested):
Family IDs aren't easily updated, and this needs a `bumpversion` commit as this changes the API (any change to routes / models).

api/routes/participant.py
EddieLF: ok sounds good, I'll remove those changes and we'll see how we go with the sequence external_id update

illusional (approved):

vivbak:
Awesome to see you making direct changes to the code base already @EddieLF! 🎉 🙌

PR #412, opened 2023-02-28
Small fixes

illusional:
- Speedup get_projects call 
- Clarify all_access method
- Make SM options clearer (by consolidating more to `settings.py`)

No bumpversion required as no change to api spec.


danielreti (approved):
LGTM :)

PR #416, opened 2023-03-06
Fastq parser fix

EddieLF:
This fix is for new filename formats which weren't working with the last iteration of the parser due to ambiguity in the location of the read pair identifier in the file name. The regex pattern has been altered to use a positive lookahead to identify the R1/R2/1/2 read pair identifier, then an optional _001/_002 is checked for after the read pair identifier.

Previously, for a file ending with something like `_2_1.fq` or `_2_2.fq`, the first digit would be incorrectly parsed as the read pair identifier. With this change, the second digit in the above examples is parsed as the read pair identifier, and the first digit is correctly identified as part of the filename prefix. A doctest has been added to show this functionality. 

Some examples of the regex working can be seen [here](https://regex101.com/r/nGZPmQ/1).


EddieLF:
I'll suss the breaking unit tests and get back to this

sample_metadata/parser/generic_parser.py
```diff
@@ -1187,11 +1191,12 @@ def parse_fastqs_structure(fastqs) -> List[List[str]]:
...
         for full_filename, (basename, matched) in r_matches.items():
             # use only file path basename to define prefix first
-
-            pre_r_basename = basename[: matched.start()]
+            pre_r_basename = basename[: matched.end()]
             bits_to_group_on = [pre_r_basename]
-            for group in matched.groups():
-                bits_to_group_on.append(group)
+            groups = matched.groups()
+            for i in range(1, 3):
```
illusional, line 1197:
```suggestion
            # group fasts based on the regex groups 1 / 2
            for i in (1, 2):
```

sample_metadata/parser/generic_parser.py
```diff
@@ -1154,6 +1155,9 @@ def parse_fastqs_structure(fastqs) -> List[List[str]]:
         >>> GenericParser.parse_fastqs_structure(['21R2112345-20210326-A00123_S70_L001_R1_001.fastq.gz', '21R2112345-20210326-A00123_S70_L001_R2_001.fastq.gz'])
         [['21R2112345-20210326-A00123_S70_L001_R1_001.fastq.gz', '21R2112345-20210326-A00123_S70_L001_R2_001.fastq.gz']]
 
+        >>> GenericParser.parse_fastqs_structure(['ACG0xx_2_1.fastq.gz', 'ACG0xx_2_2.fastq.gz', 'ACG0xx_3_1.fastq.gz', 'ACG0xx_3_2.fastq.gz'])
```
cassimons, line 1158: Who comes up with these crazy file name formats?!

sample_metadata/parser/generic_parser.py
MattWellie: 🤯  These FQ file names are nuts

sample_metadata/parser/generic_parser.py
MattWellie: 🤯 These FQ file names are nuts

illusional (approved):
You'll want to add a bumpversion commit. ngl, I don't really understand lookaheads, but the tests pass so I'm happy.

PR #426, opened 2023-03-15
Sequence type update via api endpoint

EddieLF:
It'd be great if we could update the sequence type field with the api endpoint since this doesn't seem to be possible currently. 

I've added the field to the `update_sequence` call itself in the api route, as well as in the db layer. Then in the sequence table definitions, I've added sequence type to the updaters and fields so it should get included with the update query when this is called.

@illusional @violetbrina do you think these changes will enable this field to be changed? Thanks!


EddieLF:
not sure where all these other commits came from, I think because I branched from main not dev. 

Anyway, I think they can be ignored for now, if the change is good i'll branch it properly from dev and create a PR to merge back into dev.

illusional (approved):
Need a bumpversion commit to release, but looks good to me.

PR #427, opened 2023-03-15
Sequence type update

EddieLF:
Adding in the changes to the sequence backend and API endpoint + bumpversion to 5.5.7


illusional (approved):
Missed the incorrect branching from the other PR, thanks for fixing :)

PR #429, opened 2023-03-24
added filters to front-end

danielreti:
Added filters to columns of project grid


danielreti:
This is ready for review :)
Cannot figure out why the prettier linting is failing - it's giving me green lights across the board locally

illusional (changes requested):

db/python/layers/web.py
```diff
@@ -97,35 +98,75 @@ class WebLayer(BaseLayer):
...
+                        case 'sequence':
+                            prefix = 'sq'
+                        case 'participant':
+                            prefix = 'p'
+                        case 'sample':
+                            prefix = 's'
+                        case 'family':
+                            prefix = 'f'
+                            field = 'external_id'
+                    if field == 'created_date':
```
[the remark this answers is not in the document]
danielreti: `FROM sample FOR SYSTEM_TIME ALL` is in the query and works fine?

api/routes/web.py
```diff
@@ -72,22 +73,24 @@ class SearchResponseModel(BaseModel):
...
 
-@router.get(
+@router.post(
     '/{project}/summary',
     response_model=ProjectSummaryResponse,
     operation_id='getProjectSummary',
 )
 async def get_project_summary(
     request: Request,
+    grid_filter: dict,
```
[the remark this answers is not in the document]
danielreti: Could you help me with this, not entirely sure how you want the class to look so that we can get rid of the match

db/python/layers/web.py
```diff
@@ -326,6 +368,12 @@ async def get_project_summary(
             ),
             atable.get_seqr_stats_by_sequence_type(project=self.project),
         )
+        query_values = values
+        query_values['limit'] = total_samples
+        query_values['after'] = 0
+        total_samples_in_query = len(
```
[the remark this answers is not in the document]
danielreti: This needs to be executed for pagination, we need to know how many samples matched the filter so we can paginate correctly. This number will always be equal to or less than the total_number of samples in the project, so it makes sense to pass this number as the limit.

test/test_web.py
```diff
@@ -57,14 +57,15 @@ async def setUp(self) -> None:
     @run_as_sync
     async def test_project_summary(self):
         """Test getting the summary for a project"""
-        result = await self.webl.get_project_summary(token=None)
+        result = await self.webl.get_project_summary(token=None, grid_filter={})
```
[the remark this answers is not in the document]
danielreti: Will add some now

api/routes/web.py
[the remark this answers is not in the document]
danielreti: Had a crack?

db/python/layers/web.py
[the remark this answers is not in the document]
danielreti: Removed

db/python/layers/web.py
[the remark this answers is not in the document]
danielreti: Fixed

db/python/layers/web.py
```diff
@@ -16,9 +17,29 @@
...
+    p = 'participant'
+    s = 'sample'
+    sq = 'sequence'
+    f = 'family'
+
+
+class SearchItem(SMBase):
+    """Summary Grid Filter Model"""
+
+    model_type: str
```
illusional, line 37: Is it this?

```suggestion
    model_type: MetaSearchEntityPrefix
```

db/python/layers/web.py
```diff
@@ -97,35 +119,61 @@ class WebLayer(BaseLayer):
...
+                q = f'{prefix}.{field} LIKE :{key}'
+            else:
+                q = f'JSON_VALUE({prefix}.meta, "$.{field}") LIKE :{key}'
+            wheres.append(q)
+            values[key] = self.escape_like_term(value) + '%'
         if wheres:
             where_str = 'WHERE ' + ' AND '.join(wheres)
-        sample_query = f'SELECT id, external_id, type, meta, participant_id FROM sample {where_str} ORDER BY id LIMIT :limit OFFSET :after'
 
+        sample_query = f"""
```
illusional, line 166: Can you add a comment here about why we skip the limit?

db/python/layers/web.py
```diff
@@ -257,7 +306,9 @@ async def get_project_summary(
         )
         seqr_links = self.get_seqr_links_from_project(project)
 
-        sample_rows = list(await self.connection.fetch_all(sample_query, values))
+        sample_rows_all = list(await self.connection.fetch_all(sample_query, values))
```
illusional, line 309: Can you add a comment here and say no projects even have close to 10,000 rows, so this shouldn't be a memory / resource problem. But could be optimised in the future by limiting a random number, and then just saying "Over 10k" if the results have that many?

regenerate_api.py
```diff
@@ -299,7 +299,7 @@ def main():
 
         shutil.copy(
             './resources/muck-the-duck.svg',
-            os.path.join(STATIC_DIR, 'muck-the-duck.svg'),
+            os.path.join('web/src', 'muck-the-duck.svg'),
```
illusional, line 302: Why the change for this? Is this because of my change?

web/src/pages/project/ProjectGrid.tsx
```diff
@@ -1,7 +1,13 @@
 import * as React from 'react'
 import _ from 'lodash'
-import { Table } from 'semantic-ui-react'
+import { Table, Form, Popup } from 'semantic-ui-react'
```
illusional, line 3: This might've changed from my PR, using a different Table. 
Edit: On reflection you need to merge from main)

illusional (approved):
Few minor things before addressing, but looks good to me!

db/python/layers/web.py
danielreti: I think I'm using enums not quite right - If I do it as you suggest, the frontend complains that the model_type is a string instead of of type MetaSearchEntityPrefix when I try to pass it to the back. Would appreciate a very quick double check of this

regenerate_api.py
danielreti: The duck moved from `web/src/static` to` web/src` when we migrated to Vite but haven't needed to regenerate the api since then so was an oversight until now

web/src/pages/project/ProjectSummary.tsx
```diff
@@ -81,6 +84,44 @@ const ProjectSummary: React.FunctionComponent = () => {
...
+    const updateFilters = React.useCallback(
+        (e) => {
+            if (!summary) return
+            /* eslint-disable no-param-reassign */
+            const processedFilter = Object.entries(e).reduce((filter, [column, v]) => {
+                if (!v) {
+                    return filter
+                }
+                const query = v as string
+                const participantIndex = summary.participant_keys.findIndex((i) => i[1] === column)
```
illusional, line 96: If you're filtering on a meta field that also is a first class field - this finds the non-meta field, but also causes both to highlight:

image

web/src/pages/project/ProjectGrid.tsx
```diff
@@ -11,18 +17,117 @@ import { ProjectSummaryResponse } from '../../sm-api/api'
...
+                            )
+                        }
+                        return (
+                            <SUITable.HeaderCell
+                                key={`filter-${k}-${i}`}
+                                style={{ borderBottom: 'none' }}
+                            >
+                                <div style={{ position: 'relative' }}>
+                                    <div style={{ position: 'absolute', top: 0, right: 0 }}>
+                                        <Popup
```
illusional, line 83: On further review, I don't think I'm happy with this approach - can you bind more context in here (meta or not) so you don't look up k in filter values, I think it's worth being more concrete about how columns relate to filters rather than looking it up based on the name.

illusional (changes requested):
Can we not clear the whole grid if the grid is empty. Can we show the default columns + any columns that are filtered, so you don't have to clear all if there no are results

image

web/src/pages/project/ProjectGrid.tsx
```diff
@@ -60,29 +70,37 @@ const ProjectGrid: React.FunctionComponent<ProjectGridProps> = ({
...
                         return (
                             <SUITable.HeaderCell
-                                key={`filter-${k}-${i}`}
+                                key={`filter-${title}-${i}`}
                                 style={{ borderBottom: 'none' }}
                             >
                                 <div style={{ position: 'relative' }}>
                                     <div style={{ position: 'absolute', top: 0, right: 0 }}>
                                         <Popup
                                             trigger={
```
illusional, line 104: Small request to make it easier to get into the search box
```suggestion
                                            position='top center'
                                            trigger={
```

illusional (approved):
Looks great! Just need to make the column names unique and it's all sweet!

PR #434, opened 2023-03-29
Dim mode for swagger

illusional:
I can't work out how to change the `/docs` url that fastapi puts out, but swagger itself doesn't support dark mode.


danielreti (approved):
Poor mans dark mode = dim mode

PR #435, opened 2023-03-30
Fix update sequence type

EddieLF:
So I tried this a few weeks ago but I think I got the line in `tables/sequence.py` with `updaters.append('field = :field')` the wrong way around, and there was confusion between the fields named both `type` and `sequence_type` in different sections of the code.

This time, I've standardised the field names from `sequence_type` -> `type`. The `update_sequence` API route calls the sequence layer's `update_sequence` function, which calls the sequence table's `update_sequence` function. All 3 calls now use the field name `type`, so they all align with the `update_sequence` API endpoint. The sequence type should now be update-able through the API endpoint.

I've had to add some ignore lines to the linter for redefining the builtin `type`. Would this be ok? It's only redefined in the layer and table `update_sequence` functions which don't make use of the `type` builtin.


db/python/tables/sequence.py
```diff
@@ -377,9 +377,9 @@ async def update_sequence(
             if technology:
                 updaters.append('technology = :technology')
                 fields['technology'] = technology.value
-            if sequence_type:
-                updaters.append('sequence_type = :type')
-                fields['sequence_type'] = sequence_type.value
+            if type:
+                updaters.append('type = :type')
+                fields['type'] = type.value
```
illusional, line 382: These two lines are the only ones that need to change, redefining the builtin type is a little risky so would prefer to stray away from doing that.

illusional (changes requested):

db/python/tables/sequence.py
EddieLF: I think that's totally fair. I've reverted all the other changes except for the above lines, now the updater mostly refers to the field as sequence_type.

illusional (approved):
Please address change before merging

db/python/tables/sequence.py
```diff
@@ -378,8 +378,8 @@ async def update_sequence(
                 updaters.append('technology = :technology')
                 fields['technology'] = technology.value
             if sequence_type:
-                updaters.append('sequence_type = :type')
-                fields['sequence_type'] = sequence_type.value
+                updaters.append('type = :sequence_type')
+                fields['type'] = sequence_type.value
```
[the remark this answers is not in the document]
EddieLF: Cheers @illusional, I was getting bogged down in the difference uses of `type` vs `sequence_type`, how they related to the `fields` dictionary, and the `field = :field` statements. 

Now knowing the `fields` dict keys are the `:field` components actually made it a lot clearer.

PR #437, opened 2023-04-04
Catch issues with header 

vivbak:
Trying to be more graceful & helpful.


illusional (approved):

illusional:
You may want a bumpversion commit, but not completely necessary

PR #445, opened 2023-04-06
Fix sample ids in get_sequence_by_criteria 

vivbak:
Previously the raw ids were returned instead of the full CPG identifiers.

### DIFF
--- a/api/routes/sequence.py
+++ b/api/routes/sequence.py
@@ -149,8 +149,8 @@ async def get_sequences_by_criteria(
         else None,
     )
 
-    for sequence_group in result:
-        sequence_group.sample_id = sample_id_format(sequence_group.sample_id)
+    for sequence in result:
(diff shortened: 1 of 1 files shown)


api/routes/sequence.py
```diff
@@ -149,6 +149,9 @@ async def get_sequences_by_criteria(
         else None,
     )
 
+    for sequence_group in result:
+        sequence_group.sample_id = sample_id_format(sequence_group.sample_id)
```
illusional, line 153: I'll keep this like this so I don't get confused in my branch
```suggestion
    for sequence in result:
        sequence.sample_id = sample_id_format(sequence.sample_id)
```

illusional (approved):

api/routes/sequence.py
vivbak: haha sorry I was trying to be good!!

PR #452, opened 2023-04-11
draft sample analysis finder

EddieLF:
Context: https://github.com/populationgenomics/analysis-runner/pull/613, https://github.com/populationgenomics/seqr-private/issues/73

Output: https://docs.google.com/spreadsheets/d/1_Urw0rRMdo1SyhpVWyUynj5mQcwtvYf0bn-S_3WwA2A/edit?usp=sharing

This script takes a project and sample IDs, finds all the analysis entries associated with the samples, and creates an output spreadsheet containing these entries. 

It then takes the paths from the CRAM and GVCF analysis entries and searches the bucket at these paths for any extra files (e.g. cram.crai, cram.md5, g.vcf.gz.tbi) and adds these to the output. Currently duplicate CRAM and GVCF paths are added but it probably doesn't matter too much for now.

The output spreadsheet can then be used to investigate and eventually delete objects in the buckets via Analysis Runner.

### DIFF
--- a/scripts/find_sample_files_to_delete.py
+++ b/scripts/find_sample_files_to_delete.py
@@ -1,0 +1,213 @@
+""" This script finds analysis entries for a list of samples
+    and checks the bucket for extra cram/gvcf files.
+    Outputs a csv with all analyses for each sample and
+    the paths to any extra files.
+    """
+
+import logging
+import click
+import csv
+import subprocess
+
+from collections import defaultdict
+from os.path import basename
+from google.cloud import storage
+
+from sample_metadata.apis import AnalysisApi
+from sample_metadata.model.analysis_query_model import AnalysisQueryModel
+from sample_metadata.model.analysis_type import AnalysisType
+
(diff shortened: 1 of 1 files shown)


EddieLF:
@illusional you mentioned that using only the Sample ID to find the files might be unreliable, do you think this method needs to be changed to pull in SampleSequence IDs or something of the sort?

illusional:
This approach is all good for now, eventually analysis will be keyed by the "sequencing group ID", which the CPG ID will become, I'm just signposting this change (and prompting to use metamist to fetch the data).

EddieLF:
> This approach is all good for now, eventually analysis will be keyed by the "sequencing group ID", which the CPG ID will become, I'm just signposting this change (and prompting to use metamist to fetch the data).

Good to know for future. This issue should hopefully be temporary, since we have a planned change to the metadata parser which will check sequence file sizes and flag if there's a potential genome/exome labelling issue in the metadata

PR #454, opened 2023-04-11
Handle Invalid Sample Name/ Missing Sequences 

vivbak:
Addresses https://github.com/populationgenomics/sample-metadata/issues/450 

Adds a try-except to provide more info and exit more gracefully.

### DIFF
--- a/scripts/parse_existing_cohort.py
+++ b/scripts/parse_existing_cohort.py
@@ -146,22 +146,20 @@ class ExistingCohortParser(GenericMetadataParser):
 
     def get_sequence_id(self, row: GroupedRow) -> Optional[dict[str, str]]:
         """Get external sequence ID from sequence file name"""
-        try:
-            for filename, _path in self.filename_map.items():
-                if (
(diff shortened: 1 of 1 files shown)


illusional (changes requested):
IMO this loop is structured in a weird way, I'd suggest restructuring to avoid the need of catching an unbound local variable and throw the error directly, eg:

```python
for filename, _path in self.filename_map.items():
    if (
        fastq_file_name_to_sample_id(filename)
        == row[0][Columns.MANIFEST_FLUID_X]
    ):
        split_filename = filename.split('_')[0:4]
        return {'kccg_id': external_sequence_id}
raise ValueError(
    f'Ingestion of {row[0][Columns.EXTERNAL_ID]} / {row[0][Columns.MANIFEST_FLUID_X]} failed. '
    f'Does the corresponding FASTQ exist in the search locations?'
)
```

vivbak:
That's a lot nicer, thanks @illusional!

illusional (approved):

PR #458, opened 2023-04-12
Sequence date added field

EddieLF:
So these changes are kinda simple but at the same time I'm finding it difficult to test due to how baked into the API this section is with all the awaits/coroutine happenings.

What I think this is doing:
1. Define `datetime_added` as `None` to handle the case when the `skip_checking_gcs_objects` is `True`
2. Asynchronously pass the `file_location` into `get_gcs_blob` to get the blob object
3. Define `datetime_added` as the [blob attribute](https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.blob.Blob#google_cloud_storage_blob_Blob_time_created) `time_created` and add this to the meta dict
4. Blob's your uncle

**Unit tests**
I've changed the parser to initialise with the `skip_checking_gcs_objects=True` argument for `test_single_row` and `test_cram_with_no_reference`. These tests were getting stuck at the `get_gcs_blob` which returned a `ValueError` when the fake/test data was passed to it.

**Possible unit test issue**
Since `test_single_row` ends up calling `create_secondary_file_objects_by_potential_pattern` with `skip_checking_gcs_objects=True`, secondary file objects are created for this test. So, I've modified the test to include these secondary files in the expected output. 
Only thing is, this has broken the `mock_filesize=111` part of the test, so I changed the expected filesize to None. Not sure if this is desirable or something that I should go back and try to fix.

### DIFF
--- a/sample_metadata/parser/cloudhelper.py
+++ b/sample_metadata/parser/cloudhelper.py
@@ -3,6 +3,7 @@ import os
 import logging
 from typing import Iterable, Callable, TypeVar
 from collections import defaultdict
+from datetime import datetime
 
 from cloudpathlib import AnyPath, GSPath
 from google.cloud import storage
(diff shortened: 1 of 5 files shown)


illusional (approved):
Looks good, you can skip the bumpversion as I've added one in a different PR to go out in the next release.

PR #462, opened 2023-04-14
Update Delete Project Endpoint 

vivbak:
Clears all the data by default, moves deleting the project itself to an optional parameter that is false by default.


db/python/tables/project.py
```diff
@@ -449,8 +454,9 @@ async def delete_project(self, project_id: int, author: str) -> bool:
 DELETE FROM sample WHERE project = :project;
 DELETE FROM participant WHERE project = :project;
 DELETE FROM analysis WHERE project = :project;
-DELETE FROM project WHERE id = :project;
             """
+            if delete_project:
+                _query += 'DELETE FROM project WHERE id = :project;'
```
illusional, line 459:
```suggestion
                _query += 'DELETE FROM project WHERE id = :project;\n'
```

api/routes/project.py
```diff
@@ -78,19 +78,23 @@ async def update_project(
     )
 
 
-@router.delete('/{project}', operation_id='deleteProject')
-async def delete_project(
+@router.delete('/{project}', operation_id='deleteProjectData')
```
illusional, line 81: Just noting this is technically a breaking change, but I don't think this method is used so should be fine.

illusional (approved):

api/routes/project.py
vivbak: Definitely, I didn't find another reference when I looked :)

PR #471, opened 2023-04-24
Audit main upload buckets

EddieLF:
Apologies for this somewhat overdue PR reaching you the evening before a public holiday!

Here's a script I've drafted up to support an audit of the main-upload buckets. It takes some parts from the [cleanup_fastqs](https://github.com/populationgenomics/sample-metadata/blob/main/scripts/clean_up_fastqs.py) and the [find_sequence_files_to_delete](https://github.com/populationgenomics/sample-metadata/blob/main/scripts/find_sequence_files_to_delete.py) scripts. This script is different because it focuses on a single project at a time, and does some comparisons between files in the bucket and files in Metamist to look for files which may have been moved since they were written to Metamist.

Basically, the script is doing a few things:
1. Get all participants in a project, their samples, and the sequences for their samples
2. Get all the sequence file paths that exist in the project's main-upload bucket
3. For samples without a completed CRAM - log these
           For samples with a completed CRAM - add their sequence files to the delete list
4. Any files found in the bucket but NOT found in metamist are checked to see if they ARE a sequence file from metamist, but in a different location. We match the file names and sizes for this check and consider these as "moved" sequence files.
5. Report any uningested files with the logger (TODO: report these in an output file).
           Add any "moved" files to the sequence files to delete list if the associated sample has a completed CRAM.
6. Execute the delete on the full list (if the dry-run flag is absent).

Currently, the scope of this is limited to fastqs, but I think it can be expanded to capture other data formats. I'd also like to expand it to have further control over what sequence types and read types to look at.

For a first pass this process should work to delete a bulk amount of the redundant data in the upload buckets. Subsequent passes will go through the buckets and find other files to delete, and can be run at regular intervals to maintain low levels of redundant data. 

Sorry this script is a bit of a mess in the main block - there are a lot of local variables, some of which could probably be removed. A lot of these exist as mappings between various Metamist IDs and other IDs and fields, useful for reporting. If anything isn't clear I'm happy to explain or refactor as necessary.


scripts/audit_main_upload_buckets.py
```diff
@@ -0,0 +1,620 @@
...
+from sample_metadata.model.analysis_type import AnalysisType
+from sample_metadata.model.analysis_status import AnalysisStatus
+
+
+FASTQ_EXTENSIONS = ('.fq.gz', '.fastq.gz', '.fq', '.fastq')
+BAM_EXTENSIONS = ('.bam',)
+CRAM_EXTENSIONS = ('.cram',)
+GVCF_EXTENSIONS = ('.g.vcf.gz',)
+VCF_EXTENSIONS = ('.vcf', '.vcf.gz')
+ALL_EXTENSIONS = (
```
illusional, line 44: You can add tuples, so would recommend just adding all the *_EXTENSIONS together

scripts/audit_main_upload_buckets.py
```diff
@@ -0,0 +1,620 @@
...
+            files_in_bucket.extend(get_paths_for_subdir(bucket, subdir, file_types))
+
+    return files_in_bucket
+
+
+def find_sequence_files_in_bucket(bucket_name: str) -> list[str]:
+    """Gets all the gs paths to fastq files in the projects main-upload bucket"""
+    sequence_paths = []
+    for blob in CLIENT.list_blobs(bucket_name, prefix=''):
+        if blob.name.endswith(FASTQ_EXTENSIONS):
```
illusional, line 157: We get non-fastq sequencing files from collaborators, it might be worth creating a GENOMIC_READS_EXTENSIONS (with .bam, .cram, .fastq... extensions)

scripts/audit_main_upload_buckets.py
```diff
@@ -0,0 +1,620 @@
...
+            sequence_paths.append(f'gs://{bucket_name}/{blob.name}')
+        continue
+
+    return sequence_paths
+
+
+def get_filename_from_path(path: str) -> str:
+    """Extracts filename from file path. e.g. 'subdir/file.ext' -> 'file.ext'"""
+
+    return path.rsplit('/', 1)[1]
```
illusional, line 167: Use `os.path.basename` here

scripts/audit_main_upload_buckets.py
```diff
@@ -0,0 +1,620 @@
...
+
+
+def get_filesize_from_path(bucket_name: str, path: str) -> int:
+    """Extracts filesize from gs path"""
+    bucket = CLIENT.get_bucket(bucket_name)
+    blob_path = path.split(f'gs://{bucket_name}/')[1]
```
illusional, line 173:
```suggestion
    blob_path = path.removeprefix(f'gs://{bucket_name}/')
```

illusional (commented):
I only had a quick skim through and had a couple of thoughts:

1. Could you write some tests to cover this? There's a lot of functionality here, and want to make sure we don't break it.
2. Would you consider rewriting the metamist calls with `graphql`? Reducing the calls isn't very important, but it allows you to more clearly express the query of information you want, get it all at once and not have to do so much testing later.
3. Does this make the cleanup_fastq_files / find_sequence_file_to_delete scripts obsolete, and if not, could it? I see most of the code blocks here, would be good to unify them all into one utility module :)

EddieLF:
Hey @illusional, I've written a few unit tests for the new functions in the `AuditHelper` class. Since they use other nested functions (mostly just the `get_gcs_blobs` one) I've sorta mocked the call for this with a new definition just for the testing at the top of the file. Not sure if this is ideal, maybe it should be patched into the input arguments and assigned mock return values instead. Let me know if there's anything you think should change in this unit test, I haven't developed one for the generic auditor yet.

Otherwise, the audit module seems to be functioning pretty well overall on the test upload buckets.

audit/generic_auditor.py
```diff
@@ -0,0 +1,575 @@
...
+            for sample_map in samples_all
+            for sample_id, sample_external_id in sample_map.items()
+            if sample_id not in EXCLUDED_SAMPLES
+        }
+
+        return sample_internal_external_id_map
+
+    def get_sequence_mapping(
+        self, all_sequences: list[dict]
+    ) -> tuple[dict[int, str], defaultdict[int, list]]:
```
illusional, line 187: (This is slightly mis-annotated, you're adding tuples, not a list)
One thing you can do for clarity, is define some placeholder for IDs, eg:

```python
SequenceId = int
SampleId = str
SampleExternalId = str
# ... etc

def get_sequence_mapping(all_sequences: list[dict]) -> tuple[dict[SequenceId, SampleId], dict[SequenceId, tuple[str, int]]):

**Review by illusional — CHANGES_REQUESTED** [review]
Hey @EddieLF, functionally the code is looking good!
In cases in your function where you don't use `self`, I'd strongly encourage making the methods static.

The tests need a bit of work, but basically you want to test specific functionality in other defined functions. You basically don't want _any_ logic inside the tests except setup, call function, and check.

For example, there is a lot of functionality in `check_for_uningested_or_moved_sequences`, so you could write a couple of unittests that provides some values, and checks the expected output values for different cases. Importantly, a lot of these functions don't have side-effects, or don't pull data inside the method. So you don't need to mock the values either.

---

If you want to write integration tests (which I'd definitely encourage), you'll have to mock any external calls, eg: `get_participant_data_for_project`, or `get_analysis_cram_paths_for_project_samples` where you call out to metamist. You've seen what mocks look like but happy to provide some guidance if you get stuck.

**Comment by EddieLF** [comment]
@illusional I'm pretty happy with the state this is in, I've run the auditor across most RD datasets locally and collected a tonne of files to delete. I could run the deletes locally with the Data Manager permissions but it's better if everything is logged in Hail. 

One thing I'm not sure about is the imports in the audit module/audit test, currently they are just doing 
`from audit.x import X`, 
but once the module is deployed, will it have to be 
`from sample_metadata.audit.x import X`?

Let me know if you think there are any further i's to dot or t's to cross when you get a chance, then hopefully we can merge. Cheers

File: audit/audit_upload_bucket.py [hunk 6]
```diff
@@ -0,0 +1,274 @@
...
+        sequence_files_to_ingest: list[tuple[str, str, str, int, str]],
+        incomplete_samples: list[tuple[str, str]],
+    ):
+        """Write the sequences to delete and ingest to csv files and upload them to the bucket"""
+        # Writing the output to files with the file type, sequence type, and date specified
+        today = datetime.today().strftime('%Y-%m-%d')
+
+        bucket_name = f'cpg-{self.dataset}-main-upload'
+        if access_level == 'test':
+            bucket_name = f'cpg-{self.dataset}-upload'
```
MattWellie, line 102:
```suggestion
            bucket_name = f'cpg-{self.dataset}-test-upload'
```

MattWellie, line 102: If this runs in Analysis-Runner you can probably swap this all out for 
`get_config()['storage']['default']['upload']`

audit/delete_sequence_files_from_audit.py
```diff
@@ -0,0 +1,97 @@
...
+    access_level = config['workflow']['access_level']
+
+    subdir = os.path.join('audit_results', date)
+
+    if access_level == 'standard':
+        raise ValueError(
+            'Standard cannot be used as the access level. Full or test is required.'
+        )
+
+    if access_level == 'full':
```
MattWellie, line 62: another one for `get_config()['storage']['default']['upload']` - this will automagically create the correct bucket path depending on the access level

MattWellie (commented):
still chugging through

audit/generic_auditor.py
```diff
@@ -0,0 +1,652 @@
...
+        sample_internal_external_id_map = {
+            sample_id: sample_external_id
+            for sample_map in samples_all
+            for sample_id, sample_external_id in sample_map.items()
+            if sample_id not in EXCLUDED_SAMPLES
+        }
+
+        return sample_internal_external_id_map
+
+    def get_sequence_mapping(
```
MattWellie, line 189: there are a lot of `continue`s in this block; are there known metamist results we expect to trigger all these failure conditions, or is this just super cautious?

MattWellie: If you want to soft-succeed (not sure that's a phrase) this is fine, if you want to hard fail when you encounter wacky data it might be worth appending all failure types to a list, then at the end of the method you can decide whether to return the data (no failures) or raise an exception (and print the failures)

audit/generic_auditor.py
```diff
@@ -0,0 +1,652 @@
...
+        """Return latest sequence ids for internal sample ids"""
+        all_sequences = [
+            {
+                sample.get('id'): sample.get('sequences')
+                for sample in participant.get('samples')
+            }
+            for participant in participants
+        ]
+
+        return self.get_sequence_mapping(all_sequences)
```
MattWellie, line 268: Can these methods be merged or separated? You're making yourself a unit testing headache if you tie two methods together

audit/generic_auditor.py
EddieLF: Yeah this function borrows a lot from the `find_sequence_files_to_delete` script [here](https://github.com/populationgenomics/sample-metadata/blob/72b5ea46bf73c952119bfb63990c9f55b0f7f228/scripts/find_sequence_files_to_delete.py#L69). I believe there are some older sequence entries with odd data formatting, its overly cautious now but can probably be toned back in future as new sequence/assay entries follow a standardised format

EddieLF:
@MattWellie thanks for your review. I've finished cleaning up the files with your suggested changes, including using GraphQL rather than the analysis API. This meant I had to refactor the unit tests which took a hot minute, but I believe everything is working now. Let me know if there's anything further I should test or alter before we can merge this bad boy.

audit/generic_auditor.py
```diff
@@ -4,23 +4,19 @@
 from typing import Any
 
 from audit.audithelper import AuditHelper
```
MattWellie, line 6:
```suggestion
from audithelper import AuditHelper
```

MattWellie (changes requested):
All looks very sexy. Without installing as a package I reckon this chained import will cause problems, but other than that it looks good

MattWellie (approved):

EddieLF:
Hey @illusional, just resurrecting your presence in this PR to get your thoughts on the module install and importing. I've added the audit module to the `sample_metadata` folder so that it gets picked up by the package installer, in a similar way to how the `parser` module and tests work.

The unit tests are not happy about this but I think it will be fine once the module is part of the installation. Could you have a quick look and let me know if this will be fine? Cheers

MattWellie:
@EddieLF you will have to change your mocking calls to 
`@patch('sample_metadata.audit.generic_auditor.query')`

Other than that and an irritating 
`can't import sm_url during "from sample_metadata.configuration import sm_url`" from graphql.__init__ the generic auditor tests run locally for me

illusional (commented):
A few small comments, but the import looks fine to me, strange that it's failing in test.
@MattWellie the `sm_url` error means you need to regenerate the API (`python regenerate_api.py`)

illusional (approved):

illusional (approved):

PR #472, opened 2023-04-26
Toggle dark mode

danielreti:
Added a slider on the navbar + cleaned up a bit


web/src/App.tsx
```diff
@@ -15,21 +21,50 @@ const App: React.FunctionComponent = () => (
                     <Link className="metamist-img" to="/">
                         <MuckTheDuck height={28} style={{ marginRight: '5px' }} />
                     </Link>
+
                     <Link className="metamist" to="/">
```
illusional, line 25: Should this be refactored into a Header component as it's now getting bigger and more complicated.

web/src/shared/components/DarkModeTriButton/DarkModeTriButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { ThemeContext } from '../ThemeProvider'
+
+const DarkModeTriButton: React.FunctionComponent = () => {
+    const { toggleTheme } = React.useContext(ThemeContext)
+    const [selection, updateSelection] = React.useState(
+        localStorage.getItem('theme') || 'system-mode'
+    )
+    return (
+        <div className="switch-toggle">
+            <input
```
illusional, line 16: There's a bit of code duplication here, is it worth refactoring this into a loop

web/src/shared/components/DarkModeTriButton/DarkModeTriButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+                id="on"
+                name="state-d"
+                type="radio"
+                onChange={() => {
+                    toggleTheme('light-mode')
```
illusional, line 21: Might be reasonable to make these constants that can be referred to throughout the code to avoid mistypes

web/src/shared/components/ThemeProvider.jsx
```diff
@@ -0,0 +1,45 @@
...
+            setTheme('light-mode')
+        } else if (type === 'dark-mode') {
+            localStorage.setItem('theme', 'dark-mode')
+            document.documentElement.setAttribute('data-theme', 'dark-mode')
+            setTheme('dark-mode')
+        } else {
+            localStorage.setItem('theme', 'system-mode')
+            document.documentElement.setAttribute('data-theme', preferredMode)
+            setTheme(preferredMode)
+        }
```
illusional, line 27: Nitpicky, but IMO this could be cleaner code and reduce some duplication.

```typescript
const toggleTheme = (type) => {
    let newTheme = type
    switch (type) {
        case .systemMode: {
             newTheme = preferredMode
             break;
        }
    }

    localStorage.setItem('theme', type)
    document...setAttribute('data-theme', theme);
    setTheme(theme)
}
```

web/src/shared/components/ThemeProvider.jsx
```diff
@@ -0,0 +1,45 @@
...
+    React.useEffect(() => {
+        document.documentElement.setAttribute('data-theme', savedTheme)
+        setTheme(savedTheme)
+    }, [savedTheme])
+
+    return (
+        <div>
+            <ThemeContext.Provider value={{ theme, toggleTheme }}>
+                {/* eslint-disable react/prop-types */}
+                {props.children}
```
illusional, line 39: I suspect a later version of react removed implicit children from FunctionalComponent, could be worth making it explicit in the props

web/src/shared/components/ToggleDarkMode.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
```
illusional, line 1: Should these components go in a Header folder?

illusional (approved):
A few nitpicky changes + some slight reorganisation. I'll approve to not be blocking because I don't think I need to look again, but would want those addressed first.

web/src/App.tsx
danielreti: Added NavBar component

web/src/shared/components/DarkModeTriButton/DarkModeTriButton.tsx
danielreti: Donezo

web/src/shared/components/DarkModeTriButton/DarkModeTriButton.tsx
danielreti: Done as part of the loop

web/src/shared/components/ThemeProvider.jsx
danielreti: Makes sense - I've made this change

web/src/shared/components/ToggleDarkMode.tsx
danielreti: Donezo

PR #479, opened 2023-05-05
Fix UI sequence groups

danielreti:
Making UI changes to Project Summary Grid, Family View, Sample View and Search to represent the addition of Sequence Groups


danielreti:
New Project Grid
Screenshot 2023-05-05 at 4 58 12 pm

danielreti:
New Sample View
image

danielreti:
New Family View
Screenshot 2023-05-05 at 4 59 06 pm

web/src/shared/utilities/deepPartial.ts
```diff
@@ -0,0 +1,5 @@
+export type DeepPartial<T> = T extends object
```
illusional, line 1: Could you add a comment describing how this would be used?

web/src/shared/components/SeqPanel.tsx
```diff
@@ -2,35 +2,38 @@ import * as React from 'react'
...
 import { Accordion } from 'semantic-ui-react'
 import ScienceRoundedIcon from '@mui/icons-material/ScienceRounded'
-import { GraphQlSampleSequencing } from '../../__generated__/graphql'
-import SeqInfo from './SeqInfo'
+import { GraphQlSequencingGroup } from '../../__generated__/graphql'
 import iconStyle from '../iconStyle'
+import { DeepPartial } from '../utilities/deepPartial'
+import SeqGroupInfo from './SeqGroupInfo'
 
 const SeqPanel: React.FunctionComponent<{
```
illusional, line 10: Update name

web/src/shared/components/SeqGroupInfo.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react'
+import { Table as SUITable } from 'semantic-ui-react'
+
+import _ from 'lodash'
+import { GraphQlSequencingGroup, GraphQlAssay } from '../../__generated__/graphql'
+import Table from './Table'
+import { DeepPartial } from '../utilities/deepPartial'
+import AssayInfo from './AssayInfo'
+
+const excludedSequenceGroupFields = ['id', '__typename']
```
illusional, line 10: This is gonna be petty, but I'm trying to be precise and consistent because we had a few issues earlier. 
It's "Sequenc*ing*Group", and the word "Sequence" should never appear in metamist. Maybe I should make a pre-commit hook haha

web/src/shared/components/Header/Search.tsx
```diff
@@ -169,6 +175,8 @@ const Searchbar: React.FunctionComponent = () => {
                 // alert("Family page not implemented yet");
                 navigate(`/family/${id}`)
                 break
+            case 'sequencing-group':
+                navigate(`/sample/${id}/${seqID}`)
```
illusional, line 179: I'm a little nervous routes aren't formatted by some singular helper.

web/src/shared/components/Header/Search.tsx
```diff
@@ -80,7 +81,7 @@ const resultRenderer = ({ ...props }) => {
         available = `No access to this ${props.type}`
         colour = 'gray'
     }
-    if (props.type === 'participant' || props.type === 'family') {
+    if (props.type === 'participant') {
```
illusional, line 84: Can we type this component, and formalise the type param?

web/src/pages/family/FamilyView.tsx
```diff
@@ -17,7 +17,7 @@ import iconStyle from '../../shared/iconStyle'
 import SeqPanel from '../../shared/components/SeqPanel'
 import SampleInfo from '../../shared/components/SampleInfo'
 
-const sampleFieldsToDisplay = ['active', 'type' /*'participantId'*/]
+const sampleFieldsToDisplay = ['active', 'type']
 
 // TODO: Fix this query
```
illusional, line 22: You can remove this TODO now :)

db/python/tables/sequencing_group.py
```diff
@@ -86,6 +91,34 @@ async def get_assay_ids_by_sequencing_group_ids(
...
+            _query,
+            {
+                # 'project_ids': project_ids,
+                # 'search_pattern': self.escape_like_term(query) + '%',
+                'limit': limit,
+            },
+        )
+        print('test')
+        print(rows)
+        print('test2')
```
illusional, line 120: Debugging statements

db/python/tables/sequencing_group.py
```diff
@@ -86,6 +91,34 @@ async def get_assay_ids_by_sequencing_group_ids(
...
+    async def search(
+        self, query, project_ids: list[ProjectId], limit: int = 5
+    ) -> list[tuple[ProjectId, int, str]]:
+        
+        """
+        Search by some term, return [ProjectId, SampleID, SequencingGroupID]
+        """
+        # SELECT s.project, sg.sample_id, sg.id, s.participant_id
+        _query = """
+        SELECT *
```
illusional, line 104: Can you only select the fields you use below?

db/python/tables/sequencing_group.py
```diff
@@ -86,6 +91,34 @@ async def get_assay_ids_by_sequencing_group_ids(
...
+        FROM sequencing_group sg
+        INNER JOIN sample s ON s.id = sg.sample_id
+        LIMIT :limit
```
illusional, line 107: The search_pattern is currently not used.

db/python/layers/web.py
```diff
@@ -459,7 +459,7 @@ async def get_project_summary(
             'sequencing_technology',
             'sequencing_type',
         }
-        ignore_sg_meta_keys: set[str] = set()
+        ignore_sg_meta_keys: set[str] = {'facility'}
```
illusional, line 462: Any particular reason for this?

db/python/layers/search.py
```diff
@@ -103,27 +157,62 @@ async def search(self, query: str, project_ids: list[int]) -> List[SearchRespons
             )
 
             return [response] if response else []
+    
+        if cpg_sg_id := self.try_get_sg_id_from_query(query):
```
illusional, line 161: Yeah perfect

illusional (commented):
I know you weren't quite ready / asking for review, but looks good so far! A few debugging things to clean-up, and if you can add some sequencing-group tests that would be great 🙏

web/src/shared/components/SeqPanel.tsx
danielreti: I updated SeqGroupInfo to SequencingGroupInfo - did you want the deepPartial name changed to something too?

web/src/shared/components/SeqGroupInfo.tsx
danielreti: Fixed - I will double check if I used SequenceGroup anywhere else

web/src/shared/components/Header/Search.tsx
danielreti: I think this is pretty clear but can turn into some utility if you prefer - pass a type and id(s) and it'll navigate appropriately?

web/src/shared/components/Header/Search.tsx
danielreti: Added the enum for the type param

db/python/layers/web.py
danielreti: Facility appears on both assays and sequencing groups, figured it made sense to omit from SGs and keep on assays in case any assays in the same SG came from different facilities?

web/src/shared/components/Header/Search.tsx
illusional: I guess it's fine for now, just don't like that route logic can be scattered across the application. I'd prefer a utility that you pass the type and ID and returns the constructed route (where this class performs the navigate, or alternatively it's used in a Link).

db/python/layers/web.py
illusional: I think this is a byproduct of the test data, and not of real data (I could be mistaken though)

web/src/shared/utilities/deepPartial.ts
illusional: This what resolved, but I'm still not sure what DeepPartial is

db/python/layers/search.py
```diff
@@ -103,27 +157,62 @@ async def search(self, query: str, project_ids: list[int]) -> List[SearchRespons
...
         )
+        # seq_group_participant_ids = [sg[3] for sg in sg_rows]
 
         sample_participant_eids, participant_family_eids = await asyncio.gather(
             ptable.get_external_ids_by_participant(sample_participant_ids),
             ftable.get_family_external_ids_by_participant_ids(all_participant_ids),
+            # ptable.get_external_ids_by_participant(seq_group_participant_ids),
         )
 
+        # sequencing_groups = [
```
illusional, line 195: Is this still supposed to be commented out?

db/python/layers/search.py
```diff
@@ -103,27 +157,62 @@ async def search(self, query: str, project_ids: list[int]) -> List[SearchRespons
             )
 
             return [response] if response else []
+    
+        if cpg_sg_id := self.try_get_sg_id_from_query(query):
+            # just get the sg
+            response = await self._get_search_result_for_sequencing_group(
+                cpg_sg_id, project_ids=project_ids
+            )
+            print(response)
```
illusional, line 166: Remove print statement

illusional (changes requested):
Can you add tests to search, and looks like sequencing-group searching is not implemented yet.

db/python/layers/web.py
danielreti: Ah ok - I'll add it back in then (we can always customise this later)

web/src/shared/components/Header/Search.tsx
danielreti: I'll add this to the to-do list

db/python/layers/search.py
danielreti: @illusional  I don't think the test data has SG external IDs, probably why I had such trouble writing the query. Should we merge the exact match SG internal ID search at least and add in the external ID string match later? I don't know how much use case there is for string matching on IDs - you'd probably need to know what you're searching for before you search it, and if you're just exploring the data then the project grid is more useful for that?

illusional: Yeah that seems reasonable, can you add a TODO in the spot where we should add the SG external ID searching

db/python/layers/web.py
```diff
@@ -509,12 +510,14 @@ async def get_project_summary(
...
+        assay_keys = [('type', 'type')] + sorted(
             [('meta.' + k, k) for k in assay_meta_keys]
         )
         sequencing_group_keys = [
             ('id', 'Sequencing Group ID'),
-            ('created_date', 'Created date'),
+            ('platform', 'Platform'),
+            ('technology', 'Technology'),
+            ('type', 'Type'),
         ] + sorted([('meta.' + k, k) for k in sg_meta_keys])
```
illusional, line 521: Changes that might affect the test

danielreti:
I've merged in the changes and it passed the local unit tests but something has happened to Liquibase(?) and the unittests are failing @illusional

illusional:
Hmm, that's annoying. The script polls for the latest version of liquibase, but this version has been removed: https://github.com/liquibase/liquibase/releases/tag/v4.22.0
So we either physically lock it to an earlier version, or wait for the new release that fixes it

danielreti:
Whichever solution you prefer! Also can merge this whenever since the other branches will have the same issue

PR #481, opened 2023-05-12
Add Metamist ETL infrastructure as code plugin

illusional:
Imported into cpg-infrastructure as a plugin.
Stacks on: https://github.com/populationgenomics/cpg-infrastructure/pull/130

This is _directly_ addressing the _Metamist has API endpoints for external integrations_ OKR.


violetbrina (approved):
I'm assuming the schema is empty because either we/they haven't finalised the schema yet?

PR #483, opened 2023-05-18
function to update family_id in family_participant table

EddieLF:
A draft for a simple function to update family ids for participants in the `family_participant` table. This relates to the issue described [here](https://github.com/populationgenomics/seqr-private/issues/58#issuecomment-1487766524), where there are two participants who are currently under the wrong family.


illusional (changes requested):
The participant may be in multiple families, so this would need more conditions unfortunately

EddieLF:
I've changed the `family_participant` table function to now delete based on a participant ID and family ID, so only one row will be deleted per execution.

I also added some functions in the `ParticipantLayer `which can call the delete and create row functions from the `family_participant` table class. Finally I dropped in some API endpoints to use these new functions. 

To update a participant's family, the idea would be:
1. Get the `participant_family` row for that participant/family, save the fields like paternal_id, maternal_id, affected
2. Delete the participant from the family using the new function
3. Create a new row in `participant_family` with the correct family ID, populating the row with the fields saved in step 1.

Not exactly sure if this is the right approach, or if there should only be one endpoint (like "update participant family") which executes steps 1-3 above^.

EddieLF:
Latest commits add a new endpoint to the participant API, `update_participant_family`, as well as some unit testing.

This function takes a participant ID, current participant family ID, and a new participant family ID.
1. Get the current participant/family data (paternal/maternal ID, sex, affected status)
2. Remove the participant from the current family - delete the row in `family_participant`
3. Add the participant to the new family - add row to `family_participant` with the data saved in step 1

Understand this will need a bumpversion commit too, I'll add that if this works.

api/routes/participant.py
```diff
@@ -248,3 +248,39 @@ async def get_participants(
...
+    )
+
+    # Use saved values to maintain the fields in the new row
+    return await player.add_participant_to_family(
+        family_id=new_family_id,
+        participant_id=participant_id,
+        paternal_id=fp_row['paternal_id'],
+        maternal_id=fp_row['maternal_id'],
+        affected=fp_row['affected'],
+    )
```
illusional, line 286: Can you move all this business logic to the layer. This allows it to be unit-tested.
It would also be good to wrap this around a database transaction to ensure a roll-back if an error occurs (like the new family ID doesn't exist).

db/python/tables/family_participant.py
```diff
@@ -159,6 +159,50 @@ async def get_rows(
...
+        }
+
+        wheres = ['f.id = :family_id', 'p.id = :participant_id']
+        conditions = ' AND '.join(wheres)
+
+        _query = f"""
+            SELECT {keys_str} FROM family_participant fp
+            INNER JOIN family f ON f.id = fp.family_id
+            INNER JOIN participant p on fp.participant_id = p.id
+            WHERE {conditions}"""
```
illusional, line 190: Small nitpick, but if you're not dynamically changing the conditions and the keys are static to this method, you can just include them as part of the query and remove the fstring

illusional (changes requested):
This change is looking good. One minor suggestion, but as it stands this does not have any authorization / project checking as part of the get or mutation. 

Up front, you should get the "projects" for both the participant, and families. And check that the caller as write access to both of those.

test/test_update_participant_family.py
```diff
@@ -0,0 +1,70 @@
...
+    async def test_get_and_update_family_participant_data(self):
+        """
+        Tests getting and updating a participants family data
+        """
+        pl = ParticipantLayer(self.connection)
+
+        p = await pl.get_participants(project=1)
+        pid = p[0].id
+
+        fp_row = await pl.get_family_participant_data(family_id=1, participant_id=pid)
```
illusional, line 36: This value can change between tests. You can make it more secure by storing the family ID returned by create_family (or making sure it returns the ID of he created family), and using that to query against.

test/test_update_participant_family.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+
+    @run_as_sync
+    async def test_get_and_update_family_participant_data(self):
+        """
+        Tests getting and updating a participants family data
+        """
+        pl = ParticipantLayer(self.connection)
+
+        p = await pl.get_participants(project=1)
```
illusional, line 33:
```suggestion
        p = await pl.get_participants(project=self.project_id)
```

test/test_update_participant_family.py
```diff
@@ -0,0 +1,70 @@
...
+        self.assertDictEqual(fp_row, expected_fp_row)
+
+        await pl.remove_participant_from_family(family_id=1, participant_id=pid)
+
+        await pl.add_participant_to_family(
+            family_id=2,
+            participant_id=pid,
+            paternal_id=fp_row['paternal_id'],
+            maternal_id=fp_row['maternal_id'],
+            affected=fp_row['affected'],
```
illusional, line 55: FWIW, I think I've implemented this poorly. The maternal / paternal references should be on the participant.
A problem for a future dev 🤷

test/test_update_participant_family.py
EddieLF: I've updated the participant and family IDs generated in setup to be attributes of the test instance so they can be referenced in the test functions, that should keep them consistent

api/routes/participant.py
EddieLF: Added a new layer function specifically with this logic + a transaction wrapper, now the endpoint calls that layer function only

EddieLF:
@illusional changes implemented, I added the project access check for the family/participant IDs into the `get_family_participant_data`, `add_participant_to_family`, and `remove_participant_from_family` methods.

Is it redundant to do the check in the 3 helper methods when I could just do it once in the new `update_participant_family` method?

db/python/layers/participant.py
```diff
@@ -665,6 +665,101 @@ async def update_many_participant_external_ids(
...
+        """Gets the family_participant row for a specific participant"""
+        if check_project_ids:
+            pprojects = await self.pttable.get_project_ids_for_participant_ids(
+                [
+                    participant_id,
+                ]
+            )
+            ftable = FamilyTable(self.connection)
+            fprojects = await ftable.get_projects_by_family_ids(
+                (
```
illusional, line 680: The double parenthesis look strange to me

db/python/layers/participant.py
```diff
@@ -665,6 +665,101 @@ async def update_many_participant_external_ids(
...
+            )
+            ftable = FamilyTable(self.connection)
+            fprojects = await ftable.get_projects_by_family_ids(
+                [
+                    family_id,
+                ]
+            )
+            await self.ptable.check_access_to_project_ids(
+                self.connection.author,
+                list(set().union(pprojects, fprojects)),
```
illusional, line 714:
```suggestion
                list(pprojects | fprojects),
```

db/python/layers/participant.py
```diff
@@ -665,6 +665,101 @@ async def update_many_participant_external_ids(
...
```
illusional, line 748:
```suggestion
                list(pprojects | fprojects),
```

illusional (approved):
I think it would be cleaner to check it once for the update, and skip the check in the helper methods. You might consider a helper for checking the project IDs for those two entity types ;) 

If you're looking to be exhaustive (definitely optional), you can add a test for a family ID that doesn't exist (choose like -99 or something), and check the rollback works if the transaction fails.

PR #484, opened 2023-05-22
Bug fix

vivbak:
`conditionals` can sometimes have empty values which results in query strings that look like `AND AND`

### DIFF
--- a/db/python/utils.py
+++ b/db/python/utils.py
@@ -262,7 +262,7 @@ class GenericFilterModel:
         if not conditionals:
             return 'True', {}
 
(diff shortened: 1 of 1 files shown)


db/python/utils.py
```diff
@@ -262,7 +262,7 @@ def to_sql(
         if not conditionals:
             return 'True', {}
 
-        return ' AND '.join(conditionals), values
+        return ' AND '.join(filter(None, conditionals)), values
```
illusional, line 265: Wow, I did not know None called be called like this

illusional (approved):

PR #491, opened 2023-06-07
Update packages

danielreti:
Address these 7 alerts - https://github.com/populationgenomics/sample-metadata/security/dependabot?q=is%3Aopen


illusional (approved):
I thought you'd bump the version in `package.json` and let it flow through to `package-lock.json`, is that not the case?

danielreti:
I thought so too but I did `npm audit fix` rather than changing `package.json` directly, and it only updated `package-lock.json`

danielreti:
I guess that's just how it works - https://stackoverflow.com/a/56683169

PR #498, opened 2023-06-22
Fix Mismatched Samples count

### LINKED ISSUE #489: Mismatch between number of samples in Metamist UI vs API
Context - https://centrepopgen.slack.com/archives/C020PDNH3N0/p1686024754071179

danielreti:
As per https://centrepopgen.slack.com/archives/C020PDNH3N0/p1686024754071179
This change fixes the query that returns all samples as `active: true` and also hides samples from the project grid with `active: false`
Fixes #489


db/python/layers/web.py
```diff
@@ -168,7 +168,7 @@ def _project_summary_sample_query(self, grid_filter: List[SearchItem]):
             wheres.append(q)
             values[key] = self.escape_like_term(value) + '%'
         if wheres:
-            where_str = 'WHERE ' + ' AND '.join(wheres)
+            where_str = 'WHERE ' + ' AND '.join(wheres) + ' AND s.active'
```
illusional, line 171: Can you just add it to `wheres` on initialization? We should always filter

db/python/tables/sample.py
```diff
@@ -497,6 +497,9 @@ async def get_samples_by(
         sample_rows = await self.connection.fetch_all(_query, replacements)
 
         sample_dicts = [dict(s) for s in sample_rows]
+        for sample_dict in sample_dicts:
+            sample_dict['active'] = ord(sample_dict['active'])
```
illusional, line 501: Can you move this to the Sample model?

illusional (changes requested):

illusional (approved):

PR #504, opened 2023-06-30
Add rename_embedded_ids.py script that updates `@RG SM:{id}` values

jmarshall:
Add a fairly simpleminded script that reads CRAM files from a `gs://…` URL, applies `samtools reheader` to update the `@RG SM:{id}` headers as instructed, and writes the CRAM file and associated index file back to new `gs://…` URLs constructed from the old one by updating the `{id}` similarly.

Fortunately the samtools bug I ran into (samtools/samtools#1866) only affects CRAM v2.1 files, which were superseded by v3.0 in 2015 so we surely don't have any — so this streaming method will work after all. (It was just bad luck that I grabbed one when I was testing this locally!)

It's invoked as
```
Copies CRAM files and associated .crai index files to URLs based on new IDs,
also replacing SM:{id} values on the CRAM files' @RG header lines.
Each command-line argument is

    OLDURL[,OLDID],NEWID

If OLDID is not specified, a pattern matching /CPGdd...dd/ is used instead
to match CPG-style IDs in the URL and @RG SM fields.

This script should usually be run via analysis-runner. It requires an image
that contains samtools.
```

### DIFF
--- a/scripts/rename_embedded_ids.py
+++ b/scripts/rename_embedded_ids.py
@@ -52,9 +52,15 @@ if __name__ == '__main__':
     if len(sys.argv) < 2:
         print(__doc__)
 
+    old_urls = []
     for arg in sys.argv[1:]:
         field = arg.split(',')
         if len(field) >= 3:
             reheader_cram(field[0], field[1], field[2])
         else:
             reheader_cram(field[0], 'CPG[0-9]*', field[1])
+        old_urls.append(field[0])
+        old_urls.append(f'{field[0]}.crai')
+
+    logger.info(f'Check output, then manually remove {len(old_urls)} outdated files:')
+    logger.info(f'gcloud storage rm {" ".join(old_urls)}')
--- a/scripts/rename_embedded_ids.py
+++ b/scripts/rename_embedded_ids.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+"""
+Copies CRAM files and associated .crai index files to URLs based on new IDs,
+also replacing SM:{id} values on the CRAM files' @RG header lines.
+Each command-line argument is
+
+    OLDURL[,OLDID],NEWID
+
+If OLDID is not specified, a pattern matching /CPGdd...dd/ is used instead
+to match CPG-style IDs in the URL and @RG SM fields.
+
+This script should usually be run via analysis-runner. It requires an image
+that contains samtools.
+"""
+
+import logging
+import re
+import subprocess
+import sys
+
+logger = logging.getLogger(__file__)
+logging.basicConfig(format='%(levelname)s (%(name)s %(lineno)s): %(message)s')
+logger.setLevel(logging.INFO)
+
+
+def reheader_cram(old_url: str, old_id: str, new_id: str):
+    """
+    Rewrites @RG headers to include SM:{new_id} and writes the resulting CRAM output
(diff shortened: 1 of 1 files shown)


jmarshall:
I'll soon add another commit adding a similar feature to _create_test_subset.py_.

jmarshall:
The changes to _create_test_subset.py_ are somewhat draft and need testing. The _rename_embedded_ids.py_ UI may need to become a little more general to handle all the scenarios we want it to work in.

cassimons (commented):
LGTM thanks.

vivbak (changes requested):
Looks good, John! 

As we discussed, I think it would be good to call metamist here to make sure that our analysis entries don't go out of sync when this script is run (which could have downstream implications in pipeline runs). 

Also it would be useful to add a comment on manual actions from a data management perspective (i.e. deleting the old cram files after this script is run).

vivbak:
Oh also could you look at the linting and unit tests?

jmarshall:
Turns out the unit test failure was a problem on **dev** that has since been fixed and has resolved itself with a rebase onto current **dev**.

Lint was whining about [something I consider good style](https://twitter.com/jomarnz/status/1678610667044220929), sigh. But I'll probably refactor it in due course to shut the thing up.

scripts/create_test_subset.py
```diff
@@ -840,7 +845,40 @@ def get_random_families(
...
+    elif new_path.endswith('.crai'):
+        logger.info(f'Regenerating {new_path} by indexing {new_base_path}')
+        cmd = f"""
+        gcloud storage cat {new_base_path!r} | samtools index -o - - | gcloud storage cp - {new_path!r}
+        """
+    elif new_path.endswith('.md5'):
+        logger.info(f'Regenerating {new_path} by checksumming {new_base_path} [check the file format]')
+        cmd = f"""
+        gcloud storage cat {new_base_path!r} | md5sum | gcloud storage cp - {new_path!r}
+        """
```
MattWellie, line 871: Can you replace this with a non-streaming
- copy CRAM
- reheader
- md5
- copy all results up
- rm local files

At the moment this implementation involves streaming the CRAM 3 times + 1 upload. The MD5 and `.crai` should be ignored during the copy as their contents are invalid post reheader

jmarshall: That would be somewhat non-trivial given how the script is already organised, and would require local storage. This streaming is not egress, so two extra CRAM file reads from GCS vs complexity is a question of trade offs.

MattWellie:
Can we have the same functionality added for gVCFs too?

cassimons:
Let us know if SW does not have the bandwidth to add support for VCF right now. If so we could merge this as is, then some one can follow the template to add VCFs when they have capacity.

jmarshall:
The _create_test_subset.py_ part of this has been separated off as PR #1035.

scripts/create_test_subset.py
jmarshall: It turns out that trying to run `samtools` commands directly from this script is a non-starter, because the script needs to run on an image that has the metamist API installed and any such images will not have samtools installed on them.

So this is now replaced by PR #1044, which creates a new Hail batch so is more at liberty to copy CRAMs around and do the work locally. Huzzah.

PR #508, opened 2023-07-11
Checking previous unit test failures

jmarshall:
Briefly exploring whether the unit tests failing on #504 was an intermittent issue…


jmarshall:
Confirmed that the unit test failure was just an artefact of when I happened to branch from **dev** and was shortly thereafter fixed by e94a9e757c4d4efff13a68302fee6eb230ac7fef.

PR #516, opened 2023-07-18
ADD: add post request for enum types

violetbrina:
As stated in the title. Post requests didn't exist.


daniaki (commented):
Haven't developed on metamist before - is it easy to add unit tests for the new route simulating a POST request?

violetbrina:
> Haven't developed on metamist before - is it easy to add unit tests for the new route simulating a POST request?

I believe most of the testing done in CI is mocked anyway. There isn't really a way to test db changes without running metamist. So we usually just test the API/processing layers.

PR #518, opened 2023-07-19
Fix for broken metamist deployment

daniaki:
Fix for broken metamist deployment due to failed api generation during github action, plus bumpversion for @violetbrina's new API endpoints (already in `dev`)


vivbak (approved):
Thanks so much @daniaki! 🙏 🙌

.github/workflows/deploy.yaml
```diff
@@ -76,8 +76,8 @@ jobs:
       - name: "build deployable API"
         run: |
           export OPENAPI_COMMAND="java -jar openapi-generator-cli.jar"
-          python regenerate_api.py
-          ls -lGh metamist
+          set -e;
```
jmarshall, line 79: I was a bit surprised about this, as I thought GH Actions set `-e` etc by default — as in fact they do, per [the documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) and [this testing](https://github.com/jmarshall/test/actions/runs/5597121893).

However, we set ` defaults: run: shell: bash -l {0}` above, which overrides this default. (I'm not sure why all this needs a login shell, but it's probably best to leave that as is!) So I'd suggest changing that default to

```
shell: bash -eo pipefail -l {0}
```

which will make for better error handling for all the job steps and mean lines 79–80 can be left as is.

daniaki: Thanks for the suggestion! I also have no idea why a login shell is used, or if it is required but I agree that we should leave that for another time.

regenerate_api.py
```diff
@@ -161,15 +171,17 @@ def generate_api_and_copy(output_type, output_copyer, extra_commands: List[str]
...
+        )
 
     output_copyer(tmpdir)
     shutil.rmtree(tmpdir)
 
 
 def generate_schema_file():
     """
-    Generate schema file and place int he metamist/graphql/ directory
+    Generate schema file and place int the metamist/graphql/ directory
```
jmarshall, line 184:
```suggestion
    Generate schema file and place in the metamist/graphql/ directory
```

Might as well fix the whole typo :smile:

daniaki: Good catch!

jmarshall:
Re the lint failure…

> ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements-server.txt'

Hmmm… turns out that line's been failing for ages, but we only noticed this time because of 5723e2f5a51ca804680787ff57d09843c891767c adding `set -e` to the other workflows… 🙃

jmarshall (approved):
The _regenerate_api.py_ changes look good to me. I wanted to exercise them and watch it fail, so I hacked `check_if_server_is_accessible()` to always fail (see 3412486b1e9396af0652254affbc4833de267878) and [ran the workflow manually](https://github.com/populationgenomics/sample-metadata/actions/runs/5605705330/jobs/10255120871).

This indeed aborted the workflow as intended. So :+1: from me.

PR #519, opened 2023-07-19
Fix test subset script 

vivbak:
Several changes including

- Script now queries analyses (previously this was not implemented)
- Script now upserts SG's (previously only samples and assays)
- Fixes to additional_samples and additional_families inputs, so they are more compatible with each other. 
- Implementation of multiple graphql queries (replacing api endpoints)* Not all endpoints have been replaced. 
- Fixes analyses bug. I think this script was previously responsible for test being overpopulated with analyses entries. We weren't handling upserting correctly here. 

Querying has been significantly simplified in this PR, although unfortunately upserting is more complex. So net neutral.


scripts/create_test_subset.py
```diff
@@ -116,204 +232,319 @@ def main(
...
         family_ids = transfer_families(project, target_project, participant_ids)
-        external_participant_ids = transfer_ped(project, target_project, family_ids)
+        upserted_participant_map = transfer_ped(project, target_project, family_ids)
 
-    external_sample_internal_participant_map = get_map_ipid_esid(
-        project, target_project, external_participant_ids
+    existing_data = query(EXISTING_DATA_QUERY, {'project': target_project})
+
+    samples = original_project_subset_data.get('project').get('samples')
+    transfer_samples_sgs_assays(
```
daniaki, line 295: Is there a way to add a `dry_run` option so that these `transfer_` methods print what they will do, rather than actually doing the thing? For example rather than calling `aapi.create_analaysis` it would log the details of the analysis that will be created.

daniaki (changes requested):

PR #520, opened 2023-07-19
Update GitHub Actions versions used

jmarshall:
While we're hacking in _.github/workflows/*_, we should update these actions that we use. Node.js 12 is scheduled to go away for GH Actions in less than four weeks!

Prevent ["The following action(s) uses node12 which is deprecated" warnings](https://github.blog/changelog/2023-07-17-github-actions-removal-of-node12-from-the-actions-runner/) by updating to the current releases of the Actions used. Use the current releases instead of `@main` too, as we don't want to be affected by bleeding-edge bugs.

Rewrite `::set-output` as a write to $GITHUB_OUTPUT instead; the [::save-state and ::set-output commands are also deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

(See also populationgenomics/production-pipelines#354, which was similar for production-pipelines.)


daniaki:
Tests are breaking for the same reason as the deploy issue. You will need to pull the fix in from dev for this to work.

jmarshall:
Yep, just rebased it now.

PR #523, opened 2023-07-20
Resolve 6 2 0 conflicts

daniaki:
Not sure if this is the correct way to go about resolving a merge conflict between `main` and `dev` since `dev` is push protected. I thought to create a new branch based on `main` and this PR resolves the conflicts in `dev` due to the squash merge into `main` in the previous release. This could be completely wrong, or there could be a much better way to do this, however merge conflicts are not my strong point.


jmarshall:
Okay, I've figured out what's going on here, and there **is** a better way to do it. Stand by…

jmarshall:
All the merge commits are of the form:

```
[bumpversion]
<<<<<<< HEAD
current_version = 6.1.0
||||||| 3c902cd
current_version = 6.0.6
=======
current_version = 6.2.0
>>>>>>> dev
```

i.e. it's comparing the common ancestor (6.0.6) with current **main** (6.1.0) and the release to be merged (6.2.0) and because all three are different it doesn't know whether to choose main's version or the to-be-merged version.

We can resolve this by doing the merge in two phases: first remerge 6.1.0 which will work automatically because it'll be 6.0.6 vs 6.1.0 vs 6.1.0 (this is #524); then once that's on **main** we can revisit #513 — this should no longer have conflicts because now it's 6.1.0 vs 6.1.0 vs 6.2.0.

So I'll close this PR in favour of that approach. 🤞

PR #528, opened 2023-07-31
Fix seqr sync es-index by updating layer with gql

EddieLF:
**Updating the ES-Index component of the seqr sync layer to user GraphQL in favour of the Analysis API calls and filtering.** 

We've been having a few issues lately where the dataset will sync and the ES-Index will purportedly update, but then in the seqr admin panel it becomes clear that the index has not been updated.
[Slack thread 1.1](https://centrepopgen.slack.com/archives/C01R7CKJGHM/p1689885171676539)
[Slack thread 1.2](https://centrepopgen.slack.com/archives/C01R7CKJGHM/p1690425213631899)
[Slack thread 2](https://centrepopgen.slack.com/archives/C01R7CKJGHM/p1690706970757799)

The cause of this is not clear, however manually running a syncing script that uses GQL instead of the API has successfully completed the sync. So I'm not certain this alone will fix the issue, but hopefully a step in the right direction regardless. Will revisit this when we next have new indexes to sync and see if it affects them.

### DIFF
--- a/db/python/layers/seqr.py
+++ b/db/python/layers/seqr.py
@@ -27,11 +27,10 @@ from db.python.layers.base import BaseLayer
 from db.python.layers.family import FamilyLayer
 from db.python.layers.participant import ParticipantLayer
 from db.python.layers.sequencing_group import SequencingGroupLayer
-from db.python.tables.analysis import AnalysisFilter
 from db.python.tables.project import ProjectPermissionsTable
 from db.python.enum_tables import SequencingTypeTable
-from db.python.utils import ProjectId, GenericFilter
-from models.enums import AnalysisStatus
+from db.python.utils import ProjectId
+from metamist.graphql import query_async, gql
 
 # literally the most temporary thing ever, but for complete
 # automation need to have sample inclusion / exclusion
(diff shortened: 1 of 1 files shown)


violetbrina (approved):
Looks good to me. Hopefully this resolves your syncing issue :)

EddieLF:
Hi @violetbrina, thanks for the quick review! We might not have new indexes to sync for a while so I'm not too rushed to get this out.

I can see the Unit tests are failing because of the `from metamist.graphql import gql` line. Is this a concern? I'm confused at why that import fails, but `from api.settings import ...` like on line 17 of this file does not...

violetbrina:
So the import is not the issue, it looks like metamist isn't imported. It's possible that the docker used has imported `sample_metadata` and not `metamist`. That's just a guess at this point. But the true error is this: `ModuleNotFoundError: No module named 'metamist'`

EddieLF:
Right, that sounds like it makes the most sense. Do you think this would be bad to merge into dev / eventually release into main? I fear this will cause unit tests to fail across the board if it gets into dev. Again, no big rush on this so we can have a brainstorm sometime soon and leave this PR open for now.

PR #533, opened 2023-08-17
Project Insights endpoints and dashboards

EddieLF:
This is a WIP draft PR for implementing a "Projects Summary" web endpoint. The goal is to provide users with summary stats for one or more projects they are a member of. 

So far, the stats captured here are the number of families, participants, samples, sequencing groups, and completed CRAMs, grouped by dataset AND sequencing type. 

**TODO**: 
So, in the web layer I've put each of the relevant queries into a function, but this is basically just clutter right now, the queries could exist as variables and there would be no difference.

Right now, the code iterates through the input projects and sequencing types and evaluate each of the 5 queries for all combinations of project and sequencing type, adding the result to the response array on each loop. 
This seems suuper inefficient, and I think the more obvious way to do this would be to evaluate all 5 queries for all input projects & sequencing types, _and then_ filter the results of this complete query down into distinct `ProjectsSummaryInternal` objects. 

So, could we, upon calling the `get_projects_summary(projects, sequencing_types)` function, evaluate the 5 queries for all projects and seq types (with a `fetch_all`), then subset the results of this by dataset and seq type once we get to the main loop? This way the 5 queries would only be evaluated once each. Just not sure how to do this right now.

Additional stats to add include the number of sequencing groups in the latest ES-Index and in the latest joint-call.


EddieLF:
An implementation of the Seqr projects dashboard.

- Use new ProjectAPI endpoint `get_project_insights_stats(projects, sequencing_types)` to retrieve counts of families, participants, samples, sequencing groups, CRAMs, the latest joint call analysis, SNV ES-Index analysis, and SV ES-Index analysis - **for each dataset + sequencing type combination**
- Render these as rows in a table located at seqr/stats, with the option to select which projects and sequencing types to display

TODO:
- Refactor the naming of the new files and classes, etc. Currently a mess of "Project Insights", "Insights Stats", "Insights Details", "Seqr Stats"... will choose one to decide on
- Get information about the dataset analyses (AnnotateDataset, SNV Es-Index, SV ES-Index) more generically rather than with fixed fields in the data model
- Add more of this analysis information to the mouseover tooltips - e.g. Created at timestamp, Output path
- Add a "Detailed" view, which can provide sequencing group level granularity for the selected sequencing projects and types. Use `Virtuoso` to handle the rows without declaring explicit pagination.

db/python/layers/project_insights.py
```diff
@@ -0,0 +1,667 @@
...
+            self._stats_crams_query(projects, sequencing_types),
+            self._stats_annotate_dataset_query(
+                projects, sequencing_types
+            ),
+            self._stats_es_indexes_query(projects, sequencing_types,),
+        )
+
+        response = []
+        for pid in projects:
+            project = await ptable.get_and_check_access_to_project_for_id(self.author, pid, readonly=True)
```
illusional, line 466: A call to check access to multiple projects (save this async function being in a loop)

illusional (changes requested):
As discussed on zoom

api/routes/project_insights.py
```diff
@@ -0,0 +1,52 @@
...
+router = APIRouter(prefix='/project-insights', tags=['project-insights', 'summary', 'details', 'seqr', 'dashboard'])
+
+
+@router.post(
+    '/project-insights-summary',
+    operation_id='getProjectInsightsSummary',
+    response_model=list[ProjectInsightsSummary],
+)
+async def get_project_insights_summary(
+    project_ids: list[int] = None,
```
illusional, line 19: Return `project_names: list[str]`, and then do map to IDs (which checks project access)

db/python/layers/project_insights.py
```diff
@@ -189,26 +192,37 @@ async def _stats_sequencing_groups_query(
...
+            ]
+            for row in _query_results
+        }
+        return total_sequencing_groups_by_project_id_and_seq_fields
+
+    async def _crams_by_project_id_and_seq_fields(
+        self,
+        project_ids: list[ProjectId],
+        sequencing_types: list[str],
+    ) -> defaultdict[tuple[ProjectId, str, str], list[int]]:
```
illusional, line 219: I think, there's a SequencingGroupInternalId or something

db/python/layers/project_insights.py
```diff
@@ -220,18 +234,25 @@ async def _stats_crams_query(
...
+        for row in _query_results:
+            crams_by_project_id_and_seq_fields[
+                (row["project"], row["sequencing_type"], row["sequencing_technology"])
+            ].append(row["sequencing_group_id"])
+
+        return crams_by_project_id_and_seq_fields
+
+    async def _latest_annotate_dataset_by_project_id_and_seq_type(
+        self, project_ids: list[ProjectId], sequencing_types: list[str]
+    ) -> defaultdict[tuple[ProjectId, str], dict[str, Any]]:
```
illusional, line 255: Because you use SequencingTechnology / Type so much, just declare an alias above to allow this to read easier

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+        families_by_project_id_and_seq_fields = defaultdict(list)
+        for row in families:
+            families_by_project_id_and_seq_fields[
+                (
+                    row["project"],
+                    row["sequencing_type"],
+                    row["sequencing_platform"],
+                    row["sequencing_technology"],
+                )
+            ].append(row)
```
illusional, line 409: Beware returning row pointer here, it's not a dict. In this case, it carries a lot of extra fields, and you need to know the keys by looking the code. Consider returning a FamilyInternal object, as you can leverage typing to understand exactly what the response looks like

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
-    ):
-        """Get sequencing group web report links"""
+        latest_es_indices_by_project_id_and_seq_type_and_stage: defaultdict[
+            tuple[ProjectId, str, str], dict[str, Any]
+        ] = defaultdict(dict)
+        for row in _query_results:
+            latest_es_indices_by_project_id_and_seq_type_and_stage[
+                (row["project"], row["sequencing_type"], row["stage"])
+            ] = {
+                "id": row["id"],
```
illusional, line 349: Maybe create a namedtuple, or tiny dataclass for this, be your own friend for typing

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+AND JSON_EXTRACT(meta, '$.stage') = 'Stripy';
+        """
+        
+        _query_results = await self.connection.fetch_all(
+            _query,
+            {
+                "projects": project_ids,
+            },
+        )
+        return {(row['project'], row['sequencing_group_id']): row for row in _query_results}
```
illusional, line 449: Same here, don't return a row, look for a type you can serialise

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+    a.type,
+    a.timestamp_completed,
+    asg.sequencing_group_id,
+    JSON_EXTRACT(meta, '$.outliers_detected') as outliers_detected,
+    JSON_QUERY(meta, '$.outlier_loci') as outlier_loci
+FROM analysis a
+LEFT JOIN analysis_sequencing_group asg on asg.analysis_id=a.id
+INNER JOIN (
+    SELECT
+        asg.sequencing_group_id,
```
illusional, line 428: Return the a.id here, of the max completed (with order by).

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+    WHERE type='web'
+    AND project IN :projects
+    AND JSON_EXTRACT(meta, '$.stage') = 'Stripy'
+    GROUP BY asg.sequencing_group_id
+) max_timestamps ON asg.sequencing_group_id = max_timestamps.sequencing_group_id
+AND a.timestamp_completed = max_timestamps.max_timestamp
+WHERE a.project IN :projects
+AND a.type = 'web'
+AND JSON_EXTRACT(meta, '$.stage') = 'Stripy';
+        """
```
illusional, line 441: You don't need these conditions if you check on the inner query

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+        MAX(timestamp_completed) as max_timestamp
+    FROM analysis a
+    LEFT JOIN analysis_sequencing_group asg on asg.analysis_id=a.id
+    WHERE type='web'
+    AND project IN :projects
+    AND JSON_EXTRACT(meta, '$.stage') = 'MitoReport'
+    GROUP BY asg.sequencing_group_id
+) max_timestamps ON asg.sequencing_group_id = max_timestamps.sequencing_group_id
+AND a.timestamp_completed = max_timestamps.max_timestamp
+WHERE a.project IN :projects
```
illusional, line 474: Same treatement as above, but you can probably just ditch the inner query because the other fields are the same

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+AND a.type = 'web'
+AND JSON_EXTRACT(meta, '$.stage') = 'MitoReport';
+        """
+        _query_results = await self.connection.fetch_all(
+            _query,
+            {
+                "projects": project_ids,
+            },
+        )
+        return {(row['project'], row['sequencing_group_id']): row for row in _query_results}
```
illusional, line 484: Pick a type here

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+    
+    # Helper functions
+    async def _get_sequencing_groups_by_analysis_ids(
+        self, analysis_ids: list[int]
+    ) -> dict[int, list[int]]:
```
illusional, line 489: Looking for the SequencingGroupInternalId alias here

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+            for row in group_analysis_rows.values():
+                analyses_to_query_sequencing_groups.append(row["id"])
+
+        return await self._get_sequencing_groups_by_analysis_ids(
+            analyses_to_query_sequencing_groups
+        )
+
+    def get_sg_web_report_links(
+        self,
+        sequencing_group_stripy_reports,
```
illusional, line 532: Type annotations brother

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
-        self, projects: list[int], sequencing_types: list[str]
+        report_links: defaultdict[str, dict[str, str]] = defaultdict(dict)
+        
+        stripy_report = sequencing_group_stripy_reports.get(
+            (project.id, sequencing_group_id)
+        )
+        if stripy_report:
+            report_links['stripy'] = {
+                'stripy_url': (
+                    f'https://main-web.populationgenomics.org.au/'
```
illusional, line 548: Grab the gs path from the analysis-object, and add a transformation function. Easy to write tests for

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+                latest_annotate_dataset_by_project_id_and_seq_type,
+                latest_es_indices_by_project_id_and_seq_type_and_stage,
+            ]
+        )
+
+        sequencing_technologies = await SequencingTechnologyTable(
+            self._connection
+        ).get()
+
+        # Get all possible permutations of the projects, sequencing types, and sequencing technologies
```
illusional, line 625:
```suggestion
        # Get all possible combinations of the projects, sequencing types, and sequencing technologies
```

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+                    (project.id, sequencing_type, sequencing_technology)
                 ]
-                latest_annotate_dataset_id = (
-                    self.get_val_for_project_and_sequencing_type(
-                        project.id, sequencing_type, 'id', latest_annotate_dataset
-                    )
+            )
+
+            if sequencing_technology == "short-read":
+                latest_annotate_dataset = AnalysisStatsInternal(
```
illusional, line 647: This function is long, and nested makes it harder to read, consider moving this construction to it's own function

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
+                latest_es_indices_by_project_id_and_seq_type_and_stage,
+            ]
+        )
+
+        sequencing_platforms = await SequencingPlatformTable(self._connection).get()
+        sequencing_technologies = await SequencingTechnologyTable(
+            self._connection
+        ).get()
+
+        # Get all possible permutations of the projects, sequencing types, sequencing platforms, and sequencing technologies
```
illusional, line 785: combinations

db/python/layers/project_insights.py
```diff
@@ -301,367 +332,601 @@ async def _stats_es_indexes_query(
...
-                    )
-                ]
-                latest_snv_es_index_id = self.get_val_for_project_and_sequencing_type_and_stage(
-                    project.id, sequencing_type, 'MtToEs', 'id', latest_es_indexes
+            ]
+            if not family_rows:
+                continue
+
+            if sequencing_technology == "short-read":
+                latest_annotate_dataset = AnalysisStatsInternal(
```
illusional, line 812: Same function complexity break up here

models/models/project_insights.py
```diff
@@ -1,152 +1,160 @@
...
+    id: int | None
+    name: str | None
+    sg_count: int | None
+    timestamp: str | None
 
 
 @dataclass
 class ProjectInsightsDetailsInternal:
-    """Return class for the project insights details endpoint"""
+    """Return class for the seqr projects insights details endpoint"""
```
illusional, line 43:
```suggestion
    """Return class for the projects insights details endpoint"""
```

web/src/pages/projectInsights/Details.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+interface SelectedProject {
+    id: number
+    name: string
+}
+
+const Details: React.FC = () => {
+    // Get the list of  projects from the project API
+    const [allData, setAllData] = useState<ProjectInsightsDetails[]>([])
+    const [projectNames, setProjectNames] = React.useState<string[]>([])
```
illusional, line 16: @dancoates, can you take a look at the typescript + react on this function, to see if you have suggestions on improving the code here

web/src/pages/projectInsights/DetailsTable.tsx
```diff
@@ -0,0 +1,903 @@
+// DetailsTable.tsx
+import React, { useState } from 'react'
+import { ProjectInsightsDetails } from '../../sm-api'
+import { Button, Table, Icon } from 'semantic-ui-react'
```
illusional, line 4: Instead of the table here, look for the other defined Table in this function, you'll still want to import Table as SUITable so you can access the SUITable.Row / .Cell, etc. But our table definition has some handy dark mode, compact customisations

web/src/pages/projectInsights/DetailsTable.tsx
```diff
@@ -0,0 +1,903 @@
...
+            <Table.Cell className="table-cell">
+                {details.in_latest_snv_es_index ? '✅' : '❌'}
+            </Table.Cell>
+            <Table.Cell className="table-cell">
+                {details.in_latest_sv_es_index ? '✅' : '❌'}
+            </Table.Cell>
+            <Table.Cell data-cell className="table-cell">
+                {details.sequencing_group_report_links?.stripy ? (
+                    <a
+                        href={`https://main-web.populationgenomics.org.au/${details.dataset}/stripy/${details.sequencing_group_id}.html`}
```
illusional, line 101: Can you return the URL from the server?

web/src/pages/projectInsights/DetailsTable.tsx
```diff
@@ -0,0 +1,903 @@
...
+                ).map((option) => option?.toString() || '')
+        }
+
+        return uniqueOptions
+    }
+
+    const exportToFile = (format: 'csv' | 'tsv') => {
+        const headerCells = document.querySelectorAll('.ui.table thead tr th')
+        const headerData = Array.from(headerCells).map((cell) => cell.textContent)
+        const rows = document.querySelectorAll('.ui.table tbody tr')
```
illusional, line 304: Query with a specific table ID, rather than a generic selector

web/src/pages/projectInsights/DetailsTable.tsx
```diff
@@ -0,0 +1,903 @@
...
+        document.body.removeChild(link)
+    }
+
+    return (
+        <div>
+            <div style={{ textAlign: 'right' }}>
+                <Button onClick={() => exportToFile('csv')}>Export to CSV</Button>
+                <Button onClick={() => exportToFile('tsv')}>Export to TSV</Button>
+            </div>
+            <Table sortable>
```
illusional, line 342: he internal table

web/src/pages/projectInsights/DetailsTable.tsx
```diff
@@ -0,0 +1,903 @@
...
+                                sortColumns.find((column) => column.column === 'sequencing_type')
+                                    ?.direction
+                            }
+                            onClick={(event: React.MouseEvent<HTMLElement>) =>
+                                handleSort('sequencing_type', event.shiftKey)
+                            }
+                        >
+                            Seq Type
+                        </Table.HeaderCell>
+                        <Table.HeaderCell
```
illusional, line 368: 1. Pull this out as it's own component
2. You have a clear, what's required to actually this header cell
3. `[{name: "platform", title: "Platform", selectedOptions: selectedSeqPlatforms}, ...]`

web/src/pages/projectInsights/Summary.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const [allData, setAllData] = useState<ProjectInsightsSummary[]>([])
+    const [projectNames, setProjectNames] = React.useState<string[]>([])
+    const [projectIds, setProjectIds] = React.useState<number[]>([])
+    const [selectedProjects, setSelectedProjects] = React.useState<SelectedProject[]>([])
+    // New state for sequencing types
+    const [seqTypes, setSeqTypes] = React.useState<string[]>([])
+    const [selectedSeqTypes, setSelectedSeqTypes] = React.useState<string[]>([])
+    const [selectedSeqTechnologies, setSelectedSeqTechnologies] = React.useState<string[]>([])
+
+    useEffect(() => {
```
illusional, line 23: You _could_, get allMyProjects, and then just have the seqr ones selected

web/src/pages/projectInsights/Summary.tsx
```diff
@@ -0,0 +1,116 @@
...
+            case 'sequencing_technology':
+                setSelectedSeqTechnologies(selectedOptions)
+                break
+            default:
+                break
+        }
+    }
+
+    return (
+        <div>
```
illusional, line 92:
```suggestion
        <>
```

web/src/pages/projectInsights/Summary.tsx
```diff
@@ -0,0 +1,116 @@
...
+            />
+            <SummaryTable
+                allData={allData}
+                filteredData={filteredData}
+                selectedProjects={selectedProjects}
+                selectedSeqTypes={selectedSeqTypes}
+                selectedSeqTechnologies={selectedSeqTechnologies}
+                handleSelectionChange={handleSelectionChange}
+            ></SummaryTable>
+        </div>
```
illusional, line 112:
```suggestion
        </>
```

web/src/pages/projectInsights/SummaryTable.tsx
```diff
@@ -0,0 +1,733 @@
...
+    const percentageAligned =
+        summary.total_sequencing_groups > 0
+            ? (summary.total_crams / summary.total_sequencing_groups) * 100
+            : 0
+
+    const percentageInJointCall =
+        summary.latest_annotate_dataset?.sg_count ?? 0 > 0
+            ? ((summary.latest_annotate_dataset?.sg_count ?? 0) / summary.total_sequencing_groups) *
+              100
+            : 0
```
illusional, line 53:
```suggestion
    const latestAnnotateSgCount = summary.latest_annotate_dataset?.sg_count ?? 0
    const percentageInJointCall = latestAnnotateSgCount > 0
            ? (latestAnnotateSgCount / summary.total_sequencing_groups) * 100
            : 0
```

web/src/pages/projectInsights/SummaryTable.tsx
```diff
@@ -0,0 +1,733 @@
...
+                className="table-cell"
+                style={{
+                    textAlign: 'center',
+                    backgroundColor: getPercentageColor(percentageAligned, isDarkMode),
+                }}
+            >
+                <div style={{ display: 'flex', justifyContent: 'center' }}>
+                    <HtmlTooltip
+                        title={
+                            <p>
```
illusional, line 89: Remove the para tags, I think it'll be fine

web/src/pages/projectInsights/SummaryTable.tsx
```diff
@@ -0,0 +1,733 @@
...
+                                {summary.total_crams} / {summary.total_sequencing_groups} <br />
+                                Sequencing Groups Aligned
+                            </p>
+                        }
+                    >
+                        <div>{percentageAligned.toFixed(2)}%</div>
```
illusional, line 95: I don't think HTML loves divs being inside table cells

web/src/pages/projectInsights/SummaryTable.tsx
```diff
@@ -0,0 +1,733 @@
...
+
+        return uniqueOptions
+    }
+
+    return (
+        <div>
+            <Table sortable>
+                <Table.Header>
+                    <Table.Row>
+                        <Table.HeaderCell
```
illusional, line 288: Simplify here

illusional (changes requested):
As discussed on call, looping in @dancoates for react / typescript stuff

EddieLF:
Hi @illusional - once again thank you for the last round of feedback. Once again requesting a re-review from you :) 🙏 
Please feel free to handball this to anyone else if it helps get it through. My goal right now is to have this in a satisfactory state to merge, allowing it to be more easily iterated on by future, smaller PRs. 

## Main changes since last review
- Use `getMyProjects` instead of `getSeqrProjects` when fetching available projects
- Seriously overhauled typing in `project_insights_layer.py`, now everything is very heavily typed
- Removed repetitive chunks of code in the layer, and reduced nesting / indentation by defining more functions - specifically for the `get_project_insights_details` and `get_project_insights_summary` functions at the bottom of the layer
- Better querying in the layer, using max timestamp and joining on analysis ID where required, so queries are much more readable. Also using `GROUP_CONCAT` when gathering many values in a grouping query.
- Removed all references to project ID on the user-side / front-end, only project names are used now
- Added "Insights" as a dropdown to Metamist webui navbar, with two options, "Summary" or "Details" which take you to `/insights/summary` and `/insights/details` respectively.
- Heavily refactored frontend to remove instances of highly repetitive code, especially as relates to table headers / columns.
  - Now, `HeaderCell.tsx` defines the possible column names, as well as the list of column names specifically for both the Summary and Details tables.
  - These columns for each table are exported as `summaryTableHeaderCellConfigs` and `detailsTableHeaderCellConfigs` - these are then passed around and used by the other parts of the module. This means the table columns are all defined in a centralised place and reused from there, making future updates significantly easier. 
- Refactored `ProjectAndSeqTypeSelector.tsx` to make the code here less repetitive & more re-usable, now project and sequencing type selections are both instances of the same `Selector` component. 
- Refactored `SummaryTable.tsx` and `DetailsTable.tsx`
  - These components now contain only functions and information to render cell content, handle sorting, and handle exporting.
- Refactored `Summary.tsx` and `Details.tsx`
  - These components now handle getting the initial data, storing the state for which filters have been applied (all stored in the `selectedValues` and `filteredData` state - no longer one state per column filter) and passing the data on to the `Table` components.
- Added dedicated `FilterData.tsx` component, to handle the filtering generically. Now when users click the filters on each header, the values shown are a combination of the already-applied filters, and the available values of that column with no filtering applied. Bit hard to explain, but we don't want values in the same column to drop out of the available filters immediately after filtering on a value, otherwise we can't select multiple values to filter on.
- Moved the `HtmlToolTip.tsx` into its own component in `/web/shared/utilities/`. 
- Created dedicated testing module for the Project Insights endpoints.


## TODO in future
- Add more tests!
- Add endpoints to GraphQL with filtering
- Improve layers logic, further reduce repetitive sections of code, optimise & generalise queries where possible
- Add more columns (specifically STRipy report outlier loci and report dates in the DetailsTable, SV calling status in DetailsTable, QC report links in SummaryTable)
- Generally improve UI and UX 
- Add new dashboards with more visualisations of the state of analyses in the various projects
- Transparently display which analysis (e.g. the es-indices, the annotated dataset) each sequencing group is a part of in the DetailsTable view
- Unify the Dashboards so they are accessed from the same page, with a toggle to display which dashboard table specifically, while keeping the filtering and project/seq type selection the same
- Add "export to csv-tsv" button for the SummaryTable dashboard

EddieLF:
Screenshot 2024-07-02 at 9 25 11 AM
Screenshot 2024-07-02 at 9 26 09 AM

web/src/pages/insights/DetailsTable.tsx
```diff
@@ -0,0 +1,202 @@
...
+                ])
+            } else {
+                setSortColumns([{ column, direction: 'ascending' }])
+            }
+        }
+    }
+
+    const sortedData = React.useMemo(() => {
+        const data = [...filteredData]
+        data.sort((a, b) => {
```
illusional, line 105: Maybe consider using the lodash orderBy method, it allows multiple column sort, and returns the new sorted array for you. Though I think it uses the javascript sort operator

web/src/pages/insights/HeaderCell.tsx
```diff
@@ -0,0 +1,245 @@
...
+                    </HtmlTooltip>
+                ) : (
+                    label
+                )}
+            </div>
+        </SUITable.HeaderCell>
+    )
+}
+
+export const detailsTableHeaderCellConfigs: HeaderCellConfig[] = [
```
illusional, line 90: Naming conventions are [a bit over the place](https://stackoverflow.com/questions/40291766/naming-convention-for-const-object-keys-in-es6), but I'd probably suggest using screaming snake case, to suggest it's a global constant (and ditto below)

```suggestion
export const DETAILS_TABLE_HEADER_CELL_CONFIGS: HeaderCellConfig[] = [
```

web/src/pages/insights/SummaryTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+    key: ColumnKey,
+    isDarkMode: boolean
+): React.ReactNode => {
+    // What goes into each cell of the table is determined by the column header or 'key'
+    // The percentage calculation cells are handled differently from the other cells
+    const percentageAligned =
+        summary.total_sequencing_groups > 0
+            ? (summary.total_crams / summary.total_sequencing_groups) * 100
+            : 0
+    const alignedCellTooltip = (
```
illusional, line 74: This is great to have it pulled out, you'd usually go a step further and make this it's own tiny component

web/src/pages/insights/SummaryTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+        case 'exome':
+            return 'exome-row'
+        case 'genome':
+            return 'genome-row'
+        default:
+            return 'rna-row'
+    }
+}
+
+const getSummaryCell = (
```
illusional, line 63: This is a component, make sure to follow the standard component format + naming convention

web/src/pages/insights/SummaryTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+    }
+}
+
+const SummaryTableRow: React.FC<{ summary: ProjectInsightsSummary }> = ({ summary }) => {
+    const theme = React.useContext(ThemeContext)
+    const isDarkMode = theme.theme === 'dark-mode'
+    const rowClassName = getRowClassName(summary.sequencing_type)
+    return (
+        <SUITable.Row
+            key={`${summary.dataset}-${summary.sequencing_type}-${summary.sequencing_technology}`}
```
illusional, line 183: FWIW, you only need a key inside a .map within a function, and the key just has to be neighbourly unique (within a component).

web/src/pages/insights/SummaryTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+    }, [filteredData, sortColumns])
+
+    return (
+        <div>
+            <Table sortable>
+                <SUITable.Header>
+                    <SUITable.Row>
+                        {summaryTableHeaderCellConfigs.map((config) => (
+                            <HeaderCell
+                                key={config.key}
```
illusional, line 260: EG: here you do need the component key, and it's neighbourly unique which is perf 👌

web/src/pages/insights/SummaryTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+                                config={config}
+                                sortDirection={
+                                    sortColumns.find((column) => column.column === config.key)
```
illusional, line 263: Nitpick: you could probably use a map, rather than an `O(n)` lookup for each column

illusional (approved):
💪 Huge effort @EddieLF, nothing critical from me, I think this is pretty good to go!
Great work pushing on that abstraction and getting this code nice and clean - this looks nice and easy to follow :)

If you're able to address each comment before you merge, but I don't feel like I need to review this again. You'll theoretically need a bumpversion commit because you've added routes, but practically if it's only used in typescript, those are bundled together, so less stressed.

db/python/layers/project_insights.py
```diff
@@ -426,158 +682,105 @@ async def _details_stripy_reports(self, project_ids: list[ProjectId]) -> dict[Pr
...
     # Main functions
     async def get_project_insights_summary(
-        self, project_ids: list[ProjectId], sequencing_types: list[str]
+        self, project_names: list[str], sequencing_types: list[str]
     ):
         """Combines the results of the above queries into a response"""
         ptable = ProjectPermissionsTable(self._connection)
-        projects = await ptable.get_and_check_access_to_projects_for_ids(
-            user=self.author, project_ids=project_ids, readonly=True
+        projects = await ptable.get_and_check_access_to_projects_for_names(
```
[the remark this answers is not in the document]
EddieLF: Thanks for the heads up on this. 
Hey @dancoates, I just took a look through #829, what I am understanding is that a lot of the `ProjectPermissionsTable` methods like `get_and_check_access_to_project_for_id` and `get_and_check_access_to_projects_for_names` will be deprecated, and we should transition to using the method `get_projects_accessible_by_user` instead? 

If so, it should be fairly easy to adapt here. I will just have to add a check if the projects the user has access to fully overlap with the projects they're requesting in this function.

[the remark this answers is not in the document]
dancoates: Hey @EddieLF the `get_and_check...` methods are still around, they've just moved from the ProjectPermissionsTable to the `connection` which should be available pretty much everywhere. They are also synchronous now as there is no need for a call to the db. You shouldn't need to use the `get_projects_accessible_by_user` method directly. 

https://github.com/populationgenomics/metamist/blob/fbc2897002829ca4c7f542ec855d42beaf8fddb6/db/python/connect.py#L133

I'm planning to write up some docs for this but feel free to reach out if anything isn't clear in the meantime.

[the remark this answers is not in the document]
EddieLF: Ah gotcha, thanks for clearing that up.

web/src/pages/insights/SummaryTable.tsx
EddieLF: Pulled each of the "percentage" cells into their own individual components, e.g. `GetAlignedPercentageCell`, `GetAnnotatedDatasetPercentageCell`, etc. 

Now, the `GetSummaryCell` component is a lot simpler and to the point.
```tsx
const GetSummaryCell: React.FC<{
    summary: ProjectInsightsSummary
    columnKey: ColumnKey
    isDarkMode: boolean
}> = ({ summary, columnKey, isDarkMode }) => {
    switch (columnKey) {
        case 'aligned_percentage':
            return <GetAlignedPercentageCell summary={summary} isDarkMode={isDarkMode} />
        case 'annotated_dataset_percentage':
            return <GetAnnotatedDatasetPercentageCell summary={summary} isDarkMode={isDarkMode} />
        case 'snv_index_percentage':
            return <GetSnvIndexPercentageCell summary={summary} isDarkMode={isDarkMode} />
        case 'sv_index_percentage':
            return <GetSvIndexPercentageCell summary={summary} isDarkMode={isDarkMode} />
        default:
            return (
                <SUITable.Cell className="table-cell">
                    {summary[columnKey as keyof ProjectInsightsSummary]}
                </SUITable.Cell>
            )
    }
}
```

web/src/pages/insights/DetailsTable.tsx
EddieLF: lodash orderBy looks really well suited for this. We can pass the column names as an array, sort directions `'asc'`/`'desc'` array, and get the sorted array out. Seems really simple.

However it got a bit tricky in practice because the sorting logic is tied up in the semantic-ui-react `Table.HeaderCell` component, which has a `sorted` property that only allows the values `'ascending'` and `'descending'`. 

As far as I can tell, the sorting is mostly functional as it stands. I agree that using lodash orderBy will be more optimal, but I might leave that for the next PR if it's not too big a deal. Cheers

EddieLF:
@illusional I've brought in the auth changes and latest updates from dev and implemented them into the project_insights db layer here, as well as addressing most of the changes you outlined above.

I know there's some funky codecov stuff happening, but I am happy to have this merged this into dev if there are no urgent issues to address! Thank you very much again for your time and patience :)

PR #568, opened 2023-10-06
Inbuilt groups for permissions

illusional:
Inbuilt groups for permissions, allow us to be quicker with group lookups, and create custom groups easier for specific permissions.

This creates group + group_member tables, + adds some plumbing on the ProjectPermissions table.

Would require the pulumi bot to connect to metamist to update group members, probably not a hard task tbh.

---

When adding tests, I realised the method names and implementations are a bit over the place, so I pulled them apart, renamed a bunch and added cleaner implementations in some places.


db/python/tables/project.py
```diff
@@ -50,362 +28,309 @@ class ProjectPermissionsTable:
...
+    # region AUTH
 
-        missing_project_ids = [
-            project_id
-            for project_id, has_access in zip(spids, has_access_map)
-            if not has_access
-        ]
+    async def get_all_projects(self, author: str):
+        """Get all projects"""
+        await self.check_project_creator_permissions(author)
```
milo-hyben, line 105: Just a silly question.
Does this mean if user who is not in 'project-creators' and would call api route project/all, would get Forbidden exception?
Should user gets the list of all projects he/she is allowed to view?

milo-hyben (approved):

db/python/tables/project.py
illusional: Great question! There's a /my equivalent route for that, this was handy for specific applications that need all projects or fail

.pre-commit-config.yaml
```diff
@@ -48,15 +48,14 @@ repos:
...
-          [
+        args: [
             --pretty,
             --show-error-codes,
             --no-strict-optional,
             --ignore-missing-imports,
             --install-types,
             --non-interactive,
-            --show-error-context,
+            # --show-error-context,
```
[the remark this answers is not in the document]
illusional: It seemed to show too much context, and the regular does enough to solve most problems :)

PR #573, opened 2023-10-12
Auditor filter analyses

EddieLF:
Not sure if I've set these PR's up correctly but this builds on #572, filtering the analyses returned based on their project so that analysis runner submissions for a specific dataset don't break if they find an analysis outside of that dataset.

Also replaces the fixed analysis types with a query to the enum table.


illusional (approved):
Looks right to me, you'd usually need a bumpversion commit but you've got it in the previous PR

PR #582, opened 2023-10-22
Fix group-member-missing-check

illusional:
This method should definitely have been tested, codecov would probably have caught it.
Would be awesome to have some way to annotate a function to require it to be completely tested.


db/python/tables/project.py
milo-hyben, line 1: I do not like the name of the function: "check_which_groups_member_is_missing".
It is bit confusing, as the description says: """Check if a member is in a group"""
Should not be called some like "check_if_member_of_the_group"

milo-hyben (approved):

db/python/tables/project.py
illusional: Haha that's fair feedback, I've rewritten to avoid the negative :)

PR #585, opened 2023-10-25
Update sequencing group 'project:' filter to parse names, not ids

EddieLF:
Having some issues with my queries because the GraphQL `project:` filter seems to be expecting project IDs, however it only accepts string input so I can't get it to parse anything. I tried these changes in dev environment and they got my queries working again, however I'm not sure if this might impact anything else.

Understand there were a lot of changes to the project table recently, the speed improvements have been amazing 🏎️

### DIFF
--- a/api/graphql/schema.py
+++ b/api/graphql/schema.py
@@ -633,15 +633,11 @@ class Query:
         # we list project names, but internally we want project ids
         project_id_map = {}
         if project:
-            print(project)
             project_names = project.all_values()
-            print(project_names)
             projects = await ptable.get_and_check_access_to_projects_for_names(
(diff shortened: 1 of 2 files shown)


illusional (approved):
Ah yep, thanks @EddieLF, totally my bad. It could be worth adding this as a test case to catch regressions

MattWellie:
Paging @michael-harper  who was running into this yesterday

EddieLF:
Added an equivalent fix for filtering `sample` based on `project` as that had the same issue. Co-opted one of the existing tests that already used `sequencingGroup` as the root with `project` as an existing filter.

PR #590, opened 2023-10-30
Fix delete project endpoint

illusional:
Missing deletes from the old analysis_sample table - we don't write to it anymore, but we should clear it.


db/python/tables/project.py
```diff
@@ -389,6 +389,16 @@ async def delete_project_data(
...
+DELETE FROM analysis_sample WHERE sample_id in (
+    SELECT s.id FROM sample s
+    WHERE s.project = :project
+);
+DELETE FROM analysis_sequencing_group WHERE analysis_id in (
+    SELECT id FROM analysis WHERE project = :project
+);
+DELETE FROM analysis_sample WHERE analysis_id in (
+    SELECT id FROM analysis WHERE project = :project
+);
```
michael-harper, line 401: I'm not sure how to test this locally but it looks good. 

I guess just a clarification, `analysis_id` is a foreign key reference in both `analysis_sequencing_group` and `analysis_sample` but `analysis_sequencing_group` is a parent table of `analysis_sample` and so `analysis_id` should be deleted from it first?

michael-harper (approved):

michael-harper (approved):

PR #593, opened 2023-10-30
Generate seqr project data

EddieLF:
Developed this script to help with testing some endpoints being worked on that compile stats from seqr projects.

It largely adapts [test/data/generate_data.py](https://github.com/populationgenomics/metamist/blob/dev/test/data/generate_data.py), but instead of generating a single project using the pedigree file from the same folder, it generates a dozen projects, populates them with randomly generated pedigrees filled with randomly generated family and participant IDs, and relationships. Each project is then bulked out with random numbers of samples, sequencing groups, assays, and analyses. 

It also randomly allocates a subset of the sequencing groups as "aligned" and creates completed `CRAM` analyses for these. A subset of the aligned sequencing groups are allocated as the "joint-called" sequencing groups, and an `AnnotateDataset` `custom` analysis + `es-index` analysis are created containing these sequencing groups.

### DIFF
--- a/test/data/generate_seqr_project_data.py
+++ b/test/data/generate_seqr_project_data.py
@@ -288,7 +288,6 @@ async def main():
                         ),
                     },
                     participant_id=pid,
-                    assays=[],
                     sequencing_groups=[],
                 )
                 samples.append(sample)
@@ -319,6 +318,7 @@ async def main():
                                 type=assay_type,
                                 meta={
                                     'facility': facility,
+                                    'reads' : [],
                                     'coverage': f'{random.choice([30, 90, 300, 9000, "?"])}x',
                                     'sequencing_type': stype,
                                     'sequencing_technology': stechnology,
--- a/test/data/generate_seqr_project_data.py
+++ b/test/data/generate_seqr_project_data.py
@@ -5,9 +5,9 @@ import csv
 import datetime
 import logging
 import random
+import sys
 import tempfile
 
-import sys
 from pprint import pprint
 from typing import Set, List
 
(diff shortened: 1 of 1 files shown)


test/data/generate_seqr_project_data.py
```diff
@@ -0,0 +1,400 @@
...
+                        'specimen': random.choice(
+                            ['blood', 'phlegm', 'yellow bile', 'black bile']
+                        ),
+                    },
+                    participant_id=pid,
+                    sequencing_groups=[],
+                )
+                samples.append(sample)
+
+                for stype in generate_sequencing_type():
```
milo-hyben, line 297: The main function seems a bit too long. 
I would suggest to wrap those 2 most inner loops in a function (for stype: for _ ). Will be easier to follow.
Maybe even the whole samples list creation can be refactored as a function.

Otherwise looking good to me.

EddieLF: Thanks for the feedback Milo. Agreed that the main function was too long, I've moved the `samples` list creation into its own function and added some better comments which hopefully make it easier to follow.

violetbrina: I'd maybe clean it up a bit more. Break down the main function so it just makes function calls. Each commented block should be its own function. It just makes things so much easier to interpret.

I suspect this will be used (and potentially modified) a lot, so it's worth setting is up well if people come back and want to make significant changes later on.

violetbrina (changes requested):
Main function refactor

EddieLF:
Thanks for the feedback @violetbrina. 
I've refactored the main function so that the various steps are compartmentalized into their own functions. 

Now across only 4 lines in `main()`:
1. The families/participants/pedigrees are created and inserted
2. The samples/sequencing groups/assays are created and inserted
3. The CRAM analyses are inserted
4. The AnnotateDataset/ES-Index analyses are inserted

I also changed the analyses to initialise as an empty list prior to the project iteration, and then insert in chunks at the end.

violetbrina (approved):
Looks great to me!

PR #596, opened 2023-10-31
Bil 225

violetbrina:
Update to the cost over time page to show a table with the same data as the charts.


milo-hyben (approved):
Looking good!

PR #598, opened 2023-11-01
Seqr prop map v2

illusional:
- Also adds gt, gte, lt, lte filters
- Rewrites the implementation to be cleaner in my opinion, remembering that you must include all the past to correctly know. There are still concessions, like if a value is changed, we're not looking for crams at specific points in time. But this is good enough and close enough.

No bumpversion commit required as the API doesn't change.


violetbrina (approved):
Yay v2 woohoo! Looking good to me.

PR #600, opened 2023-11-02
Bil 242 - Hide billing pages when env variables aren't set

violetbrina:
Hide billing view in absence of billing data. 

Combines:
- Conditional import of the billing routes only if all of the required env variables are present
- Frontend only renders billing tabs if the routes are present and return 200 OK (uses getTopics to check)

I also did a few file moves to clean up the billing pages.


milo-hyben (approved):
Looking good!

PR #602, opened 2023-11-03
Audit logs

illusional:
Dumping some ✈️ thoughts on an audit log implementation.
Also adds a catch for ar-guid automatically to the audit log. Better than sticking it in meta on some entities :)

- A slight refactor of connections / imports to avoid circular imports
- Moves the ProjectId definition into a more sensible place (that would've been good to NOT do in this PR)
- Adds an audit_log_id to every single table
- Add some helper methods to make this audit_log_id widely available
- Fixes a bunch of tests that shouldn't check the author

I've tested this locally, and great to see a bunch of the same entities catch the same changelog ID!

Note: I wonder if instead of `auth_project` I should change it to `auth_group` so it's abundantly clear how they were able to access this data?


illusional:
Hey @milo-hyben! I think I've got a strong first pass, keen to collect some initial thoughts on this _audit_ log change :) 
I've only added it to one entity in graphql for now, with plans to extend that later, and then potentially go backwards to find all the entities that were modified by this change. Could be worth jamming more information onto this entity in a meta, like IP address, other request info, etc.

db/python/layers/audit_log.py
```diff
@@ -0,0 +1,29 @@
+# pylint: disable=too-many-arguments
```
milo-hyben, line 1: Is this pragma really needed here?

models/base.py
```diff
@@ -7,3 +7,15 @@
 
 class SMBase(BaseModel):
     """Base object for all models"""
+
+
+def parse_sql_bool(val: str | int | bytes) -> bool | None:
```
milo-hyben, line 12: Would not be better to have this func in db/python/utils or at least inside model/utils?

milo-hyben (approved):
Looking good to me!
I like the refactoring you have done together with adding audit_log, 
esp. Dbase and NotFoundError are in better spots. 
Added a few comments, not a biggie.

milo-hyben:
Just one more comment, so basically we replaced all author fields with audit_log_id.
I just wonder if one ar_guid process creates a record and then you have another one updates it. From the current logic we only keep that last audit_log_id in the updated record table. 
And if we delete a record we would lose audit track.

What about if we remove the audit_log_id from each table and would have a lookup table with fields like:

audit_log_id,
table_name,
primary key of record,
operation [INSERT,UPDATE,DELETE], potentially operation can be inside audit_log record.

db/python/connect.py
```diff
@@ -44,12 +46,43 @@ class Connection:
...
+    async def audit_log_id(self):
+        """Get audit_log ID for write operations, cached per connection"""
+        if self.readonly:
+            raise InternalError(
+                'Trying to get a audit_log ID, but not a write connection'
+            )
+
+        with audit_log_lock:
+            if not self._audit_log_id:
+                # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
illusional: Added a comment, but it would cause a circular import

models/base.py
illusional: Yep, good call!

api/utils/exceptions.py
```diff
@@ -1,4 +1,4 @@
-from db.python.connect import NotFoundError
+from db.python.utils import NotFoundError
```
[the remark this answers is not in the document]
illusional: Yes, not sure why this didn't collapse together automatically.

milo-hyben (approved):
Changes looking good to me.

PR #606, opened 2023-11-13
Getting gcp_project data from gcp_billing view instead of aggregate v…

milo-hyben:
SM_GCP_BQ_BILLING_VIEW is the new env variable.
There is a bit of challenge to create materialised view on top of gcp_billing_export.
Partition can only be the same as the base table or its subset, we want to use different field, so I am keeping the existing part_time field in the view, partition by it, so when we run query we add extra filter based on part_time.
As well part_time is off by 4-5 days, basically when GCP load data there is a few days delay so I had hardcoded to be 7 days.


violetbrina (approved):
Looks good to me!

PR #609, opened 2023-11-15
Billing addressing comments

violetbrina:
Some more comments are addressed here. It will be worth testing the merge into billing as well to double check all the functionality still works.


milo-hyben (approved):
Looking good!

PR #610, opened 2023-11-15
Handle in progress analysis entries. 

vivbak:
This really shouldn't be an issue anymore because we've removed the mechanism of creating in-progress analysis entries in production pipelines. Nevertheless these entries exist.

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -21,11 +21,16 @@ from collections import Counter
 
 from google.cloud import storage
 
-from metamist.apis import (AnalysisApi, AssayApi, FamilyApi, ParticipantApi,
-                           SampleApi)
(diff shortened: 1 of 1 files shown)


KatalinaBobowik (approved):

scripts/create_test_subset.py
```diff
@@ -21,17 +21,11 @@
 
 from google.cloud import storage
 
-from metamist.apis import AnalysisApi, AssayApi, SampleApi, FamilyApi, ParticipantApi
+from metamist.apis import (AnalysisApi, AssayApi, FamilyApi, ParticipantApi,
```
illusional, line 24: Can you revert this sorting, usually this means you have a conflict with isort and black in vscode.

illusional (approved):

scripts/create_test_subset.py
vivbak: Done!

PR #616, opened 2023-11-19
Update existing cohort parser with warning flag

### LINKED ISSUE #614: Introduce a Warning Flag for Missing Fastq Files in parse_existing_cohort.py
At the moment the [parse_existing_cohort.py](https://github.com/populationgenomics/metamist/blob/c05b11e6f3ee57a217353b3b95f29bd585059955/scripts/parse_existing_cohort.py) script raises a `ValueError` if fastq’s can’t be found for participants in manifests. This functionality is desired as it provides a way of highlighting missing data that the user and/or PM team may not be aware of

However, if this data is known to be missing and we still want to ingest the data, then currently the script is unable to allow this to happen. 

Proposed change is to add a flag to warn the user about these missing fastq’s rather than stop the script running. This flag is to be used when issues with fastq’s are known

michael-harper:
Closes #614 

Added warning flag to log missing data when parsing manifests as opposed to raising a `ValueError`. Also added test case that fails when a `ValueError` is raised and the warning flag is set to `True`.


vivbak:
Could you resolve the linting and test failures first? Also did you do a test-run of this change?

vivbak (changes requested):
Could you resolve the linting and test failures first? Also did you do a test-run of this change?

web/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
     "name": "metamist",
-    "version": "6.3.0",
```
vivbak, line 3: Did you intend to bump version here?

.gitignore
```diff
@@ -1,6 +1,7 @@
 db/postgres*.jar
 .vscode/
 env/
+venv/
```
vivbak, line 4: I think we should put this change in a separate PR if we want to add this, it's outside the scope of the change.

scripts/parse_existing_cohort.py
```diff
@@ -205,6 +219,12 @@ def get_existing_external_sequence_ids(self, participant_map: dict[str, dict]):
 @click.option(
     '--include-participant-column', 'include_participant_column', is_flag=True
 )
+@click.option(
+    '--allow-missing-files',
+    'allow_missing_files',
+    is_flag=True,
+    help='Set this flag to parse manifests with missing data',
```
vivbak, line 226: Nit, can we be more specific with this comment? i.e. 
```suggestion
    help='Set this flag to parse/ingest sequencing groups with missing reads',
```

test/test_parse_existing_cohort.py
```diff
@@ -232,3 +235,42 @@ async def test_existing_row(
         self.assertEqual(0, summary['assays']['update'])
 
         return
+
+    @run_as_sync
+    async def test_get_read_filenames_no_reads_fail(self):
+        """Test when allow_missing_files is False and records with missing fastqs, ValueError is raised"""
```
vivbak, line 241: Nit, readability. 
```suggestion
        """Test ValueError is raised when allow_missing_files is False and sequencing groups have no reads"""
```

test/test_parse_existing_cohort.py
```diff
@@ -232,3 +235,42 @@ async def test_existing_row(
...
+        )
+        parser.filename_map = {}
+
+        with self.assertRaises(ValueError):
+            # this will raise a ValueError because the allow_missing_files=False,
+            # and there are no matching reads in the filename map
+            await parser.get_read_filenames(sample_id='', row=single_row)
+
+    @run_as_sync
+    async def test_get_read_filenames_no_reads_pass(self):
```
vivbak, line 260: Could we also test that a warning is raised?

vivbak (changes requested):
Awesome job! Just a few nits.

web/package-lock.json
michael-harper: I did not, what version should it be?

vivbak (approved):
Great job! We need a release for this to work. I'll action now.

PR #620, opened 2023-11-21
Fix test subset script bugs

EddieLF:
Addressing some issues discussed in the slack thread [here](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1700532187689939).

Main fixes:

1. When creating the `AssayUpsert`, we iterate through the assays being moved and check if they already exist in the test project. However the `get_existing_assay` function currently does not match the returned assay with the assay from the iteration. This change passes the entire `assay` entry to the `get_existing_assay` function. It then checks through the existing assays and if the type and meta fields match the input `assay`, then the existing assay that matches is returned.

2. When upserting participants in the `transfer_participants()` function, we check if the participants already exist in the test project. 
If they don't, we insert them as new participants into the test project. 
**If they do already exist, we update the participant.** 

In the latter circumstance, the participant being updated will be the _original participant_, that is, the participant from main. So as the script progresses, any samples, sequencing groups, assays, and analyses that we want to insert into test will instead be inserted into the participant in the main project.

This change creates a `external_id : internal_id` map for participants in the test project, and if a participant already exists in the test project then we will update it using its test participant id, obtained from the map.


vivbak (changes requested):
Hey @EddieLF! Thanks so much & this looks good to me. Could we add some new tests to capture these cases better?

EddieLF:
> Hey @EddieLF! Thanks so much & this looks good to me. Could we add some new tests to capture these cases better?

Thanks @vivbak.

I've reformatted the `create_test_subset.py` script with a class definition for `TestSubsetGenerator`. I basically just moved the existing code into class methods and cleaned up the main function by calling the various class methods. I also added in a locally defined random seed so that the random seeding is consistent for the test cases.

I made a start on the test cases, however I'm a little stuck on testing the transferring of participants into the test project. In the test cases, the default project created is `"test"`. I think creating a project called `"test-test"` to test this script is fine (albeit a little confusing), but I've not been able to get the `create_project()` method working from the project permissions table.

vivbak (changes requested):
Hey @EddieLF, 

Firstly I am sorry! When I suggested the test cases I for some reason thought we already had an established testing framework for this script. I did not realise it would be this much work, and wouldn't have asked if I did! 

But great job! I think it's important to do. Can we create a new draft PR to fix testing for this script. Cherry-pick all of this wonderful refactoring and test work you've done to that PR, and strip this back down to the bug fixes required? 

I ask because this PR is blocking some PA work, and the discussion/work on testing will probably take some time. 

I can run some tests locally & manually so we can feel more confident in your bug fixes for now.

EddieLF:
@vivbak thanks again for the feedback. I've stripped this PR back to just the bugfixes, and the refactoring is now in a new draft PR [here](https://github.com/populationgenomics/metamist/pull/621).

vivbak (approved):
Amazing thanks so much @EddieLF! Awesome job 🙌 I'm going to merge this now so we can get a release going.

scripts/create_test_subset.py
```diff
@@ -583,14 +583,16 @@ def transfer_participants(
...
-        participant['external_id'] for participant in existing_participants
-    ]
+    target_project_pid_map = {
+        participant['external_id']: participant['id'] for participant in existing_participants
+    }
 
     participants_to_transfer = []
     for participant in participant_data:
-        if participant['externalId'] not in target_project_epids:
+        if participant['externalId'] in list(target_project_pid_map.keys()):
```
illusional, line 592: I'm coming to this late and this isn't a super big deal because it's in auxillary code, but this line turns an O(1) lookup into an O(n). Can just leave it as:

```suggestion
        if participant['externalId'] in target_project_pid_map:
```

EddieLF: This is actually really awesome to know, cheers @illusional. I'll make sure thats fixed in the refactoring PR 👍

PR #626, opened 2023-11-27
Billing API IsBillingEnabled

milo-hyben:
Added simple API point to check if Billing is enabled.
ATM it checks only for env var regarding Billing Aggreg view, which should be enough.
I do not think we would need to check all the billing env variables.
I had added as well check to each billing API point, so it fails soon.


violetbrina (approved):
Have we checked that this actually works?

milo-hyben:
> Have we checked that this actually works?

Yes, I did test it, both with SM_GCP_BQ_AGGREG_VIEW defined and undefined.

PR #628, opened 2023-11-29
Tob wgs metadata harmonisation

michael-harper:
This script is designed to harmonise assay metadata from the old tob-wgs project with the new metadata schema. This new metadata schema is not yet agreed upon.

Key functionalities include:

- Checking assay metadata for each assay in the project.
- Verifying if each field is present in the FIELD_MAPPING dictionary.
- Creating a new_assay_data dictionary with harmonised field names as keys.
- Printing a ValueError if a field is not in the mapping dictionary.
- Ensuring that if a field is in the dictionary but has a None value, it is not added to the new_assay_data.
- Comparing values if multiple mappings lead to the same field, printing a console message if values differ.


tob_metadata_harmonisation.py
```diff
@@ -0,0 +1,252 @@
...
+      id
+      assays {
+        id
+        meta
+      }
+    }
+  }
+}
+"""
+SG_ASSAY_QUERY = """
```
MattWellie, line 48: this is also a duplicate query, but with neither true or false specified? I'm still working through this, but could this either be replaced with the two individual true/false calls, or could you just use one query for everything where `active` is returned as an attribute, then filter the results for true/false as required

tob_metadata_harmonisation.py
```diff
@@ -0,0 +1,252 @@
...
+        It also checks if there are any fields in the assay data that are not in the FIELD_MAPPING dictionary and raises
+        an error if any are found.
+        """
+        harmonised_data = {}
+
+        # check if multiple fields map to the same harmonised field
+        new_assay_data: Dict[str, Dict] = {}
+        for old_field, new_field in FIELD_MAPPING.items():
+            old_value = assay['meta'].get(old_field)
+            if old_value is not None:
```
MattWellie, line 202:
```suggestion
            if old_value := assay['meta'].get(old_field):
```
Walrus!

MattWellie (approved):
I don't have too much insight here, but seems reasonable. Could be slimmed down in a few places

tob_metadata_harmonisation.py
MattWellie: Just checked, this isn't used at all - should probably be removed

tob_metadata_harmonisation.py
```diff
@@ -0,0 +1,252 @@
...
+    sequencingGroups(activeOnly: {eq: false}) {
+      id
+      assays {
+        id
+        meta
+      }
+    }
+  }
+}
+"""
```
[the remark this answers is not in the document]
michael-harper: Good call!

tob_metadata_harmonisation.py
michael-harper: I found out that if you have `activeOnly:{}` without a boolean in there, it returns both `active` and `inactive`

But yes, you're right, this can be removed

PR #631, opened 2023-12-10
default-sequence-type change to align with input parameter requirements

michael-harper:
Changed click.option flag from default-sequence-type to default-sequencing-type to align with input parameter requirements


EddieLF:
Looks good to me

EddieLF (approved):

PR #635, opened 2023-12-14
Create md5s script

EddieLF:
Iterates through a gs directory and creates .md5 checksum files for each file in the directory. Useful for composite objects since GCP does not provide md5 checksums for these by default. Skips `.crai` and `.tbi` files as this will mostly be used on directories containing CRAM / gVCFs and their indexes - and the latter do not need checksums.

### DIFF
--- a/scripts/create_md5s.py
+++ b/scripts/create_md5s.py
@@ -54,10 +54,7 @@ def create_md5(job: hb.batch.job, file) -> hb.batch.job:
     job.command(
         f"""\
     gcloud -q auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS
-    md5_hash=$(gsutil cat {file} | md5sum | cut -d " " -f1) 
-    spaces='  '
-    md5_command="printf '%s%s%s' '${{md5_hash}}' \"${{spaces}}\" '{os.path.basename(file)}' > /tmp/uploaded.md5"
-    eval $md5_command
+    gsutil cat {file} | md5sum | cut -d " " -f1  > /tmp/uploaded.md5"
     gsutil cp /tmp/uploaded.md5 {md5}
     """
     )
--- a/scripts/create_md5s.py
+++ b/scripts/create_md5s.py
@@ -43,7 +43,8 @@ def validate_all_objects_in_directory(gs_dir):
 
 def create_md5(job: hb.batch.job, file) -> hb.batch.job:
     """
-    Streams the file with gsutil and calculates the md5 checksum, 
+    Streams the file with gsutil and calculates the md5 checksum,
+    then uploads the checksum to the same path as filename.md5.
     """
 
     # Calculate md5 checksum.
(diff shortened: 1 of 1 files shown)


scripts/create_md5s.py
```diff
@@ -0,0 +1,71 @@
+import os
+from typing import Set
+
+import click
+from cpg_utils.hail_batch import get_batch, get_config, copy_common_env
+from google.cloud import storage
+
+DRIVER_IMAGE = 'australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest'
```
illusional, line 8: Get this from the config :)

scripts/create_md5s.py
```diff
@@ -0,0 +1,71 @@
...
+
+    config = get_config()
+    if not billing_project:
+        billing_project = config['hail']['billing_project']
+
+    bucket_name, *components = gs_dir[5:].split('/')
+
+    client = storage.Client()
+    blobs = client.list_blobs(bucket_name, prefix='/'.join(components))
+    files: Set[str] = {f'gs://{bucket_name}/{blob.name}' for blob in blobs}
```
illusional, line 24:
```suggestion
    files: set[str] = {f'gs://{bucket_name}/{blob.name}' for blob in blobs}
```

scripts/create_md5s.py
```diff
@@ -0,0 +1,71 @@
...
+    for obj in files:
+        if obj.endswith('.md5') or obj.endswith(skip_filetypes):
```
illusional, line 26: Does endswith take a tuple here, that feels wild to me

scripts/create_md5s.py
```diff
@@ -0,0 +1,71 @@
...
+            continue
+        if f'{obj}.md5' not in files:
```
illusional, line 28: Do you only want to create if it doesn't exist, or add some option to force recreate?

scripts/create_md5s.py
```diff
@@ -0,0 +1,71 @@
...
+            print('Creating md5 for', obj)
+
+        job = b.new_job(f'Create {os.path.basename(obj)}.md5')
+        copy_common_env(job)
+        job.image(DRIVER_IMAGE)
```
illusional, line 33: Maybe move these into the create_md5 function?

scripts/create_md5s.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def create_md5(job, file, billing_project):
+    """
+    Streams the file with gsutil and calculates the md5 checksum,
+    then uploads the checksum to the same path as filename.md5.
+    """
+    md5 = f'{file}.md5'
+    job.command('set -euxo pipefail')
+    job.env('GOOGLE_APPLICATION_CREDENTIALS', '/gsa-key/key.json')
```
illusional, line 46: I thought this came through with `copy_common_env`, but I guess no harm setting this

illusional (approved):
Approve with nitpicks

PR #638, opened 2024-01-02
Bump pre-commit versions for the new year

illusional:
We should add isort soon, maybe after the billing PR + audit log to avoid lots of conflicts


milo-hyben:
Sounds good to me.

milo-hyben (approved):

PR #643, opened 2024-01-09
Bump server versions

illusional:
- FastAPI new lifecycle method: https://fastapi.tiangolo.com/advanced/events/
- Field aliases: https://docs.pydantic.dev/latest/concepts/fields/#field-aliases
- `model_*` protected naming was a warning in the console
- - `orm_mode` -> `from_attributes` was a warning in the console


milo-hyben (approved):
Looking good to me, but pip is not happy :-)

illusional:
Upgrading fastapi (past [this commit](https://github.com/tiangolo/fastapi/pull/9770)) meant our openapi version was pushed to 3.1, which was a pain because:

- We use an intentionally outdated openapi-generator (as newer ones didn't work), which doesn't support 3.1
- Swagger-ui-react should technically support 3.1, but bumping this up to a version that does causes pain and conflicts with React and React Router.

So we should park this for now, and try again later. Though, to upgrade this, we may need to ditch our current API generator, or hack it to make it work with 3.1.

PR #647, opened 2024-01-10
fix!: Allow null value on externalIds in the Assay model

### LINKED ISSUE #547: externalId Assay field can't be returned in GQL
I'm not sure if this is a bug, or 

Screenshot 2023-09-15 at 6 09 58 pm

Querying for an existing assay, the external_ids field is empty

https://sample-metadata.populationgenomics.org.au/api/v1/assay/172688/details

nevoodoo:
## Description
This addresses #547 and allows the query to handle null values for `externalIds` by returning an empty dictionary instead of raising an exception.

## Changes Made
- Patched the `GraphQLSequencingGroup` model in `api/graphql/schema.py` to allow only JSON type for the `external_ids` field, to be consistent with the `GraphQLAssay` model.
- Changed the `SequencingGroupInternal` model in `models/models/sequencing_group.py` to default to an empty `{}` for `external_ids`.
- Change the `AssayInternal` model in `models/models/assay.py` to default to an empty `{}` for `external_ids`.

## Closes
- [x] Closes #547


illusional (approved):
Awesome, thanks for the quick fix!

PR #650, opened 2024-01-11
Squash GQL Transport Logs

### LINKED ISSUE #540: Squash GQL result logging
# Issue

Our use of the `gql` library is resulting in all our sample-metadata results being written to log files on query execution.

https://github.com/populationgenomics/sample-metadata/blob/e94a4d8c1b8d33d439fb7abf0e38128911c0689b/metamist/graphql/__init__.py#L135

Can I gauge thoughts on adding an argument to `query`/`query_async` which adjusts the logging sensitivity? 

By default the [gql.client.execute module](https://github.com/graphql-python/gql/blob/d4c975198ad617265e4abb4716e5643f31b75e08/gql/transport/requests.py#L24) obtains a logger without setting an explicit level. A lot of the code we run is done with a logger keyed to INFO, which is resulting in over a MB of sample-metadata results being farted into the Hail logs every time a metadata query is executed. [This](https://batch.hail.populationgenomics.org.au/batches/427395/jobs/1) pipeline invocation totalled 1.7MB of just metamist results.
 
Probably the lines responsible:
https://github.com/graphql-python/gql/blob/d4c975198ad617265e4abb4716e5643f31b75e08/gql/transport/requests.py#L225-L227

I've put a `don't log this` escape around gql calls before [e.g. here](https://github.com/populationgenomics/saige-tenk10k/blob/main/genotypes_processing.py#L99C17-L101), but might be a good idea to implement something similar upstream in this library. This feels like a pretty ghetto solution but could be effective:

```python
import logging
from typing import Bool, Dict
...

def query(
    _query: str | DocumentNode, variables: Dict = None, client: Client = None, log_reponse: Bool = False
) -> Dict[str, Any]:
    """Query the metamist GraphQL API"""
    if variables is None:
        variables = {}
    
    current_level = logging.root.level  # even if not set this defaults to WARN(30) on logging import
    if current_level >= logging.INFO and not log_response:
        logging.getLogger().setLevel(logging.WARN)
    response = (client or configure_sync_client()).execute_sync(
        _query if isinstance(_query, DocumentNode) else gql(_query),
        variable_values=variables,
    )
    # reset logger to baseline
    logging.getLogger().setLevel(current_level)
    return response
```

A decorator to accomplish the same would be way sexier, but I can't put one of those together in an Issue text box

nevoodoo:
## Description
This addresses #540 and allows suppressing `gql` logging to a less verbose level  of `WARNING` instead of `INFO`.

## Changes Made
- As per @MattWellie 's suggestion, introduced a `log_response` argument to `query` and `query_async`  methods in `metamist/graphql/__init__.py`,  that can be toggled to set the level to `WARNING` jut for that query execution before setting it to the original level.

## Notes
- I was not able to replicate this locally despite trialing with the `logger` from `gql.transport.[requests | aiohttp]` and the `logging` libraries. Would be nice to test this somehow to see if it works as expected. This should maintain the same functionality as @MattWellie 's fix in theory. 

## Closes
- [x] Closes #540

### DIFF
--- a/metamist/graphql/__init__.py
+++ b/metamist/graphql/__init__.py
@@ -5,11 +5,13 @@ GraphQL utilities for Metamist, allows you to:
     - validate queries with metamist schema (by fetching the schema)
 """
 import os
-from typing import Dict, Any
+from typing import Dict, Any, Bool
 
 from gql import Client, gql as gql_constructor
 from gql.transport.aiohttp import AIOHTTPTransport
+from gql.transport.aiohttp import log as aiohttp_logger
 from gql.transport.requests import RequestsHTTPTransport
+from gql.transport.requests import log as requests_logger
 
 from cpg_utils.cloud import get_google_identity_token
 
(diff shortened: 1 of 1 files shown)


illusional (approved):
Awesome!

PR #651, opened 2024-01-16
Update md5 creating script for requester pays buckets

EddieLF:
This fix allows the script to be used to create md5s for files in requester pays buckets. 
**Includes a bumpversion patch to facilitate a release.**

Previously, the filepaths were obtained by calling `list_blobs` on the `storage.Client()` object
```
client = storage.Client()
blobs = client.list_blobs(bucket_name)
```

This left no opportunity to specify a billing project, so files in requester pays buckets could not be accessed. 

A small change allows specifying a user project to bill to by getting the `bucket` object and working with that:
```
client = storage.Client()
bucket = client.bucket(bucket_name, user_project=billing_project)
blobs = bucket.list_blobs()
```

Also adds the `-u <billing_project>` argument to the `gsutil cat` command, which was also failing when attempting to read files in a requester pays bucket.


illusional (approved):
Supportive, and this looks fine. You don't need a bumpversion commit because this file is in scripts, and not in the deployed API. When you run this through the analysis-runner, you implicitly check out this repo at the commit, and the script directly (not through the installed metamist python library).

Can you add a commit that explicitly ignores mypy on the linting failed line.

```
api/server.py:147: error: "Middleware" has no attribute "options" 
[attr-defined]
                and request_origin in cors_middleware.options['allow_origi...
```

EddieLF:
Thanks @illusional, to run the script with `full` access level I need this script to be merged into main. 
If I merge this into dev, can I then open a PR to merge dev into main straight away without any other actions? Or will that need a bumpversion commit?

illusional:
Yep @EddieLF, you can trigger the dev -> main release. There's one other commit in there, but it doesn't need a bumpversion commit. Only some changes to models, and the api.

PR #653, opened 2024-01-17
Populate tob-wgs-test metamist with nagim gvcf files

michael-harper:


scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+    help='The name of the project to add samples to.',
+)
+@click.option(
+    '--project_id',
+    required=True,
+    type=int,
+    help='The ID of the project to add samples to.',
+)
+@click.option(
+    '--new_samples_path',
```
illusional, line 60: Historical bash command line params use dashes - instead of underscores.

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+    required=True,
+    help='The path to the file containing the newly created samples.',
+)
+def main(project: str, projectID: int, new_samples_path: str):
```
illusional, line 64: Is there a reason you want the project and project ID? You are getting this in the graphql query above

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+    # Read the CSV file into a dictionary
+    extID_to_row = {}
```
illusional, line 66: Python variable names are snake case (no capitals except for constants). Can you fix these up below?

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+        project(name: $project) {
+            id
+            participants {
+            externalId
+            id
+            samples {
+                id
+                externalId
+                sequencingGroups {
+                externalIds
```
illusional, line 28: Can you format this to show theyre fields on the sequencingGroup?

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+@click.option(
+    '--project_id',
+    required=True,
+    type=int,
+    help='The ID of the project to add samples to.',
+)
+@click.option(
+    '--new_samples_path',
+    required=True,
+    help='The path to the file containing the newly created samples.',
```
illusional, line 62: Can you be more descriptive to what this means? Including the expected structure of the file, and maybe an anonymised example of the first couple of lines?

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+                        technology=sg['technology'],
+                        platform=sg['platform'],
+                        meta=None,
+                        sample_id=None,
+                        external_ids=None,
+                        assays=[
+                            AssayUpsert(
+                                type=sg['assays'][0]['type'],
+                                meta={
+                                    'sequencing_type': sg['assays'][0]['meta'][
```
illusional, line 104: These are already fields on the sg that you're selecting from graphql

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+                    )
+                )
+            p.samples.append(s)
+        p_upserts.append(p)
+
+    upserted_participants = ParticipantApi().upsert_participants(project, p_upserts)
+
+    for participant in upserted_participants:
+        for sample in participant['samples']:
+            old_extID = sample['externalId'][:-5]  # remove '-test' from the end
```
illusional, line 125: Use .removesuffix here instead of the substring

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+            gvcf_path = extID_to_row[old_extID][1]  # get gvcf path from dictionary
+            AnalysisApi().create_analysis(
+                project,
+                Analysis(
+                    type='gvcf',
+                    status=AnalysisStatus('completed'),
+                    output=gvcf_path,
+                    sequencing_group_id=sample['sequencingGroups '][0]['id'],
+                    project=projectID,
```
illusional, line 134: You don't need to fill this project in here, as it's in the call above. This is an issue with metamist using the same model to create, and for created models.

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+                    active=True,
+                    external_id=sample['externalId'],
```
illusional, line 136: I don't think these last two fields external_id and sample are required or possible to fill in

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,143 @@
...
+)
+def main(project: str, projectID: int, new_samples_path: str):
+    # Read the CSV file into a dictionary
+    extID_to_row = {}
+    with to_path(new_samples_path).open() as f:
+        reader = csv.reader(f)
+        next(reader)  # skip the header
+        for row in reader:
+            sgid, extID, gvcf, gvcf_idx = row[0], row[1], row[2], row[3]
+            extID_to_row[extID] = (sgid, gvcf, gvcf_idx)
```
illusional, line 72: You should consider using a namedtuple or data class here to be super clear how this data is organised.

illusional (changes requested):
Looking good! I've left some inline code quality suggestions. I'd also probably suggest that this script can be genericised with a little bit of work. A suggestion: migrate participant metadata to test.

You could parameterise the suffix added on to avoid externalID clashes

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
michael-harper: Oh that's super cool. I've just implemented a dataclass!

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
michael-harper: So something like this?


```python
sg=SequencingGroupUpsert(
    sg,
    assays=[
        ...
    ]
``` 
I'm a bit confused by what you're referring to

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
michael-harper: I changed the parameter name because it doesn't actually reflect what we're parsing, here's the new:
```python
@click.option(
    '--sample-path-mappings',
    required=True,
    help='''The path to a CSV file containing mappings of `main` CPG ID's, 
            the `external_id` and `gvcf` paths. 
            The file should have at least four columns: sgid, ext_id, gvcf, and gvcf_idx. 
            Here's an example of what the first couple of lines might look like:

            sgid,ext_id,gvcf,gvcf_idx
            sg1,ext1,gvcf1,gvcf_idx1
            sg2,ext2,gvcf2,gvcf_idx2
        '''
)
```

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
illusional:
```suggestion
                                    'sequencing_type': sg['type']
```

Something like this

illusional: You can apply the same treatment to the platform and platform

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,198 @@
...
+            old_ext_id = sample['externalId'].removesuffix(f'-{suffix}')
+            row_data = ext_id_to_row[old_ext_id]
+            gvcf_path = row_data.gvcf  # get gvcf path from dictionary
+            AnalysisApi().create_analysis(
+                project,
+                Analysis(
+                    type='gvcf',
+                    status=AnalysisStatus('completed'),
+                    output=gvcf_path,
+                    sequencing_group_id=sample['sequencingGroups '][0]['id'],
```
illusional, line 189:
```suggestion
                    sequencing_group_id=sample['sequencingGroups'][0]['id'],
```

scripts/add_nagim_gvcfs_to_tob_wgs_test_metamist.py
```diff
@@ -0,0 +1,198 @@
...
+                    active=True,
+                ),
+            )
+
+
+if __name__ == '__main__':
+    # pylint: disable=no-value-for-parameter
+    main()
+    # pylint: enable=no-value-for-parameter
```
illusional, line 198:
```suggestion
    main()  # pylint: disable=no-value-for-parameter
```

illusional (approved):
Awesome work - looking great! Can you rename the file, and retitle the PR.
Once you apply that review feedback, you're good to merge (I don't need to see it again if you can action this feedback).

You'll also need to merge `dev` to `main` if you want to run this on main (but I don't think you need to, you can just run this as an analyst - as analysts can write to the metamist test project).

PR #657, opened 2024-01-23
First class files

nevoodoo:
## Description
Currently within metamist, we record paths in a few different ways, but inherently they’re stored in a text field. This means it’s difficult to work out:

- Where files are used in the centre,
- Metadata about specific files

The way we represent analysis and outputs has also evolved through the usage of metamist, and this is a good time to reflect on our future usage.

This replaces the current `Analysis` model to use a proper `File` class for it's `output` field instead of a `str`. This will allow us to support multiple outputs in a structured way along with useful file stats. 

This issue has also been raised before with some prior work being done on it a while ago on #376 

## Proposal
We should model our File off the CWL definition (https://www.commonwl.org/v1.2/CommandLineTool.html#File). This has been partially implemented in the parsers ([github:metamist/parser/generic_parser.py#L1328-L1340](https://github.com/populationgenomics/metamist/blob/ad85785af0c39baca3c8a24aeb591076b158b7d8/metamist/parser/generic_parser.py#L1328-L1340)). It’s probably reasonable to make this a table with foreign keys out to place.

We should additionally store whether a file exists or not, it will provide more utility than removing references to specific files.

Consider how to reference secondaryFiles to ensure efficient queries. It’s safe to assume that we’ll only see one level of nesting as secondaryFiles, and secondaryFiles won’t be linked directly to analysis / assays.

### References to files 

*Analysis*
- Currently we link one output to an analysis object
- As part of this, it would be great to support multiple outputs per analysis (potentially a JSON?)
*Assays*
- reads?: A list of reads that were provided as part of the experiment
- variants?: A file with variants specified. This is likely more legacy 
- *?: We should have flexibility to provide arbitrary files to assays

### Migration
We’ll need to write a migration that takes the existing references to files, and migrates them to this new structure. This can act directly on a database, and could be run manually.

### Extension
Later, we should capture pub/sub notifications for google cloud storage buckets, so we can mark files as archived / deleted to automatically deprecate analysis.

### Special Considerations
- `.mt` files need to be treated differently as they aren't technically files but rather directories on cloud storage. This means it will not have a checksum. We should also store `meta` for the `.mt` derived from the `matrix-type` key under the corresponding `metadata.json.gz` file. - ON HOLD for now
- Once we support a nested object containing the multiple outputs for each analysis, we should ideally be tracking the structure of the outputs received so that for all queries, we can restructure the response json based off this.
- The current `output` field will be deprecated in favour of the newer `outputs` field which will accept the nested object containing multiple outputs. This is to avoid any confusion regarding interchangeability between the two fields.
- Secondary files will be supplied during the creation/update of the analysis/assay as part of the nested object in the `outputs` field.
- Files are never deleted, even if the relationship between the file and the analysis/assay is broken. We simply remove the connection between the two entities.
- For output values that currently do not have proper `gs` prefixes, a separate `output` column will be added to the `analysis_file` relationship to store this, as a proper file will not be created without a valid checksum.
- The checksum implementation requires use of the `crc32c` hash as MD5 does not support multi-part uploads, and a lot of files may be missing the checksum on cloud storage. We advise you to use [Google's Python crc32c wrapper](https://github.com/googleapis/python-crc32c) to validate the file checksum in your code.

## Changes
- [x] New File model
- [x] Liquibase XML to new schema
- [x] Migration Script for existing objects
- [x] API/GraphQL support for new model
- [x] Test cases written
- [ ]  Update dbdiagrams UML


milo-hyben:
Hey @nevoodoo, I had briefly look at the sql schema and I am not big fan of table name 'file'. :-)
Wonder if we should call it e.g. analysis_file, there are already various analysis_*  tables.

api/routes/analysis.py
```diff
@@ -45,7 +45,7 @@ class AnalysisModel(BaseModel):
     type: str
     status: AnalysisStatus
     meta: dict[str, Any] | None = None
-    output: str | None = None
+    outputs: str | None = None
```
milo-hyben, line 48: Should not outputs be of type Union[dict, str] ?

db/python/layers/analysis.py
```diff
@@ -7,6 +8,7 @@
 from db.python.layers.base import BaseLayer
 from db.python.layers.sequencing_group import SequencingGroupLayer
 from db.python.tables.analysis import AnalysisFilter, AnalysisTable
+from db.python.tables.file import FileTable
```
milo-hyben, line 11: Should this be OutputFileTable?

db/python/layers/analysis.py
```diff
@@ -48,6 +50,7 @@ def __init__(self, connection: Connection):
 
         self.sampt = SampleTable(connection)
         self.at = AnalysisTable(connection)
+        self.ft = FileTable(connection)
```
milo-hyben, line 53: as well here?

db/python/layers/analysis.py
```diff
@@ -48,6 +50,7 @@ def __init__(self, connection: Connection):
 
         self.sampt = SampleTable(connection)
```
milo-hyben, line 51: I am not big fan of abbreviations like this, I know this is a bit of legacy, so no drama :-)
'self.sample_table' is clearer, the same for self.analysis_table, self.output_file_table

db/python/tables/file.py
```diff
@@ -0,0 +1,170 @@
...
+
+from cloudpathlib import AnyPath, GSClient
+from google.auth.credentials import AnonymousCredentials
+from google.cloud.storage import Client
+
+from db.python.tables.base import DbBase
+from models.models.file import FileInternal
+
+
+class FileTable(DbBase):
```
milo-hyben, line 13: Should be renamed to OutpuFileTable

models/models/file.py
```diff
@@ -0,0 +1,174 @@
+from typing import Any, Union
+
+from cloudpathlib import AnyPath, GSPath
+from google.cloud.storage import Client
+from pydantic import BaseModel
+
+from models.base import SMBase, parse_sql_bool
+
+
+class FileInternal(SMBase):
```
milo-hyben, line 10: Maybe OutputFileInternal?

models/models/file.py
```diff
@@ -0,0 +1,174 @@
...
+            else:
+                file_root['output'] = file
+                if 'output' in root:
+                    root['output'].append(file)
+                else:
+                    root['output'] = [file]
+        return root
+
+
+class File(BaseModel):
```
milo-hyben, line 144: OutputFile ?

scripts/20240124_migrate_output_to_file.py
```diff
@@ -0,0 +1,168 @@
...
+                    await get_file_info(path=path, client=client)
+                ))
+                print('Extracted and added.')
+    return files
+
+
+async def insert_files(connection, files):
+    """Insert files"""
+    query = dedent(
+        """INSERT INTO file (path, basename, dirname, nameroot, nameext, file_checksum, size, valid)
```
milo-hyben, line 118: INSERT INTO output_file

test/testbase.py
```diff
@@ -154,6 +154,12 @@ async def setup():
                 )
                 formed_connection.project = cls.project_id
 
+                # # Convert the relative path to an absolute path
```
milo-hyben, line 157: Do you want to keep this commented section?

milo-hyben (commented):
I've left a few comments.

nevoodoo:
> Hey @nevoodoo, I had briefly look at the sql schema and I am not big fan of table name 'file'. :-) Wonder if we should call it e.g. analysis_file, there are already various analysis_* tables.

You make a good point @milo-hyben, I have updated the schema now to reflect a better name :)

api/routes/analysis.py
nevoodoo: You're right, somehow missed this since I've been interacting from the layer level directly! I've also just reviewed all other type annotations for the `Analysis` model to use `Union, Optional` instead of the `|` operator.

db/python/layers/analysis.py
nevoodoo: Corrected!

db/python/layers/analysis.py
nevoodoo: Good suggestion, I've updated all those now :)

db/python/tables/file.py
nevoodoo: Updated now!

scripts/20240124_migrate_output_to_file.py
nevoodoo: Updated!

test/testbase.py
nevoodoo: Was thinking maybe we may need to use this in another test class but I should probably remove it for now!

nevoodoo:
I'd like to add some more tests here, for example:

- What happens if I add a valid filename as a `str` to the `output(s)` fields of the `Analysis`. In theory, it should just present this back to me as a string in the result. But if it is a valid pathname, the current logic would create an `OutputFile` record and as such, the results would be presented back as a `JSON`, and not a `str`

nevoodoo:
> I'd like to add some more tests here, for example:
> 
> * What happens if I add a valid filename as a `str` to the `output(s)` fields of the `Analysis`. In theory, it should just present this back to me as a string in the result. But if it is a valid pathname, the current logic would create an `OutputFile` record and as such, the results would be presented back as a `JSON`, and not a `str`

I was right, this is in fact what happened, all patched now however, and tests added to capture this :)

PR #659, opened 2024-01-24
Implement multiple external IDs for participants and samples

jmarshall:
Adds new `participant_external_id` and `sample_external_id` tables to record additional external IDs for a participant/sample — the table can potentially have several rows for each participant/sample. The `participant` and `sample` table's `external_id` fields are migrated to the new tables and the columns renamed so no code can use them (but not removed, as that would interfere with the table versioning).

Further additions to be finalised as the effects of this ripple across the metamist code base…

Addresses #139.

The representation follows that of assay etc: instead of a simple `external_id: str` field, we now have `external_ids: dict[str, str]`: a dictionary of external_organisation → external_id pairs. One of these is designated the PRIMARY_EXTERNAL_ORG and represented by having the `''` empty string as its key. This corresponds to the previous single external_id.

Upsert methods take a similar `external_ids: dict[str, str | None] | None` optional dictionary.

When the upsert is an insert creating a new participant/sample, `external_ids` must be non-None and contain a primary `''` entry with a non-None external_id value; it may contain other additional pairs, and any pairs with value None are ignored.

When the upsert is an update to an existing participant/sample, `external_ids` may be None indicating no changes are being made to any external ids. Otherwise it contains entries to be updated: None values indicate that the external id for that key is to be deleted, and non-None values provide a new value to update that key to. (The value may not be None for the primary key, `''`.)

TODO: Find somewhere appropriate to document this in the metamist source code!


jmarshall:
The following are source files that either have “sample” in their filename or whose contents match `/external_?id/` case insensitively, together with notes as to whether they need updating for this PR.

(NA = not applicable, false positive match; “lower” indicates that this is a wrapper layer, needed behaviour changes will occur due to code changes in lower layers; “one” is generally an indication that ~1 instance of external_id will need updating but others can remain unchanged.)

```
NA	README.md
NAlower	api/graphql/loaders.py
TODO?	api/graphql/schema.py  # GraphQLSample probably do ext_ids JSON a la Assay
NA	api/routes/assay.py
NA	api/routes/family.py
PROB NA	api/routes/participant.py
NAlower	api/routes/sample.py
DONE	db/project.xml
DONE	db/python/connect.py
NA	db/python/enum_tables/sample_type.py
NA	db/python/layers/assay.py
NA	db/python/layers/family.py
one?	db/python/layers/participant.py
TODO	db/python/layers/sample.py  # Mostly LOWER, but check upsert_sample()
one?	db/python/layers/search.py
NA	db/python/layers/seqr.py
um...	db/python/layers/web.py
NA	db/python/tables/analysis.py
NA	db/python/tables/assay.py
NA	db/python/tables/family.py
NA	db/python/tables/participant.py
DONE	db/python/tables/project.py
DONE	db/python/tables/sample.py 
DONE	db/python/tables/sequencing_group.py
NA	etl/extract/main.py
NA	etl/test/test_etl_load.py
TODO	metamist/api/sample_api.py
NA	metamist/audit/README.md
	metamist/audit/audit_upload_bucket.py
	metamist/audit/generic_auditor.py
	metamist/model/body_get_samples.py
	metamist/model/nested_sample.py
	metamist/model/sample_search_response_data.py
	metamist/model/sample_upsert.py
	metamist/parser/generic_parser.py
	metamist/parser/sample_file_map_parser.py
NA	metamist_infrastructure/README.md
NA	models/models/assay.py  # Has multiple external_ids args
NA	models/models/family.py
NA	models/models/participant.py
TODO	models/models/sample.py  # Probably convert to multiple external_ids args
one?	models/models/search.py  # Maybe update SequencingGroupSearchResponseData
NA	models/models/sequencing_group.py  # Has multiple external_ids args
NA	models/utils/sample_id_format.py
	sample_metadata/api/sample_api.py
	sample_metadata/model/body_get_samples.py
	sample_metadata/model/nested_sample.py
	sample_metadata/model/new_sample.py
	sample_metadata/model/sample_batch_upsert.py
	sample_metadata/model/sample_batch_upsert_body.py
OKAY?	sample_metadata/model/sample_search_response_data.py  # All already plural
	sample_metadata/model/sample_type.py
	sample_metadata/model/sample_update_model.py
	scripts/20230420_sequencinggroupmigration.py
	scripts/arbitrary_sm.py
	scripts/back_populate_sequences.py
	scripts/create_test_subset.py
	scripts/fix_family_ids.py
	scripts/fix_participant_ids.py
	scripts/fix_pbmc_sample_participant.py
	scripts/parse_existing_cohort.py
	scripts/parse_participant_meta.py
	scripts/parse_sample_file_map.py
	scripts/parse_vcgs_manifest.py
	scripts/sync_seqr.py
	test/data/generate_data.py
	test/data/generate_seqr_project_data.py
	test/test_analysis.py
	test/test_assay.py
	test/test_audit_log.py
	test/test_generic_auditor.py
	test/test_get_participants.py
	test/test_graphql.py
	test/test_import_individual_metadata.py
	test/test_metamist.py
	test/test_models.py
	test/test_parse_existing_cohort.py
	test/test_parse_generic_metadata.py
	test/test_parse_ont_sheet.py
	test/test_pedigree.py
	test/test_sample.py
	test/test_search.py
	test/test_sequencing_groups.py
	test/test_update_participant_family.py
	test/test_upsert.py
	test/test_web.py
NA	tob_metadata_harmonisation.py
NA	update_sample_status/main.py  # Entirely commented out
HMMM	web/src/pages/family/FamilyView.tsx  # Displays sample eid
NA	web/src/pages/family/FamilyViewTitle.tsx
NA?	web/src/pages/project/ProjectGrid.tsx
TODO	web/src/pages/sample/SampleView.tsx  # Displays sample eid
NA	web/src/shared/components/AssayInfo.tsx  # Demonstrates assay eid list
OKAY?	web/src/shared/components/Header/Search.tsx  # Maybe line 271 needs changing
```

db/python/tables/participant.py
```diff
@@ -249,11 +323,13 @@ async def get_participants_by_families(
         self, family_ids: list[int]
     ) -> tuple[set[ProjectId], dict[int, list[ParticipantInternal]]]:
         """Get list of participants keyed by families, duplicates results"""
-        _query = """
-            SELECT project, fp.family_id, p.id, p.external_id, p.reported_sex, p.reported_gender, p.karyotype, p.meta
+        _query = f"""
+            SELECT fp.family_id, {self.keys_str}
             FROM participant p
             INNER JOIN family_participant fp ON fp.participant_id = p.id
+            INNER JOIN participant_external_id peid ON p.project = peid.project AND p.id = peid.participant_id
```
illusional, line 330: I don't think you need the project check here, I don't know if it has any project implications though. I guess it opens up can a participant have external IDs outside their project? Sounds a little dangerous

models/models/participant.py
```diff
@@ -101,7 +102,7 @@ class Participant(SMBase):
 
     id: int
     project: ProjectId
-    external_id: str = None
+    external_ids: dict[str, str]
```
illusional, line 105: Can we leave the regular external_id here too (including on the Sample), so we can backwards compatible support putting the default ID here (like we do for graphql schema)?

illusional (commented):
Hey @jmarshall, I had a quick glance, amazing how many places we use external IDs! I know this is still WIP, and no reviewers have been tagged, but could you add a list of business rules in the description, or somewhere in a markdown.

IE: I'm interpreting that there must always be an ID with the name as an empty string, if that's the case, could you add a description for this, and add a participant / sample without a default org ID to make sure it fails.

BTW, looking great!

db/python/tables/participant.py
jmarshall: Yes, `p.id` is universally unique, so just checking that would probably be fine.

The backstory is I put a project column in the participant_external_id table because #139 listed one, and eventually because that makes it easy to wipe a project out of the table in `delete_project_data()`.

I thought I had the primary key as project+participant_id and thus there was some cargo-cult programming going on along these lines: therefore there is an index on project+participant_id, therefore it will be most efficient to do the join on the components of the index.

However actually the primary key is participant_id+name (each participant can only have one external_id of each flavour), and there is a unique constraint on project+name+external_id (each external_id must be unique within its flavour within each project).  (Usually I say “external organisation” rather than “flavour”!)

So this is all a bit artificial.

Maybe the family_participant table points to the right answer… Maybe the participant_external_id table shouldn't actually have a `project` column at all?! It doesn't really need one, as participant_id uniquely points to a particular participant regardless of project. Particular implications would be:

1. Slightly harder to wipe out on a per-project basis in `delete_project_data()`, but again family_participant shows how to do that using a sub-select.
2. That unique constraint would just be on name+external_id, so external_ids for a particular external org would need to be unique universally rather than within a particular project. That's a bit bad in that it reduces project independence, but in practice probably doesn't matter at all…
3. I wouldn't need to do the stupid query in the `to_update` part of `ParticipantTable.update_participant()` to figure out what the project is. :+1:
4. I guess not duplicating `project` down here would mean the table was better normalised. :+1:

jmarshall:
Yep, definitely time for people to start taking a look at this — so thanks for having a quick look.

One of my earlier drafts said “TODO add documentation of what the dict[str,str] means” and I still haven't got around to that. Yep, on the basis that external_ids are already non-optional (have I got that right?), the idea is that the empty-string-named one corresponds to the existing primary default external_id and must always be present. Will do re writing up the rules.

models/models/sample.py
```diff
@@ -15,14 +15,16 @@ class SampleInternal(SMBase):
...
-    external_id: str
+    external_ids: dict[str, str]
     meta: dict
     project: int
     type: str | None
     participant_id: int | None
     active: bool | None
     author: str | None = None
 
+    audit_log_id: int | None = None
```
jmarshall, line 26: This is actually more fallout from PR #752, so the getHistoryOfSample endpoint is likely already broken — `SampleTable.get_history_of_sample()` returns an `audit_log_id` field that its `….from_db()` call will choke on. So theoretically this could be a separate PR! (Or the alternative fix would be to avoid the problem by dropping `audit_log_id` from the fields selected / data returned.)

db/python/tables/participant.py
jmarshall: I prepared the commit of not having a `project` column in these two *_external_id tables, and I think I like it: see this [mult-extids-no-project branch](https://github.com/populationgenomics/metamist/compare/multiple-sample-extids...mult-extids-no-project).

Interested in your thoughts @illusional, on this and on whether having a “external_ids must be unique within each external org” constraint is worthwhile.

jmarshall: I've removed the project checks on all those JOINs, and removed the now superseded `project,external_id` constraints on the participant and sample tables.

Thus it turns out that we were indeed previously enforcing “external ids must be unique within a project”, so (apropos (2) above) we should indeed continue to enforce “external ids for the same org name must be unique **within a project**” — hence it is worth keeping the `project` column in these tables.

(Just for posterity: @illusional Was that your reasoning for disliking removing these `project` columns, or were there other aspects as well?)

db/python/layers/web.py
```diff
@@ -443,7 +450,7 @@ async def get_project_summary(
                 pmodels.append(
                     NestedParticipantInternal(
                         id=p['id'],
-                        external_id=p['external_id'],
+                        external_ids=json.loads(p['external_ids']),
```
illusional, line 453: From database call here?

illusional (commented):
@jmarshall, this is looking awesome! Huge effort catching so many different places 💪 

The testing gives me pretty good confidence that this change is pretty ready. I've left a few sporadic comments, I haven't done a good job of distinguishing nitpicks, and what's important, though hopefully it's mostly clear from context.

db/python/tables/sample.py
```diff
@@ -83,7 +94,7 @@ async def query(
     async def get_sample_by_id(
         self, internal_id: int
     ) -> tuple[ProjectId, SampleInternal]:
-        """Get a Sample by its external_id"""
+        """Get a Sample by its internal_id"""
```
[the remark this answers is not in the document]
jmarshall: It was grep what found it :smile:

db/python/tables/participant.py
jmarshall: Now addressed: the tables retain their `project` columns, and `name` no longer contributes to the uniqueness constraint.

models/models/sample.py
jmarshall: Yep. Now removed again.

db/project.xml
```diff
@@ -1281,4 +1281,109 @@
...
+		</createTable>
+		<addPrimaryKey
+			tableName="participant_external_id"
+			columnNames="participant_id,name"
+			constraintName="PK_PARTICIPANT_EXTERNAL_ID"
+			validate="true"
+		/>
+		<addUniqueConstraint
+			tableName="participant_external_id"
+			columnNames="project,name,external_id"
```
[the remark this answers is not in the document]
jmarshall: Yes, my thinking was initially to enable multiple collaborators to use the same external id, even confusingly to different samples/participants. Overall, probably better to adjust it as you suggest, so external ids are unique (within a project) and the first collaborator effectively gets to name the `name`.

db/project.xml
```diff
@@ -1281,4 +1281,109 @@
...
+					nullable="false"
+					foreignKeyName="FK_SAMPLE_SAMPLE_EXTERNAL_ID"
+					references="sample(id)" />
+			</column>
+			<column name="name" type="VARCHAR(255)" />
+			<column name="external_id" type="VARCHAR(255)">
+				<constraints nullable="false" />
+			</column>
+			<column name="meta" type="LONGTEXT" />
+			<column name="audit_log_id" type="INT">
```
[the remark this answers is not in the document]
jmarshall: Done.

db/project.xml
```diff
@@ -1281,4 +1281,109 @@
...
+			tableName="sample_external_id"
+			columnNames="project,name,external_id"
+			constraintName="UK_SAMPLE_EXTERNAL_ID_UNIQUE_EIDS"
+			validate="true"
+		/>
+
+		<sql>ALTER TABLE participant_external_id ADD SYSTEM VERSIONING;</sql>
+		<sql>ALTER TABLE sample_external_id ADD SYSTEM VERSIONING;</sql>
+
+		<sql>INSERT INTO participant_external_id (project, participant_id, name, external_id)
```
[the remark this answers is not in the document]
jmarshall: Good idea, very tidy. It was massively painful for this MariaDB newb to do, but it's done now!

db/python/tables/sample.py
```diff
@@ -171,30 +184,100 @@ async def insert_sample(
...
+                        'name': name,
+                        'external_id': eid,
+                        'audit_log_id': audit_log_id,
+                    }
+                    for name, eid in to_update.items()
+                ]
+                await self.connection.execute_many(_update_query, _eid_values)
+        else:
+            # "touch" all sample_external_id rows for this sample
+            _sync_query = """
```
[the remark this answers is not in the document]
jmarshall: This was due to the history endpoint (see the various commit messages for details). Now reverted.

db/python/utils.py
```diff
@@ -431,6 +431,11 @@ def get_logger():
     return _logger
 
 
+def from_db_json(text):
+    """Convert DB's JSON text to Python object"""
+    return json.loads(text)
```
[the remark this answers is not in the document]
jmarshall: I checked the [documentation](https://docs.python.org/3.10/library/json.html#json.load) and it turns out that since py3.6 `json.loads()` can handle `bytes` itself! So I don't think we need that bit inside `from_db_json` (can just depend on `json.loads()` to handle it itself), and can simplify some of the pre-existing code by converting it to use `from_db_json` instead.

Where other calling code would be simplified would be if `from_db_json` also handled None:

```python
if text is None:
    return None
```

We wouldn't need that for external_ids, but it wouldn't hurt either. Any opinion on handling that, @illusional?

db/python/layers/web.py
jmarshall: I just followed the existing code in that file, which uses `json.loads` directly.

However — see other json-related reply — I think the way to go will be to change **all** the existing `from_db()` functions to use the new `from_db_json()` as appropriate, which I've got done modulo the question in the other reply.

db/python/utils.py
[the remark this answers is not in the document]
illusional: Fun! I swear we added it because something happened, but if tests pass, I'm happy!

db/python/layers/web.py
illusional: Yep, this sounds good to me!

illusional (approved):
Looking great @jmarshall, pretty stoked with everything here. I think this is looking pretty good to go! I've resolved most of my comments, the remainder I'm indifferent. 

You'll need to add a bumpversion commit (probably minor, as this is backwards compatible), then should be good to go!

db/python/layers/web.py
jmarshall: Excellent. It's now a bit off-topic for this PR, so I've separated it into PR #821, to come back to after we've merged this.

db/python/utils.py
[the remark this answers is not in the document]
jmarshall: Improvements to `from_db_json()` is now a followup PR.

test/test_external_id.py
```diff
@@ -0,0 +1,344 @@
...
+from db.python.utils import NotFoundError
+from models.models import (
+    PRIMARY_EXTERNAL_ORG,
+    ParticipantUpsertInternal,
+    SampleUpsertInternal,
+    SequencingGroupUpsertInternal,
+)
+
+
+class TestParticipant(DbIsolatedTest):
```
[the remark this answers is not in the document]
jmarshall: Yeah, I flip-flopped on this one. Turns out there are already duplicate `TestSample` and `TestBillingGcpDailyTable` classes in other separate test files, so the precedent is there and it seemed slightly tidier this way. :shrug:

models/models/participant.py
```diff
@@ -126,7 +127,7 @@ class ParticipantUpsert(SMBase):
     """External upsert model for participant"""
 
     id: int | OpenApiGenNoneType = None
-    external_id: str | OpenApiGenNoneType = None
+    external_ids: dict[str, str | OpenApiGenNoneType] | OpenApiGenNoneType = None
```
jmarshall, line 130: Drawing this to your attention @illusional — is this an appropriate use of `OpenApiGenNoneType`? Want to express that dict entries can be `"foo": None` to remove an entry from the database.

illusional: Yeah I think this is sweet! We'll find out :)

jmarshall:
Cool, cool… I was expecting to have some more tests newly merged in from **dev** that will need updating for multiple external_ids… 😅

PR #660, opened 2024-01-24
Enhanced Data Transfer Process with Assay File Inclusion

michael-harper:
Previously, when upserting assays the meta field was copied across without copying files present in this assay meta field. This field can contain raw `fastq` files that would be necessary to align during a pipeline run. 

`copy_files_in_dict` is capable of iterating through a dictionary (such as the `assay['meta']` field) and creating new file locations in the test bucket based on file locations present in the assay meta.


vivbak (approved):

michael-harper:
Have run this locally and it is working as expected 🙂 thanks @vivbak !

PR #662, opened 2024-01-25
Replaced json with orjson for large size loads

### LINKED ISSUE #658: Use orjson in place of json
I think switching to [orjson](https://github.com/ijl/orjson#performance) over the `json` library would offer significant performance enhancement when loading data into memory. While the units of measurement are only in `ms`, I believe this might add up over time for big batches of analyses.

To summarise some of the metrics from their README:

This measures serializing the citm_catalog.json fixture, more of a worst
case due to the amount of nesting and newlines, as compact (489KiB) or
pretty (1.1MiB):

| Library    |   compact (ms) |   pretty (ms) |   vs. orjson |
|------------|----------------|---------------|--------------|
| orjson     |           0.59 |          0.71 |          1   |
| ujson      |           2.9  |          3.59 |          5   |
| rapidjson  |           1.81 |          2.8  |          3.9 |
| simplejson |          10.43 |         42.13 |         59.1 |
| json       |           4.16 |         33.42 |         46.9 |

This measures serializing 92MiB of JSON from an `numpy.ndarray` with
dimensions of `(50000, 100)` and `numpy.float64` values:

| Library    | Latency (ms)   | RSS diff (MiB)   | vs. orjson   |
|------------|----------------|------------------|--------------|
| orjson     | 194            | 99               | 1.0          |
| ujson      |                |                  |              |
| rapidjson  | 3,048          | 309              | 15.7         |
| simplejson | 3,023          | 297              | 15.6         |
| json       | 3,133          | 297              | 16.1         |

This measures serializing 100MiB of JSON from an `numpy.ndarray` with
dimensions of `(100000, 100)` and `numpy.int32` values:

| Library    | Latency (ms)   | RSS diff (MiB)   | vs. orjson   |
|------------|----------------|------------------|--------------|
| orjson     | 178            | 115              | 1.0          |
| ujson      |                |                  |              |
| rapidjson  | 1,512          | 551              | 8.5          |
| simplejson | 1,606          | 504              | 9.0          |
| json       | 1,506          | 503              | 8.4          |

This measures serializing 105MiB of JSON from an `numpy.ndarray` with
dimensions of `(100000, 200)` and `numpy.bool` values:

| Library    | Latency (ms)   | RSS diff (MiB)   | vs. orjson   |
|------------|----------------|------------------|--------------|
| orjson     | 157            | 120              | 1.0          |
| ujson      |                |                  |              |
| rapidjson  | 710            | 327              | 4.5          |
| simplejson | 931            | 398              | 5.9          |
| json       | 996            | 400              | 6.3          |

The RSS column measures peak memory
usage during serialization. 

Overall, we would benefit from lower latency, and lower memory usage. Thoughts on this?

nevoodoo:
## Description
This is based off #658 which proposes to switch to `orjson` in place of the standard `json` library, for deserialising values that might be large. This switch should offer a significant cut in loading times, without introducing any code complexity.

## Changes Made
- Replaced `json.loads` calls with `orjson.loads` where deemed appropriate. This replacement has not been done on low-impact calls that deserialise small objects.

## Notes
- We should probably use `orjson` when introducing any new features/fixes that involving deserialising of large objects.

## Closes
- [x] Closes #658

### DIFF
--- a/db/python/layers/web.py
+++ b/db/python/layers/web.py
@@ -1,10 +1,11 @@
 # pylint: disable=too-many-locals, too-many-instance-attributes
 import asyncio
 import itertools
-import json
 import re
 from collections import defaultdict
 from datetime import date
+
+import orjson
 
 from api.utils import group_by
 from db.python.layers.base import BaseLayer
@@ -15,15 +16,10 @@ from db.python.tables.assay import AssayTable
 from db.python.tables.base import DbBase
 from db.python.tables.project import ProjectPermissionsTable
 from db.python.tables.sequencing_group import SequencingGroupTable
-from models.models import (
-    AssayInternal,
-    FamilySimpleInternal,
-    NestedParticipantInternal,
-    NestedSampleInternal,
-    NestedSequencingGroupInternal,
-    SearchItem,
-    parse_sql_bool,
-)
+from models.models import (AssayInternal, FamilySimpleInternal,
+                           NestedParticipantInternal, NestedSampleInternal,
+                           NestedSequencingGroupInternal, SearchItem,
+                           parse_sql_bool)
 from models.models.web import ProjectSummaryInternal, WebProject
 
 
@@ -109,7 +105,7 @@ class WebDb(DbBase):
             AssayInternal(
                 id=seq['id'],
                 type=seq['type'],
-                meta=json.loads(seq['meta']),
+                meta=orjson.loads(seq['meta']),  # pylint: disable=maybe-no-member
                 sample_id=seq['sample_id'],
             )
             for seq in assay_rows
@@ -148,7 +144,7 @@ class WebDb(DbBase):
             sg_id_to_sample_id[sg_id] = row['sample_id']
             sg_by_id[sg_id] = NestedSequencingGroupInternal(
                 id=sg_id,
-                meta=json.loads(row['meta']),
+                meta=orjson.loads(row['meta']),  # pylint: disable=maybe-no-member
                 type=row['type'],
                 technology=row['technology'],
                 platform=row['platform'],
@@ -186,7 +182,7 @@ class WebDb(DbBase):
                 id=s['id'],
                 external_id=s['external_id'],
                 type=s['type'],
-                meta=json.loads(s['meta']) or {},
+                meta=orjson.loads(s['meta']) or {},  # pylint: disable=maybe-no-member
                 created_date=str(sample_id_start_times.get(s['id'], '')),
                 sequencing_groups=sg_models_by_sample_id.get(s['id'], []),
                 non_sequencing_assays=filtered_assay_models_by_sid.get(s['id'], []),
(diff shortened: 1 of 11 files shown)


openapi-templates/api_client.mustache
```diff
@@ -1,6 +1,6 @@
 {{>partial_header}}
 
-import json
+import orjson
```
illusional, line 3: If you introduce it into the generated openapi client, you'll need to add it to the `setup.py` (not just requirements.txt, as that is only for the server).

db/python/layers/web.py
```diff
@@ -15,15 +16,10 @@
...
-from models.models import (
-    AssayInternal,
-    FamilySimpleInternal,
-    NestedParticipantInternal,
-    NestedSampleInternal,
-    NestedSequencingGroupInternal,
-    SearchItem,
-    parse_sql_bool,
-)
+from models.models import (AssayInternal, FamilySimpleInternal,
```
illusional, line 19: This is a fun linting one, soon to be resolved in metamist, but your default isort profile in VSCode is incorrect, in your VSCode settings.json you can add:

```json
    "isort.args": [
        "--profile=black"
    ],
```

We intend to add this to metamist's pyproject.toml soon (to replicate https://github.com/populationgenomics/cpg-python-template-repo/blob/6e2da21e1012e6d2b15ba0c385da579a9b80b550/pyproject.toml#L7), but we were waiting for other big PRs to be finalised before introducing lots of merge conflicts.

illusional (approved):
Awesome stuff! Note my 2 review comments before merging, but broadly supportive of this change. I was going to suggest looking at replacing it for Pydantic (I don't actually know which version we use), but looks like there is some ongoing discussion here: https://github.com/pydantic/pydantic/discussions/6388

So maybe best to leave that for now.

nevoodoo:
> Awesome stuff! Note my 2 review comments before merging, but broadly supportive of this change. I was going to suggest looking at replacing it for Pydantic (I don't actually know which version we use), but looks like there is some ongoing discussion here: [pydantic/pydantic#6388](https://github.com/pydantic/pydantic/discussions/6388)
> 
> So maybe best to leave that for now.

I'm in this rabbit hole now since the discussion leads to `mgspec` which reports to be even quicker than orjson as per their benchmarks. I'll hold off on merging this PR until I've done some more tests to see what the better option is I think. I see Pydantic are also having discussions with the creator of `mgspec` so there's a good chance they may borrow from that library instead. Thoughts?

db/python/layers/web.py
nevoodoo: Reverted and fixed. I assume that replication to metamist will happen along with the new `pre-commit` hooks?

illusional:
I don't feel well versed with the different python libraries, looks like the conversation is pretty fresh so happy to wait - (what version of Pydanctic do we use?)

nevoodoo:
> I don't feel well versed with the different python libraries, looks like the conversation is pretty fresh so happy to wait - (what version of Pydanctic do we use?)

Believe `fastapi==0.85.1` and `pydantic==1.10.13` from circa 2022. Current versions are `fastapi==0.109.0` and `pydantic==2.5.3`

illusional:
Awesome, I tried to bump our fastapi, but there's an issue with the openapi version and the generator we use #643 (I think this issue has more commentary).

Is this PR good to merge if we don't want to touch that side?

nevoodoo:
> Awesome, I tried to bump our fastapi, but there's an issue with the openapi version and the generator we use #643 (I think this issue has more commentary).
> 
> Is this PR good to merge if we don't want to touch that side?

I think this should be good to merge for now. `mgspec` is still relatively new so maybe it's something to be looked at again in the future :)

PR #666, opened 2024-01-26
Functionality: Adding sequencing-type as a parameter to allow ingestion of exome data

michael-harper:
Adding sequencing-type as a parameter for future-proofing when exome data comes along and sequencing groups can be ingested as exomes instead of defaulting to genome.

At the moment, default_sequencing_type is set to genome, this fix would allow default_sequencing_type to be either exome or genome

### DIFF
--- a/scripts/parse_existing_cohort.py
+++ b/scripts/parse_existing_cohort.py
@@ -111,6 +111,7 @@ class ExistingCohortParser(GenericMetadataParser):
         batch_number,
         include_participant_column,
         allow_missing_files,
+        sequencing_type,
     ):
         if include_participant_column:
             participant_column = Columns.PARTICIPANT_COLUMN
@@ -131,6 +132,7 @@ class ExistingCohortParser(GenericMetadataParser):
             assay_meta_map=Columns.sequence_meta_map(),
             batch_number=batch_number,
             allow_extra_files_in_search_path=True,
+            default_sequencing_type=sequencing_type,
         )
 
     def _get_dict_reader(self, file_pointer, delimiter: str):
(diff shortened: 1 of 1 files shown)


scripts/parse_existing_cohort.py
```diff
@@ -210,6 +212,12 @@ def get_existing_external_sequence_ids(self, participant_map: dict[str, dict]):
     '--project',
     help='The metamist project to import manifest into',
 )
+@click.option(
+    '--sequencing-type',
+    type=click.Choice(['genome', 'exome']),
+    required=True,
```
vivbak, line 218: Do we need to set required to be true? Esp since we have a default 'genome' described below

vivbak (changes requested):
Looks good! Take a look at your linting failures & test failures. In this case, you need to update the relevant tests in line with your new change. You can also add new tests to capture this new case.

scripts/parse_existing_cohort.py
michael-harper: Good call!

test/test_parse_existing_cohort.py
```diff
@@ -280,3 +285,113 @@ async def test_get_read_filenames_no_reads_pass(self):
         self.assertIn('No read files found for ', cm.output[0])
 
         self.assertEqual(len(read_filenames), 0)
+
+    @run_as_sync
+    async def test_exome_sequencing_type(self):
```
vivbak, line 290: Maybe either rename or split into two so it's accurate to what we're testing

vivbak (approved):

PR #668, opened 2024-01-29
Bumping to actions to latest versions that use Node 20.x

nevoodoo:
# Description
Node 16.x has reached end of life, and some of the actions are now deprecated. Just bumping them here.

## Changes
Bumped all `actions` to the latest release. 

## Reference
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/


illusional (approved):
Looks great, thanks for the suggestion!

PR #672, opened 2024-01-31
Add script to clear test project metadata from Metamist

michael-harper:
This commit introduces a new script that uses the ProjectApi to delete all data from a specified project. It does not delete the project itself, just the data within it. 

If users run in to permission issues with deleting data from a test project, then this script is available.

### DIFF
--- a/scripts/clear_project_test_metadata.py
+++ b/scripts/clear_project_test_metadata.py
@@ -1,10 +1,22 @@
-from metamist.apis import ProjectApi
 import click
+
+from metamist.apis import ProjectApi
 
 
 @click.command()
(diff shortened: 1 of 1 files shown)


MattWellie:
Screenshot 2024-01-31 at 12 20 57
Just being nosy, but isn't this a Swagger endpoint?

vivbak (changes requested):
@MattWellie, yup but most people including myself don't actually have permissions to use it. 

@michael-harper, can we actually move this to a private repo, like https://github.com/populationgenomics/transfer-private 
(which I know isn't the right place for it, but I dont see a better option), and actually hard-code the test project name?

I think exposing the endpoint like this may be a little too risky for my liking.

PR #676, opened 2024-02-04
Rna ingestion updates

EddieLF:
This updates the parsing module to facilitate the ingestion of RNA data. 

There are also minor updates, renaming some functions, rewriting some docstrings, and replacing instances of `"sequence_<field>"` to `"sequencing_<field>"`, or "sequence" to "assay". 

--- 

Any RNA samples / assays ingested should have four fields added to their sequencing group's meta.
- Sequencing facility - `sequencing_facility` ('VCGS', 'KCCG', etc.)
- Sequencing library - `sequencing_library` ('TSStrmRNA', 'TSStrtRNA', etc.)
- Read end type - `read_end_type` ('paired', 'single')
- Read length - `read_length` (some integer, almost all will be 150 or 151)

Exome samples / assays should also have the sequencing facility and library present in their sequencing group's meta, however this won't be enforced for the time being.

---

### `sample_file_map_parser.py`
- Added the four above fields as optional columns in the sample file map
- Added `click` optional args for those fields to provide default values 
- Added `verbose` option to print or not print the final json that gets ingested

### `generic_metadata_parser.py`
- Added get functions for the four fields which gets values from the columns in the sample file map being parsed, or from the parser defaults if set with the `click` options.
- Updated `get_assays_from_group` function to add required fields to`collapsed_assay_meta`, for exomes and RNA assays.
- Added line that calls `get_sequencing_group_meta_from_assays` to add the assay meta fields to the sequencing group meta.

### `generic_parser.py`
- Added function `get_sequencing_group_meta_from_assays` - checks the `assay.meta` for the four fields and adds them to the `sequencing_group.meta`
- Added get functions for the four fields which gets values from the parser defaults where needed
- Added function `prepare_detail` - uses `tabulate` to print each sample being added, it's sequencing type, and how many assays are being ingested. This will be printed in favour of the json dump when `verbose=false`, e.g.:
```
+---------------+--------------+-------------------+----------+
| Participant   | Sample       | Sequencing Type   |   Assays |
+===============+==============+===================+==========+
|               | <sample-id>  | polyarna          |        1 |
+---------------+--------------+-------------------+----------+
|               | <sample-id2> | totalrna          |        1 |
+---------------+--------------+-------------------+----------+
```
- Rename `group_assays` function to `get_sample_sequencing_groups` - since this function groups all sequencing groups for a sample based on the `get_sequencing_group_key` value, the old name did not really fit.

### test_parse_file_map.py
- Added test for valid RNA file map rows
- Added test for invalid RNA file map rows
- Added test for RNA file map rows where the field values for `facility`, `library`, `end_type`, and `read_length` are not present in the file map, but instead instantiated with the default field value parser options.

### requirements.txt
- Add `tabulate` package requirement (for printing the "detail" view which summarises each sample/participant row being ingested)


metamist/parser/sample_file_map_parser.py
```diff
@@ -77,16 +96,21 @@
 class SampleFileMapParser(GenericMetadataParser):
     """Parser for SampleFileMap"""
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
```
milo-hyben, line 99: Disabling 'too-many-arguments', looks like we might need to do a bit of refactoring. 
Just an idea, feel free to ignore :-)
What if all the Sequencing is wrapped in its own class, so it contains type, technology, platform, facililty, library and all the defaults.

metamist/parser/generic_metadata_parser.py
```diff
@@ -553,7 +608,7 @@ async def get_sequencing_group_meta(
 
         return meta
 
-    async def get_assays_from_group(
+    async def get_assays_from_group(  # pylint: disable=too-many-branches
```
milo-hyben, line 611: I wonder if all the read and checksum can be extracted in a separate function, or maybe even that section if reads_type == 'cram' as a separate fun. 
Just making it a bit easier to read.
As an outsider looks like there is a lot of going on there.

metamist/parser/generic_parser.py
```diff
@@ -662,6 +683,31 @@ def prepare_summary(
...
+        sample_participants = {}
+        for sample in samples:
+            sample_participants[sample.external_sid] = sample.participant.external_pid if sample.participant else None
+        sample_sequencing_groups = {sample.external_sid: sample.sequencing_groups for sample in samples}
+
+        details = []
+        for sample, participant in sample_participants.items():
+            for sg in sample_sequencing_groups[sample]:
+                sg_details = {
+                    'Participant': '',
```
milo-hyben, line 697: What about:
'Participant': participant if participant else ''
Then you can remove that if condition afterwards.

milo-hyben (commented):
Here are a few comments, no biggie, feel free to ignore them :-)

cassimons (commented):
The intent of the changes all looks good to me.

metamist/parser/generic_parser.py
```diff
@@ -682,31 +728,33 @@ def prepare_message(
 
         assays_count: dict[str, int] = defaultdict(int)
         sequencing_group_counts: dict[str, int] = defaultdict(int)
-        for s in assays:
-            assays_count[str(s.assay_type)] += 1
+        for a in assays:
+            assays_count[str(a.meta.get('sequencing_type'))] += 1
```
illusional, line 732: Can you we do this by both the assay_type and sequencing_type, even though the former will just be 'sequencing' for now.

requirements.txt
```diff
@@ -22,3 +22,4 @@ SQLAlchemy==1.4.41
 cryptography>=41.0.0
 python-dateutil==2.8.2
 slack-sdk==3.20.2
+tabulate==0.9.0
```
illusional, line 25: These are the server requirements, can you add tabulate to the `setup.py` requirements instead?

illusional (approved):
Thanks @EddieLF, this is a super well structured change! I really appreciate the comprehensive pull-request text, additions of text, and cleanup to some old variable names - makes this a breeze to review :)

Broadly supportive of this change, but as Milo notes in his review, if we get more requirements, we might want to consider fracturing these parsers into different entrypoints, maybe one for each sequencing_type so we can more closely tailor the information required. 

Though, this likely pulls on the larger thread of the lack of a specific ontology or enforced structure for each class of data we store in metamist.

metamist/parser/sample_file_map_parser.py
EddieLF: @milo-hyben thanks for the suggestion, sorry to take a few weeks to come back to this. 

I added a class in `generic_parser.py` called `DefaultSequencing`, which holds `seq_type`, `technology`, `platform`, `facility`, and `library`, and I updated the parsing modules and tests to use this class when entering these sequencing fields.

metamist/parser/generic_metadata_parser.py
EddieLF: Cheers, I split this out into a function `get_read_and_ref_files_and_checksums` which parses the read filenames, checksums, and reference assemblies; and another separate new function `parse_cram_assays` which parses any CRAM assays and links the reference assemblies to the secondary files. No code changes, just splitting the large function into a few smaller ones.

requirements.txt
EddieLF: Cheers, removed from requirements.txt and added to setup.py instead

metamist/parser/generic_parser.py
EddieLF: Added the count of `assay.assay_type` back in in addition to the `assay.meta.sequencing_type` count.

milo-hyben (approved):

EddieLF:
Merging this into `dev` now. I'll see about adding a bumpversion commit and triggering a release next week, don't want to cause any issues on Friday afternoon. Cheers

PR #679, opened 2024-02-08
Fix Existing Cohort Parser Sequencing Type Param

vivbak:
When running the EC parser without a default sequencing type specified the following error was produced 
```
   File "/Users/.../lib/python3.11/site-packages/metamist/parser/generic_metadata_parser.py", line 200, in get_sequencing_type
    value = value.lower()
            ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'lower'
```

Despite the default value being set in the function declaration. It appears that Click was overwriting this value and setting it to None. 

This PR sets the default value within the click param, resolving the issue.

### DIFF
--- a/scripts/parse_existing_cohort.py
+++ b/scripts/parse_existing_cohort.py
@@ -240,7 +240,7 @@ async def main(
     project: str,
     search_locations: List[str],
     batch_number: Optional[str],
-    sequencing_type:str,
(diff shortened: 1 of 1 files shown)


michael-harper (approved):
Interesting it gets set to `None` when not provided despite having a default value in the function declaration. Will have to remember that

PR #680, opened 2024-02-09
Fix pid sgid map api

EddieLF:
This endpoint was broken because it was using the sample ID formatter for the SGID ints that the table query returns, providing incorrect sample IDs. Using the sg_id formatter instead fixes the endpoint, it has been renamed as such. 

I've also added the option to filter SGs of a particular sequencing type, leaving this blank will return all SGs.


illusional (approved):
Thanks @EddieLF, you'll need to add a bumpversion commit (make it a minor, despite this technically being a breaking change).

milo-hyben (approved):

PR #678, opened 2024-02-07
Support for ETL parser configurations

illusional:
Improve the configuration options for the parsers in the ETL flow. Effectively allow:

- Lock specific parsers to specific users
- Default options that always get provided and never overridden
- A way to override the parser (from the request.type / url param)

Configured by: https://github.com/populationgenomics/cpg-infrastructure/pull/179

PR #682, opened 2024-02-13
Fix role description

illusional:
I hate this deploys on failure :/
Introduced in #678


milo-hyben (approved):
Looks like linting isort is not playing nicely :-)

PR #686, opened 2024-02-19
Graphql mutations

nevoodoo:
## Description
As a means to deprecate the use of the Open API Generator, we're attempting to implement GraphQL mutations. Implementation of this would help begin the process and we can start looking at other options (one mentioned below). This change will also allow us to update our Pydantic package to the newest release, from its ~2021 release atm.

## Changes Made
- TODO

## Notes
- Given our `metamist` client is currently being used and is generated by Open API Generator, we would need to consider an alternative that replaces this functionality, while using GraphQL as an API layer between. One current option that seems somewhat stable, and is backed by Mirumee Software (who also created [saleor.io](https://saleor.io)) is [Ariadne Codegen](https://github.com/mirumee/ariadne-codegen) which is able to generate async graphql clients from given schema/queries. **CAVEAT**: Ariadne Codegen only consumes SDL query and schema definitions. We would require exporting the schema from strawberry using the [schema export](https://strawberry.rocks/docs/guides/schema-export) functionality, and then consuming that for the codegen when generating the client.
- Another alternative option is [turms](https://jhnnsrs.github.io/turms/docs/usecases/documents/client) which seems to be a relatively recent project and smaller support.
- We will probably have to discuss whether it is feasible to maintain graphql object/input types, alongside the model definitions, as code reuse declines when we have to update the field definitions in multiple places. Currently, reusing object types does not work for mutations and is advised against. According to the GraphQL specification:
> The GraphQL Object type (ObjectTypeDefinition)… is inappropriate for re‐use (as input), because Object types can contain fields that define arguments or contain references to interfaces and unions, neither of which is appropriate for use as an input argument. For this reason, input objects have a separate type in the system."


api/graphql/mutations/sequencing_groups.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @strawberry.mutation
+    async def update_sequencing_group(
+        self,
+        sequencing_group_id: str,
+        sequencing_group: strawberry.scalars.JSON,
+        info: strawberry.types.Info,
+    ) -> bool:
+        """Update the meta fields of a sequencing group"""
+        connection = info.context['connection']
```
illusional, line 22: We'd need to make sure the connection it gets here is a `write` connections, or a projectless connection, and then decide auth later on. Otherwise it won't allow writes correctly

api/graphql/mutations/cohort.py
```diff
@@ -0,0 +1,121 @@
...
+    cohort_template_id_transform_to_raw,
+)
+
+
+@strawberry.type
+class CohortMutations:
+    """Cohort mutations"""
+
+    @strawberry.mutation
+    async def create_cohort_from_criteria(
```
illusional, line 25: Should we call this just create_cohort?

requirements.txt
```diff
@@ -275,14 +292,18 @@ sniffio==1.3.1
 sqlalchemy==2.0.30
     # via databases
 starlette==0.37.2
-    # via fastapi
-strawberry-graphql[fastapi]==0.229.0
+    # via
+    #   fastapi
+    #   strawberry-graphql
+strawberry-graphql[debug-server]==0.235.0
```
illusional, line 298: Should we have both fastapi and debug-server as options, or does one supercede the other?

illusional (commented):
This is looking great, really appreciate that you'd standardised the method names. I've marked a few endpoints to skip, and huge effort with these tests! I wonder if we should split them into files, based on the section? Or at least have separate classes, like `TestGraphqlAnalysisMutations(TestBase): ...` to keep them logically grouped.

I think this is just about ready though, keen to see the CLI generation for mutations, and how it applies to the parser

nevoodoo:
Moved to #966

PR #687, opened 2024-02-20
Liquibase Manual Deploy Action

nevoodoo:
## Description
Instead of running the liquibase deployment manually via shell onto our database, we should have a Cloud Run app that executes the liquibase schema upgrade remotely.

## Changes Made
- Added the `db/deploy` folder that contains all the config for the cloud run app. This will take in the XML content for the schema `changeLog` and a query to dictate whether it is to be executed on `prod` or `dev`. 

## Notes
- Currently untested but the goal is to add a GitHub Action that lets us call this endpoint to trigger a schema update. 

## TODO
- [x] Add Cloud Run config
- [x] Fix/Add GitHub Action workflow to invoke the Cloud Run endpoint


.github/workflows/update_schema.yaml
```diff
@@ -0,0 +1,43 @@
...
+        curl -L https://github.com/liquibase/liquibase/releases/download/v4.26.0/liquibase-4.26.0.zip -o liquibase.zip
+        unzip liquibase.zip -d liquibase
+
+    - name: Download MariaDB JDBC Driver
+      run: |
+        wget https://repo1.maven.org/maven2/org/mariadb/jdbc/mariadb-java-client/3.0.3/mariadb-java-client-3.0.3.jar
+
+    - name: Run Liquibase Migration
+      env:
+        DATABASE_URL: jdbc:mariadb://${{ secrets.DATABASE_URL }}
```
illusional, line 32: I thought our database won't allow connections like this without ssh tunelling, and if it does we should block it.
I had considered cloud run job doing the actual schema migration, and this action triggering that with the required project.xml file.

We should also have some way to target the production / development database.

nevoodoo: Yes! You're totally right. I've updated the `yaml` files and made the `Dockerfile` needed to build the Cloud Run job. The code for the job should be in the python file however. One of the steps in the updated github action is to build and push the docker image to a registry. I haven't explored our docker registry so that is probably I step I would need assistance with so that we are pushing to the right registry.

illusional (commented):
This is looking great - thanks for putting this together @nevoodoo! 

I'm definitely not against this approach of building a docker container for each stage. If we go down this approach, it's probably worth caching all the build (java, liquibase, google cloud, etc), and adding one layer with the changed project.xml file.

Throwing another idea out to just make sure both are evaluated, you could have a cloud run instance that takes in a project.xml file (or the contents) provided by the github action, and the cloud run instance could respond update the schema with the provided file (and environment to update) - reporting back the stdout / stderr.

Again, totally happy with both! If you go with the former, pre-caching the layers saves us a bit of storage (and heaps of build time) :)

nevoodoo:
@illusional I think we can test this out now. I have moved `liquibase` to the `Dockerfile` and ensured the cloud run endpoint runs. Was successfully able to push a schema update via the endpoint locally. Any issues here on should be limited to the workflows only 💯

db/deploy/main.py
```diff
@@ -0,0 +1,76 @@
...
+
+app = FastAPI()
+
+# Setup for Google Cloud clients and logging
+SECRET_CLIENT = secretmanager.SecretManagerServiceClient()
+LOGGING_CLIENT = logging.Client()
+SECRET_PROJECT = 'sample-metadata'
+SECRET_NAME = 'liquibase-schema-updater'
+log_name = 'lb_schema_update_log'
+changelog_file = 'project.xml'
```
illusional, line 17: Can you add a comment here on why keeping this as project.xml is important?

db/deploy/main.py
```diff
@@ -0,0 +1,76 @@
...
+        response = SECRET_CLIENT.access_secret_version(request={'name': secret_path})
+        return json.loads(response.payload.data.decode('UTF-8'))[env]
+    except Exception as e:  # Broad exception for example; refine as needed
+        text = f'Failed to retrieve or parse secrets: {e}'
+        logger.log_text(text, severity='ERROR')
+        raise HTTPException(status_code=500, detail=text) from e
+
+
+@app.post('/execute-liquibase')
+async def execute_liquibase(request: Request, environment: Literal['prod', 'dev'] = Query(default='dev', regex='^(prod|dev)$')):
```
illusional, line 34: Is black skipping formatting this file? I expect different formatting here

db/deploy/main.py
```diff
@@ -0,0 +1,76 @@
...
+        temp_file_path = temp_file.name  # Store file path to use later
+
+    # Clean up the local temporary file
+    credentials = read_db_credentials(env=environment)
+    db_username = credentials['username']
+    db_password = credentials['password']
+    db_hostname = credentials['host']
+    db_name = credentials['dbname']
+
+    remote_file_path = f'{os.path.basename(temp_file_path)}'
```
illusional, line 50:
```suggestion
    remote_file_path = os.path.basename(temp_file_path)
```

db/deploy/main.py
```diff
@@ -0,0 +1,76 @@
...
+
+@app.post('/execute-liquibase')
+async def execute_liquibase(request: Request, environment: Literal['prod', 'dev'] = Query(default='dev', regex='^(prod|dev)$')):
+    """Endpoint to remotely trigger Liquibase commands on a GCP VM using XML content."""
+    xml_content = await request.body()
+
+    # Temporary file creation with XML content
+    with open(changelog_file, 'wb') as temp_file:
+        temp_file.write(xml_content)
+        temp_file_path = temp_file.name  # Store file path to use later
```
illusional, line 41: This could cause some weird issues if multiple requests are made (eg: one to each environment). You could use tempfile to generate a temp directory, even chdir to that directory to read the next bit, eg:

```python
with tempfile.TemporaryDirectory() as tempdir:
    # Specify the file path within the temporary directory
    with contextlib.chdir(tempdir):
        with open('project.xml', 'w+) as f:
            f.write(xml_content)

        # do the subprocess.run in this indentation with the right cwd and tempdir
```

db/deploy/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN wget https://github.com/liquibase/liquibase/releases/download/v${LIQUIBASE_VERSION}/liquibase-${LIQUIBASE_VERSION}.zip \
+    && unzip liquibase-${LIQUIBASE_VERSION}.zip -d /opt/liquibase \
+    && chmod +x /opt/liquibase/liquibase \
+    # Clean up to reduce layer size
+    && rm liquibase-${LIQUIBASE_VERSION}.zip \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Download the MariaDB JDBC driver
+ENV MARIADB_JDBC_VERSION=3.0.3
```
illusional, line 22: Can you bubble up all the versions to the top so they're easy to scan?

illusional (approved):
Looks great to me! Nothing structural, a few nitpicks and some commentary about the temporary file. Awesome work!!

db/deploy/main.py
nevoodoo: Good point! Have made that change, thank you for that snippet :)

db/deploy/main.py
nevoodoo: Doesn't look like it's skipping on my end, nothing in the config should cause this folder to be skipped from what I can see

db/deploy/main.py
nevoodoo: Yep, added now

db/deploy/Dockerfile
nevoodoo: Done!

PR #689, opened 2024-02-22
Updated database backup script

nevoodoo:
## Description
After a recent database upgrade, we have reviewed the process we use to execute the daily backups. This newer version of the program ensures we are not logging the password in plaintext when an error is logged. 

## Changes Made
- Pass the user's password via environment variable to the subprocess.
- Set 770 permissions on the file so that only the owner/group can read/write/execute the script instead of allowing it to everyone.

### DIFF
--- a/db/backup/backup.py
+++ b/db/backup/backup.py
@@ -1,9 +1,10 @@
-#!/usr/bin/python3.7
+#!/usr/bin/python3
 # pylint: disable=broad-exception-caught,broad-exception-raised
 """ Daily back up function for databases within a local
 MariaDB instance """
 
 import json
+import os
 import subprocess
 from datetime import datetime
 from typing import Literal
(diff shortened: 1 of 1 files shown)


db/backup/backup.py
```diff
@@ -66,10 +67,11 @@ def perform_backup():
                 '--backup',
                 f'--target-dir={tmp_dir}/',
                 f'--user={db_username}',
-                f'-p{db_password}',
             ],
             check=True,
             stderr=subprocess.DEVNULL,
+            # pass the password with stdin to avoid it being visible in the process list
+            env={'MYSQL_PWD': db_password, **os.environ},
```
illusional, line 74: 💪

illusional (approved):
Amazing!! Great to see, can you rotate the password as part of deploying this please 🙏

db/backup/backup.py
illusional: (NB: I think the commend here is stale)

nevoodoo:
> Amazing!! Great to see, can you rotate the password as part of deploying this please 🙏

Absolutely can! Will be reflected on Secret Manager shortly.

PR #690, opened 2024-02-23
Changes to /analysis-runner API point

milo-hyben:
This PR contains the following changes:
1. Fixing search by author string
2. Added ar_guid as extra optional query parameter

### DIFF
--- a/db/python/layers/analysis.py
+++ b/db/python/layers/analysis.py
@@ -596,7 +596,7 @@ class AnalysisLayer(BaseLayer):
     async def get_analysis_runner_log(
         self,
         project_ids: list[int] = None,
-        author: str = None,
+        # author: str = None,
         output_dir: str = None,
         ar_guid: str = None,
     ) -> list[AnalysisInternal]:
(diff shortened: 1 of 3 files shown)


milo-hyben:
Hey @illusional , I have made changes to /analysis-runner, please review.
Thank you!

db/python/layers/analysis.py
```diff
@@ -596,10 +598,11 @@ async def get_analysis_runner_log(
         project_ids: list[int] = None,
         author: str = None,
         output_dir: str = None,
+        ar_guid: str = None,
```
illusional, line 601: Do you think we should use those filters here to expand the usability here, though unrelated to your change so probably a bad idea haha

db/python/tables/analysis.py
```diff
@@ -502,14 +503,18 @@ async def get_analysis_runner_log(
             values['project_ids'] = project_ids
 
         if author:
-            wheres.append('audit_log_id = :audit_log_id')
-            values['audit_log_id'] = await self.audit_log_id()
+            wheres.append('author = :author')
+            values['author'] = author
```
illusional, line 507: I don't think we can filter this way for new ar-entries after the audit_log migration.
I think (a bit awkward), but we have to do some lookup on the audit_log for the new ones, and maybe more strongly consider some migration to unify this.

Noting that the UI is actually wrong too, it should look at the `on_behalf_of` column in the audit log.

illusional (commented):
I don't think `author` is fixed by this, and I don't have a clear mechanism for doing it. Should we consider disabling the search ar by author approach for now?

illusional (approved):

PR #692, opened 2024-02-26
Update es-index types for seqr sync API and UI

EddieLF:
This PR updates the seqr syncing layer to allow for input of an ES-Index "Type". This is useful since we are now creating multiple types of ES-Indexes in metamist analysis records, with analysis type `es-index`. Additionally, In a recent seqr update, the [`datasetType` option for SNV (Haplotypecaller) ES-Indexes was updated](https://github.com/populationgenomics/seqr/blob/main/ui/shared/utils/constants.js#L114). It was previously `"VARIANTS"` and has been updated to `"SNV_INDEL"`. 

The `web` API endpoint has been updated to allow entering a string field for ES-Index type. The available options are from those [displayed by the seqr UI](https://github.com/populationgenomics/seqr/blob/main/ui/pages/Project/components/EditDatasetsButton.jsx#L67). `"Haplotypecaller"` for `"SNV_Indel"` indexes, `"SV_Caller"` for `"SV"` indexes, and `"Mitochondria_Caller"` for `"Mito"` indexes.

To find the latest index for each type in Metamist, we can search for analyses and filter by the `"stage"` field in the analysis `meta`. 
```
SNV_INDEX:  "meta": {"stage": "MtToEs"    }
SV_INDEX:   "meta": {"stage": "MtToEsSv"  }
MITO_INDEX: "meta": {"stage": "MtToEsMito"}  # currently not supported
```
This is done by updating the `AnalysisFilter` passed to the analysis layer query in the `seqr.py` layer with the `"stage"` label corresponding to the input ES-Index type. The types are mapped like above using dictionary mappings added to `layers/seqr.py`
```python
DATASET_TYPE_SNV_INDEL_CALLS = 'SNV_INDEL'
DATASET_TYPE_SV_CALLS = 'SV'
DATASET_TYPE_MITO_CALLS = 'MITO'

ES_INDEX_DATASET_TYPES = {
    'Haplotypecaller': DATASET_TYPE_SNV_INDEL_CALLS,
    'SV_Caller': DATASET_TYPE_SV_CALLS,
    'Mitochondria_Caller': DATASET_TYPE_MITO_CALLS,
}

ES_INDEX_STAGES = {
    DATASET_TYPE_SNV_INDEL_CALLS: 'MtToEs',
    DATASET_TYPE_SV_CALLS: 'MtToEsSv',
    DATASET_TYPE_MITO_CALLS: 'MtToEsMito',
}
...
class SeqrLayer(BaseLayer):
  ...
  async def update_es_index(project, connection, es_index_type, sequencing_type, ...):
    ...
    alayer = AnalysisLayer(connection=self.connection)
    es_index_analyses = await alayer.query(
      AnalysisFilter(
         project=GenericFilter(eq=self.connection.project),
         type=GenericFilter(eq='es-index'),
         status=GenericFilter(eq=AnalysisStatus.COMPLETED),
         meta={'sequencing_type': GenericFilter(eq=sequencing_type),
               'stage': GenericFilter(eq=ES_INDEX_STAGES[ES_INDEX_DATASET_TYPES[es_index_type]])},
         )
    )
```

The UI update adds a dropdown selection to the Seqr Sync options, to select from the ES-Index types. It also meant that the `updateStateFromForm` function had to be split into two, `updateStateFromCheckbox` to handle the existing boolean checkboxes, and a new function `updateStateFromDropdown` to handle the new string options in the drop down menu.
image

image

One final change was in `web/package_lock.json`, the Metamist module was updated to 6.8.0. Not sure if this was necessary to change or commit, it was updated by the install step:
```
python regenerate_api.py \
    && pip install .
```

### DIFF
--- a/db/python/layers/seqr.py
+++ b/db/python/layers/seqr.py
@@ -43,6 +43,22 @@ from models.utils.sequencing_group_id_format import (
 
 SEQUENCING_GROUPS_TO_IGNORE = {22735, 22739}
 
+DATASET_TYPE_SNV_INDEL_CALLS = 'SNV_INDEL'
+DATASET_TYPE_SV_CALLS = 'SV'
+DATASET_TYPE_MITO_CALLS = 'MITO'
+
+ES_INDEX_DATASET_TYPES = {
+    'Haplotypecaller': DATASET_TYPE_SNV_INDEL_CALLS,
+    'SV_Caller': DATASET_TYPE_SV_CALLS,
+    'Mitochondria_Caller': DATASET_TYPE_MITO_CALLS,
+}
+
+ES_INDEX_STAGES = {
+    DATASET_TYPE_SNV_INDEL_CALLS: 'MtToEs',
+    DATASET_TYPE_SV_CALLS: 'MtToEsSv',
+    DATASET_TYPE_MITO_CALLS: 'MtToEsMito',
+}
+
 _url_individuals_sync = '/api/project/sa/{projectGuid}/individuals/sync'
 _url_individual_meta_sync = '/api/project/sa/{projectGuid}/individuals_metadata/sync'
 _url_family_sync = '/api/project/sa/{projectGuid}/families/sync'
@@ -114,6 +130,7 @@ class SeqrLayer(BaseLayer):
         sync_individual_metadata: bool = True,
         sync_individuals: bool = True,
         sync_es_index: bool = True,
+        es_index_type: str = 'Haplotypecaller',
         sync_saved_variants: bool = True,
         sync_cram_map: bool = True,
         post_slack_notification: bool = True,
@@ -206,6 +223,7 @@ class SeqrLayer(BaseLayer):
                     self.update_es_index(
                         sequencing_type=sequencing_type,
                         sequencing_group_ids=sequencing_group_ids,
+                        es_index_type=es_index_type,
                         **params,
                     )
                 )
@@ -359,6 +377,7 @@ class SeqrLayer(BaseLayer):
     async def update_es_index(
         self,
         session: aiohttp.ClientSession,
+        es_index_type: str,
         sequencing_type: str,
         project_guid,
         headers,
@@ -398,7 +417,8 @@ class SeqrLayer(BaseLayer):
                 project=GenericFilter(eq=self.connection.project),
                 type=GenericFilter(eq='es-index'),
                 status=GenericFilter(eq=AnalysisStatus.COMPLETED),
-                meta={'sequencing_type': GenericFilter(eq=sequencing_type)},
+                meta={'sequencing_type': GenericFilter(eq=sequencing_type),
+                      'stage': GenericFilter(eq=ES_INDEX_STAGES[ES_INDEX_DATASET_TYPES[es_index_type]])},
             )
         )
 
@@ -448,7 +468,7 @@ class SeqrLayer(BaseLayer):
             req1_url,
             json={
                 'elasticsearchIndex': es_index,
-                'datasetType': 'VARIANTS',
+                'datasetType': ES_INDEX_DATASET_TYPES[es_index_type],
                 'mappingFilePath': fn_path,
                 'ignoreExtraSamplesInCallset': True,
             },
--- a/web/src/pages/project/SeqrSync.tsx
+++ b/web/src/pages/project/SeqrSync.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import _ from 'lodash'
-import { Button, CheckboxProps, Form, Message, Modal } from 'semantic-ui-react'
+import { Button, CheckboxProps, DropdownProps, Form, Message, Modal } from 'semantic-ui-react'
 import { WebApi } from '../../sm-api'
 import MuckTheDuck from '../../shared/components/MuckTheDuck'
 
@@ -14,6 +14,7 @@ interface SeqrSyncFormOptions {
     syncIndividualMetadata?: boolean
     syncIndividuals?: boolean
     syncEsIndex?: boolean
+    esIndexType: string
     syncSavedVariants?: boolean
     syncCramMap?: boolean
     postSlackNotification?: boolean
@@ -30,6 +31,7 @@ const SeqrSync: React.FunctionComponent<SeqrSyncProps> = ({ syncTypes, project }
         syncIndividualMetadata: true,
         syncIndividuals: true,
         syncEsIndex: true,
+        esIndexType: 'Haplotypecaller',
         syncSavedVariants: true,
         syncCramMap: true,
         postSlackNotification: true,
@@ -45,6 +47,7 @@ const SeqrSync: React.FunctionComponent<SeqrSyncProps> = ({ syncTypes, project }
             .syncSeqrProject(
                 seqType,
                 project,
+                syncOptions.esIndexType,
                 syncOptions.syncFamilies,
                 syncOptions.syncIndividualMetadata,
                 syncOptions.syncIndividuals,
@@ -73,8 +76,18 @@ const SeqrSync: React.FunctionComponent<SeqrSyncProps> = ({ syncTypes, project }
             })
     }
 
-    const updateStateFromForm = (e: React.FormEvent<HTMLInputElement>, data: CheckboxProps) =>
-        setSyncOptions({ ...syncOptions, [data.id || '']: data.checked })
+    const updateStateFromCheckbox = (e: React.FormEvent<HTMLInputElement>, data: CheckboxProps) => {
+        const value = data.checked
+        setSyncOptions({ ...syncOptions, [data.id || '']: value })
+    }
+
+    const updateStateFromDropdown = (
+        e: React.SyntheticEvent<HTMLElement, Event>,
+        data: DropdownProps
+    ) => {
+        const value = (data as any).value
+        setSyncOptions({ ...syncOptions, [data.id || '']: value })
+    }
 
     return (
         <>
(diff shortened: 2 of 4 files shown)


cassimons:
Is there a reason we don't include a "sync all" option? The typical use case would be to ensure we are using the most recent index for all variant types so I would have thought doing them all would make sense unless it is harmful in some way?

cassimons (commented):
The intent LGTM thanks Ed. I will leave the technical approval to others.

EddieLF:
@cassimons that's a good suggestion, I've updated the code so that the endpoint expects a list of string ES-Index types to sync. The sync code will loop through each of the input ES-Index types and find the latest index of that type, constructing the POST to seqr to sync that index, before moving on to the next type.

In the UI I've added a dropdown box that allows for multiple selection, with Haplotypecaller selected as default.

e.g. selecting Haplotypecaller and SV_Caller
image

---

How the other types look in the list:
image

db/python/layers/seqr.py
```diff
@@ -392,71 +411,75 @@ async def update_es_index(
         fn_path = os.path.join(SEQR_MAP_LOCATION, filename)
         # pylint: disable=no-member
 
+        with AnyPath(fn_path).open('w+') as f:  # type: ignore
+            f.write('\n'.join(rows_to_write))
```
illusional, line 415: My gut is we should do this _per_ index, or not once for all? I'm less clear on how strict seqr wants this file, but would've thought seqr will try to link all people listed against the index. Or at the very least if this isn't true, can you add a comment RE this.

db/python/layers/seqr.py
```diff
@@ -392,71 +411,75 @@ async def update_es_index(
...
+
         alayer = AnalysisLayer(connection=self.connection)
-        es_index_analyses = await alayer.query(
-            AnalysisFilter(
-                project=GenericFilter(eq=self.connection.project),
-                type=GenericFilter(eq='es-index'),
-                status=GenericFilter(eq=AnalysisStatus.COMPLETED),
-                meta={'sequencing_type': GenericFilter(eq=sequencing_type)},
+        for es_index_type in es_index_types:
+            es_index_analyses = await alayer.query(
```
illusional, line 419: Query these all at once, and not individually in a loop

db/python/layers/seqr.py
```diff
@@ -392,71 +411,75 @@ async def update_es_index(
...
 
-        with AnyPath(fn_path).open('w+') as f:  # type: ignore
-            f.write('\n'.join(rows_to_write))
+            es_index = es_index_analyses[-1].output
 
-        es_index = es_index_analyses[-1].output
+            messages = []
 
-        messages = []
+            if sequencing_group_ids:
```
illusional, line 443: Maybe consider moving these to a separate function, and then you can run them all in parallel using an `asyncio.gather`!

db/python/layers/seqr.py
```diff
@@ -392,71 +411,75 @@ async def update_es_index(
...
-        )
+            es_index_analyses = sorted(
+                es_index_analyses,
+                key=lambda el: el.timestamp_completed,
+            )
 
-        if len(es_index_analyses) == 0:
-            return ['No ES index to synchronise']
+            if len(es_index_analyses) == 0:
+                return ['No ES index to synchronise']
```
illusional, line 437: I think you'll want to factor this out RE my comment above, but this would cause a hard return if an earlier es-index type doesn't have an es-index to sync, rather than skip and report back.

db/python/layers/seqr.py
```diff
@@ -43,6 +43,22 @@
 
 SEQUENCING_GROUPS_TO_IGNORE = {22735, 22739}
 
+DATASET_TYPE_SNV_INDEL_CALLS = 'SNV_INDEL'
+DATASET_TYPE_SV_CALLS = 'SV'
+DATASET_TYPE_MITO_CALLS = 'MITO'
+
+ES_INDEX_DATASET_TYPES = {
```
illusional, line 50: Can you add some commentary about what definitions go where (even consider more descriptive variable names). IE: of these is seqr, one is production-pipelines, and one is UI driven if I'm understanding correctly. If it's one that metamist controls, then our favour is complete upper case

db/python/layers/seqr.py
```diff
@@ -114,6 +130,7 @@ async def sync_dataset(
         sync_individual_metadata: bool = True,
         sync_individuals: bool = True,
         sync_es_index: bool = True,
+        es_index_types: list[str] = None,
```
illusional, line 133: The values must be known to avoid key errors, so using an enum here is best IMO. Then defining it once, and putting it on the API means you can reference it from the UI (and one change here automatically populates the select grid).

web/src/pages/project/SeqrSync.tsx
```diff
@@ -101,38 +116,59 @@ const SeqrSync: React.FunctionComponent<SeqrSyncProps> = ({ syncTypes, project }
...
                             label="Sync families"
                         />
                         <Form.Checkbox
                             id="syncIndividualMetadata"
                             checked={syncOptions.syncIndividualMetadata}
-                            onChange={updateStateFromForm}
+                            onChange={updateStateFromCheckbox}
                             label="Sync individuals metadata"
                         />
                         <Form.Checkbox
```
illusional, line 134: Just considering some UX here, could you indent the esIndexTypes to show they're grouped. Just throwing out these ideas:

- Disabling the syncESIndex checkbox disables the esIndexTypes and shows an empty list
- (less keen on) removing the checkbox and only having the multi-select. Even though it's simpler, I think it actually makes it more annoying to unselect a specific index.

illusional (changes requested):
Hey @EddieLF, thanks for putting this together! I've dropped a few comments in, but broadly try to avoid making calls to the database in a loop (even though the extra coordination does lead to more complexity) and consider locking down the syncTypes to specific enum values. 

Happy to chat more over zoom too, let me know :)

EddieLF:
## Changes to this PR:

### API / DB
- Use `enum` ES index data types, added to `/models/enums/web.py`. Each enum value is what seqr expects in the `datasetType` field of the POST request which updates the ES index.
- Use `asyncio.gather` to execute all POST requests in parallel, along with new function that does the post request: `post_es_index_update`. 
- Further atomize code by moving the "sequencing groups check" code into its own function in `/db/layers/seqr.py`: `check_updated_sequencing_group_ids`. 
- Get all ES indexes for the project / sequencing type with `AnalysisLayer.query`, then filter the complete list of analyses when iterating through the ES index types being synced.
- Add option to sync `GCNV` indexes, which seqr interprets as `datasetType: "SV_WES"`.

### UI / UX
- Add custom CSS `SeqrSync.css` for displaying the ES Index types in a dropdown multi-select component which is slightly indented from the left, which indicates it is a sub-choice of the `elastic-search` checkbox choice.
- Disable the `ES index type` dropdown when the `Sync elastic-search index` option is unchecked - greying out the dropdown box.

image


---

@illusional thanks again for the review, I think I have addressed your comments above. 

To clarify the first comment:
> My gut is we should do this per index, or not once for all? I'm less clear on how strict seqr wants this file, but would've thought seqr will try to link all people listed against the index. Or at the very least if this isn't true, can you add a comment RE this.

My understanding is that this is not true. I think that Seqr is not concerned if there are extra individuals in the mapping file, just that the mapping file contains all the sgids that are in the ES index being synced. If we get all sgids of a sequencing type in the project (using `player.get_external_participant_id_to_internal_sequencing_group_id_map`) and build the map from that, it will cover all cases for syncing the different ES indexes. Which is why it serves to write the pid-sgid map just once for all pid/sgid of that sequencing type in the project. I've left a comment which somewhat explains this.

---

My final concern is around the requests and the final `asyncio.gather` call which adds the request response texts to the `messages` variable that gets returned. Is the one-liner at the end of the `update_es_index` function (`messages.extend(await asyncio.gather(*requests))`) clear and robust enough? I found it difficult to test this locally.

api/routes/analysis.py
```diff
@@ -241,8 +247,9 @@ async def query_analyses(
 @router.get('/analysis-runner', operation_id='getAnalysisRunnerLog')
 async def get_analysis_runner_log(
     project_names: list[str] = Query(None),  # type: ignore
-    author: str = None,
+    # author: str = None, # not implemented yet, uncomment when we do
```
illusional, line 250: This feels like you've merged something bad?

db/python/layers/seqr.py
```diff
@@ -356,9 +367,50 @@ async def sync_individual_metadata(
             f'Uploaded individual metadata for {len(processed_records)} individuals'
         ]
 
+    def check_updated_sequencing_group_ids(self, sequencing_group_ids: set[int], es_index_analyses: list[AnalysisInternal]):
+        """Check if the sequencing group IDs have been updated"""
+        messages = []
+        if sequencing_group_ids:
+            sequencing_groups_in_new_index = set(
+                es_index_analyses[-1].sequencing_group_ids
```
illusional, line 375: I would re-sort this at the top, it's a small operation and less dangerous than presuming it's sorted. (I know your usage is sorted on the way in, but there's no other condition that states this on this function)

illusional (commented):
@EddieLF, not sure how you've done it, but the changes that are merged in here are showing in the diff log. How did you merge? It makes it difficult to properly review

api/routes/analysis.py
EddieLF: Wow something definitely went wrong in that merge commit, did not realise how mangled this branch got. I'll have to look at my terminal history to see what exactly happened. Sorry bout this, I'll fixerup in the morning 🪠

PR #695, opened 2024-03-01
Etl load trials

illusional:
- Fix parser config format
- Pass row as received: [Don't pre-convert to list as a downstream convert in the BBV parser fails]
- Add a fallback to ^ in the generic_parser
- Convert from pkg_resource to importlib.metadata due to deprecation in py3.10 (we use py311 for the functions).


milo-hyben (approved):
Looking good to me!

PR #704, opened 2024-03-14
Update es-index types for seqr sync API and UI

EddieLF:
Re-created this PR after rebasing on top of the latest dev changes following a merge commit that went awry. [Old PR here](https://github.com/populationgenomics/metamist/pull/692).

---

Updates the seqr syncing layer to allow for input of an ES-Index "Type". This is useful since we are now creating multiple types of ES-Indexes in metamist analysis records, with analysis type `es-index`. Additionally, In a recent seqr update, the [`datasetType` option for SNV (Haplotypecaller) ES-Indexes was updated](https://github.com/populationgenomics/seqr/blob/main/ui/shared/utils/constants.js#L114). It was previously `"VARIANTS"` and has been updated to `"SNV_INDEL"`. 

The `web` API endpoint has been updated to allow entering a fixed enum field for ES-Index type. The available enumerations describe the type of index, while the enum values contain the string value expected by seqr in the POST request.

To find the latest index for each type in Metamist, we can search for analyses and filter by the `"stage"` field in the analysis `meta`. 
```
SNV_INDEX:  "meta": {"stage": "MtToEs"    }
SV_INDEX:   "meta": {"stage": "MtToEsSv"  }
GCNV_INDEX: "meta": {"stage": "MtToEsCNV" }
MITO_INDEX: "meta": {"stage": "MtToEsMito"}  # currently not supported
```
This is done by updating the `AnalysisFilter` passed to the analysis layer query in the `seqr.py` layer with the `"stage"` label corresponding to the input ES-Index type. The types are mapped like above using dictionary mappings added to `layers/seqr.py`

The UI update adds a dropdown selection to the Seqr Sync options, to select from the ES-Index types. The dropdown will be disabled when the checkbox above has been unchecked, abd CSS has been used to align the dropdown box nicely under the sync es index checkbox.

image


web/src/pages/project/SeqrSync.css
```diff
@@ -0,0 +1,11 @@
+.checkbox-dropdown-container {
```
illusional, line 1: Would you consider adding this inline on the component - rather than creating a new styles.css.

illusional (approved):
Looks good to me, thanks @EddieLF!

In general, I'd just force push the revised branch, eg:

```bash
git checkout fixed-branch
git checkout -D broken-branch
git checkout -b broken-branch # creates a branch called broken-branch at HEAD of fixed
git push --force
```

web/src/pages/project/SeqrSync.css
EddieLF: Yeah I like it better with inline too

image

PR #711, opened 2024-03-19
Update back_populate_library_type to affect sg meta as well

EddieLF:
Rejigging this script so that you can optionally populate the sequencing group meta with the `facility` and `library_type` fields, alongside the populating of the assay meta. 

### Changes
- Query sequencing groups (of a specified sequencing type) and all their assays for a project using GraphQL, rather than querying all assays for a project with the AssayAPI. 
- Use CLI flag `-u` to set `update_sequencing_groups` to `True`, which will mean sequencing groups are updated as well.
- Move part of main function that iterates through assays into a new function - `check_assay_meta_fields` - which iterates through a sequencing group's assays and picks out the meta fields, returning them to be updated / logged.
- Update references to "sequences" to "assays".

I've tried to maintain the spirit of the script so it can still be used to update assays without affecting sequencing groups. The only loss of functionality is that you now have to specify a sequencing_type filter for the GraphQL query. Meaning to update all the assays in a project, you will have to run the script for each sequencing type. 

Not sure if this is good enough, if we'd prefer to split this out into two scripts I can revert some of these changes and add a new script which specifically works on sequencing groups. Or I could pull in all sequencing types from the sequencing_type enum table and stick them in the query filter if the user does not specify any type in particular.


EddieLF:
With the `--update-sequencing-groups` flag:

```
>python back_populate_library_type.py -p my-project -s exome -u -d
INFO (root 204): Dummy run. Would have updated 9 assays. [{1: {'facility': 'vcgs', 'library_type': 'SSQXTCRE'}}, {2: {'facility': 'vcgs', 'library_type': 'SSQXTCRE'}}, {3: {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {4: {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {5: {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {6: {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {7: {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}, {8: {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}, {9: {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}]
INFO (root 208): Dummy run. Would have updated 6 sequencing groups. [{'SG1': {'facility': 'vcgs', 'library_type': 'SSQXTCRE'}}, {'SG2': {'facility': 'vcgs', 'library_type': 'SSQXTCRE'}}, {'SG3': {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {'SG4': {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}, {'SG5': {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}, {'SG6': {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}]
```

Without the flag:

```
>python back_populate_library_type.py -p my-project -s exome -d
INFO (root 204): Dummy run. Would have updated 9 assays. [{1: {'facility': 'vcgs', 'library_type': 'SSQXTCRE'}}, {2: {'facility': 'vcgs', 'library_type': 'SSQXTCRE'}}, {3: {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {4: {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {5: {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {6: {'facility': 'vcgs', 'library_type': 'SSQXTCREV2'}}, {7: {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}, {8: {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}, {9: {'facility': 'vcgs', 'library_type': 'TwistWES1VCGS1'}}]
```

cassimons:
My brain still glitches on SG vs assay and what meta should be attached to which. Putting the same data in two places seems like an anti-pattern, should we be picking just one and sticking to it (as clearing any data already placed in the other)?

EddieLF:
I agree that having the data in two places is unintuitive... It depends how we want to gather the sequencing groups I guess. In this case, we want to gather exome sequencing groups and separate them by library type / facility. We could do this by writing a query to pull all sequencing groups that have assays of a particular library type / facility in their meta, without changing sequencing group meta at all.  

But there's an argument to be made that this information should live at the sequencing group level, since each sequencing group describes a collection of one or more read pairs (assays), and sequencing groups are the things that are passed around by the pipelines, not their assays. 

By putting it in the sequencing group meta, we could filter like 
`Return all sequencing groups where library type = "SSQXTCRE" and facility = "VCGS"`

If we just keep it at the assay level, it's slightly more awkward to filter
`Return all sequencing groups where the sequencing group's assays have library type = "SSQXTCRE" and facility = "VCGS"`

@MattWellie Since you will be trying to filter sequencing groups for batching the CNV callsets, which of these is preferable? Is it desirable to port those assay meta fields over into sequencing group meta or totally unnecessary? 
Cheers

cassimons:
From what you describe above I agree the SG meta seems to be the logical place for this to live. Is there any reason for us to not just go with that (and strip it from the assay for clarity)?

This is value that is shared for all files in a SG so it seems the SG is the current place as well as making life more covenant.

MattWellie:
My heart says.. I just want to be able to query metamist for all SG IDs generated from a specific capture/library prep

But isn't it also true that a Sample + Library prep = Assay, and one or more assays = SG ID. Is it technically possible that one SG ID could be comprised of reads from multiple libraries?

If (big **if**) that's true, we should annotate Assays with Library preps, and if a SG ID is created entirely from one consistent library, we should annotated the SGID with that as well. I don't think there's a situation where we would mix libraries, especially for exomes, so annotating both the Assay and SGID seem fine?

cassimons:
> Is it technically possible that one SG ID could be comprised of reads from multiple libraries?
No. A sequencing group could contain sequencing files run on multiple days (ie for a top up), but that should always be from the same library. A different library == a different SG.

MattWellie:
Cool, then I'm v happy with SG IDs being annotated with library, that makes the metamist queries much simpler

cassimons (approved):

PR #713, opened 2024-03-20
Use asyncio gather to execute updates in backpopulate library type script

EddieLF:
Speeding this script up by concurrently executing the API update requests rather than sequentially executing them. 

Also fixing the sequencing group update requests by using the `SequencingGroupMetaUpdateModel` model as input to the `update_sequencing_group()` API call.


MattWellie (approved):
looks reasonable

PR #714, opened 2024-03-25
Load frozen metamist python api client from frozen branch

dancoates:
### Context

We've had a lot of problems with the openapi generator that metamist uses. It is currently stopping us from updating pydantic and switching to using graphql to generate our API client. 

We need to continue supporting existing users of the python openapi client, so it is important that the current version of the python client remains usable. 

### Solution

The best workaround for this is to commit the generated python code to a branch of metamist [metamist-legacy-client](https://github.com/populationgenomics/metamist/tree/metamist-legacy-client) and pull that generated code in instead of attempting to generate it.

### Notes/Caveats

Doing this means that we essentially have a code-freeze on any updates to the REST routes that are read by the OpenAPI generator. If necessary we can update the [metamist-legacy-client](https://github.com/populationgenomics/metamist/tree/metamist-legacy-client) branch with new changes but this will be a slow and highly manual process so is best to avoid. 

Any new routes should be implemented in graphql, and if major changes are needed to existing routes, they should be ported over to graphql endpoints instead.

### DIFF
--- a/.github/actions/get_frozen_client/action.yaml
+++ b/.github/actions/get_frozen_client/action.yaml
@@ -1,0 +1,9 @@
+name: Get frozen OpenAPI client files
+
+runs:
+  using: composite
+  steps:
+    - name: "Copy python api files"
+      run: |
+        python scripts/get_frozen_api_files.py f0a0e546c73202e3c90f152d8861448e1f4203f4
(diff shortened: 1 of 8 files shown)


nevoodoo:
Closed due to the merge of #776

PR #718, opened 2024-04-02
Update seqr sync es response

EddieLF:
Small change to the Update ES-Index POST request in the seqr layer. ~Includes a bumpversion for patching this in main.~

The function `post_es_index_update` returns the response.text for each request's `ClientResponse` object. This floods the returned messages with way too much info about the response, including all family and sample GUIDs involved (see the output in the #product-seqr-internal Slack channel).

Instead of the response.text, we will now return the Index ID (string from analysis output field) along with the status code and reason. So we should see messages like `Updated ES index xyzdataset-exome-2024_0331_0203_0rizz: 200 - OK` instead of the massive wall of response text currently returned. 

Also adds the ES Index ID to the `check_updated_sequencing_group_ids` function return, so that we can see what sequencing groups were added to which index.


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.8.0
```
illusional, line 2: FYI You don't need to bumpversion because you're not changing the python API (anything in `routes/` or `metamist/`).

db/python/layers/seqr.py
```diff
@@ -408,7 +408,7 @@ async def post_es_index_update(self, session: aiohttp.ClientSession, url: str, p
             headers=headers,
         )
         resp.raise_for_status()
-        return await resp.text()
+        return f'Updated ES index {post_json["elasticsearchIndex"]}: {resp.status} - {resp.reason}'
```
illusional, line 411: This is fine, but IMO you could be more defensive (using `.get`) to stop the whole thing failing if the key doesn't exist.

illusional (approved):

.bumpversion.cfg
EddieLF: Thanks for clarifying, removed the bumpversion changes

PR #720, opened 2024-04-03
Adds backoff retry to query and query_async

### LINKED ISSUE #700: Wrap Query() calls in a retry wrapper
# Problem

RE:  https://github.com/populationgenomics/metamist/blob/45d22459cb8ef787a877dc7a5e7ecbf7274557e3/metamist/graphql/__init__.py#L140

I've seen many instances where starting the pipeline, registering results in metamist, or other frequent queries have brought down metamist. Part of this is due to the way that these queries are constructed - pull all results for a specific project, then move on to the next project and repeat.

## Pipeline

1. We run the pipeline with a number of input datasets (for seqr_loader pipeline runs this could contain 15-20 separate projects)
2. At startup queries run for each dataset in turn, gathering relevant participants/SG IDs
3. (in future) pull all analysis entries for all SG IDs, and use those as a source of truth for presence of files in GCP

## GATK-SV

For GATK-SV I periodically run [a query script](https://github.com/populationgenomics/production-pipelines-configuration/blob/main/config/gatk-sv/find_called_samples.py) across all datasets, using the analysis db entries to identify which samples require new data to be generated. At the moment I run this locally, and it's also prone to failure.

## AIP

I have a script called [report_hunter](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/helpers/report_hunter.py), which runs as a cron task under the `talos` project in GCP. This queries for all AIP analysis entries in all projects, and creates an index page for analysts to use. [This fails more times than not](https://batch.hail.populationgenomics.org.au/batches/434832/jobs/1).

## Solution?

For AIP I put together a minimal backoff retry wrapper [here](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/dda0c01f2647853674807eb6dee32c62a7ec0adb/reanalysis/utils.py#L72), and all of my problems went away. This is nothing special, just a try: retry catch block, but since then I've seen 0 failures.

Instead of adding this everywhere that uses metamist, can we patch this at source? Maybe something more robust like https://github.com/litl/backoff would be a good solution to wrap the query/async_query methods

MattWellie:
Closes #700 

Adds https://github.com/litl/backoff as a project dependency, and wraps the query and query_async methods in an exponential backoff. Max tries: 3, max wait: 10s. I've included 3 specific exceptions to the catch-retry, instead of a catch-everything behaviour

Not sure of a good test case for this...


illusional (approved):

jmarshall:
In the week that the xz malfeasance has come to light, should we explicitly consider the pros and cons of coding retries ourselves vs gaining another relatively obscure dependency?

MattWellie:
> should we explicitly consider the pros and cons of coding retries ourselves vs gaining another relatively obscure dependency?

Generally I think that's a good idea, but this is a very simple library that contains all the code we need in pure python and zero dependencies outside the core libs. That includes the boilerplate wrapping, alongside much more technical wrapping of async methods. In this instance I'm inclined to think it's near zero-risk for a substantial benefit.

PR #721, opened 2024-04-03
Capture specific environment variables on write

illusional:
No-op change for most people, but capture more provenance information.
I tested this locally and it works.


milo-hyben (approved):
Looking good to me.

PR #729, opened 2024-04-04
Remove max_tries in backoff logic

MattWellie:
Example https://batch.hail.populationgenomics.org.au/batches/440584/jobs/1

The backoff-retry logic is working, but the minimum wait from `backoff.expo` is 0.5 secs based on the logging. This isn't the behaviour I expected. Based on [this](https://github.com/litl/backoff/blob/master/backoff/_wait_gen.py#L8) the minimum wait should be `1 * 2 ** 0` seconds, so... 1.

The current decorator is also breaking after 2 tries, when the max attempts is set at 3... 

This change removes the max_tries, so that the query will be retried until the retry interval hits 10 seconds.


illusional (approved):
Need a bumpversion commit, I'll add this.

EddieLF:
Will 10 seconds be enough? For example, in the linked batch

```
2024-04-04 23:10:59 INFO (root 243): Getting sequencing groups for dataset xyz
2024-04-04 23:15:59 INFO (backoff 105): Backing off query(...) for 0.5s (gql.transport.exceptions.TransportServerError: 504 Server Error: Gateway Timeout for url: https://sample-metadata-api-xxxx.a.run.app/graphql)
2024-04-04 23:21:00 ERROR (backoff 120): Giving up query(...) after 2 tries (gql.transport.exceptions.TransportServerError: 504 Server Error: Gateway Timeout for url: https://sample-metadata-api-xxxx-ts.a.run.app/graphql)
```

This is like 10 minutes between the initial API call and the giving up. I haven't read too deeply into the backoff module but will this be handled fine?

MattWellie:
hmm... the successful queries are immediate, so I'm guessing that the 5 mins there is due to some metamist-side error state causing the 5 minute timeout error to trigger. That seems like a really long timeout itself 🤔  I've seen this happen IRL and it definitely didn't take 5 mins to trigger.

It's very possible that the 10 sec maximum won't be enough for metamist to recover from anything substantially wrong, but its probably ok in those situations. This is supposed to be lubricant for transient errors, not a catch all for metamist really going down

PR #730, opened 2024-04-05
Billing small fix to show hail batches which have no ar_guid available.

milo-hyben:
This PR contains:
1. small fix to show batches where ar_guid is not present. We have a few of them in Feb 2024 and before.
2. Preparation for very large batches, where jobs billing data gets aggregated, to get number of jobs instead of COUNT(jobs), we look at MAX(JobID).

### DIFF
--- a/db/python/tables/bq/billing_ar_batch.py
+++ b/db/python/tables/bq/billing_ar_batch.py
@@ -17,7 +17,7 @@ class BillingArBatchTable(BillingBaseTable):
 
     async def get_batches_by_ar_guid(
         self, ar_guid: str
-    ) -> tuple[datetime, datetime, list[str]]:
+    ) -> tuple[datetime | None, datetime | None, list[str]]:
         """
         Get batches for given ar_guid
         """
@@ -46,15 +46,24 @@ class BillingArBatchTable(BillingBaseTable):
         # return empty list if no record found
         return None, None, []
 
-    async def get_ar_guid_by_batch_id(self, batch_id: str) -> str:
+    async def get_ar_guid_by_batch_id(
+        self, batch_id: str | None
+    ) -> tuple[datetime | None, datetime | None, str | None]:
         """
-        Get ar_guid for given batch_id
+        Get ar_guid for given batch_id,
+        if batch_id is found, but not ar_guid, then return batch_id back
         """
+        if batch_id is None:
+            return None, None, None
+
         _query = f"""
-        SELECT ar_guid
+        SELECT ar_guid,
+            MIN(min_day) as start_day,
+            MAX(max_day) as end_day
         FROM `{self.table_name}`
         WHERE batch_id = @batch_id
-        AND ar_guid IS NOT NULL
+        GROUP BY ar_guid
+        ORDER BY ar_guid DESC
         LIMIT 1;
         """
 
@@ -63,7 +72,14 @@ class BillingArBatchTable(BillingBaseTable):
         ]
         query_job_result = self._execute_query(_query, query_parameters)
         if query_job_result:
-            return query_job_result[0]['ar_guid']
+            ar_guid = query_job_result[0]['ar_guid']
+            start_day = query_job_result[0]['start_day']
+            end_day = query_job_result[0]['end_day'] + timedelta(days=1)
+            if ar_guid:
+                return start_day, end_day, ar_guid
+
+            # if ar_guid not found but batch_id exists
+            return start_day, end_day, batch_id
 
         # return None if no ar_guid found
-        return None
+        return None, None, None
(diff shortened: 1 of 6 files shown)


db/python/tables/bq/billing_ar_batch.py
```diff
@@ -46,15 +46,24 @@ async def get_batches_by_ar_guid(
...
-        SELECT ar_guid
+        SELECT ar_guid,
+            MIN(min_day) as start_day,
+            MAX(max_day) as end_day
         FROM `{self.table_name}`
         WHERE batch_id = @batch_id
-        AND ar_guid IS NOT NULL
+        GROUP BY ar_guid
+        ORDER BY ar_guid DESC
         LIMIT 1;
```
illusional, line 67: Why do an order by then LIMIT, maybe catch multiple, and then handle 0 (None), 1 (Single), many (Exception) to give a contextual exception for debugging?

test/test_bq_billing_ar_batch.py
```diff
@@ -168,7 +168,7 @@ async def test_get_ar_guid_by_batch_id_no_data(self):
         self.bq_result.result.return_value = []
 
         # test get_ar_guid_by_batch_id function
-        ar_guid = await self.table_obj.get_ar_guid_by_batch_id(batch_id)
+        _st, _ed, ar_guid = await self.table_obj.get_ar_guid_by_batch_id(batch_id)
```
illusional, line 171:
```suggestion
        _, _, ar_guid = await self.table_obj.get_ar_guid_by_batch_id(batch_id)
```

test/test_bq_billing_ar_batch.py
```diff
@@ -180,9 +180,15 @@ async def test_get_ar_guid_by_batch_id_one_rec(self):
...
+            {
+                'ar_guid': expected_ar_guid,
+                'start_day': datetime.datetime(2024, 1, 1, 0, 0),
+                'end_day': datetime.datetime(2024, 1, 2, 0, 0),
+            }
+        ]
 
         # test get_ar_guid_by_batch_id function
-        ar_guid = await self.table_obj.get_ar_guid_by_batch_id(batch_id)
+        _st, _ed, ar_guid = await self.table_obj.get_ar_guid_by_batch_id(batch_id)
```
illusional, line 192:
```suggestion
        _, _, ar_guid = await self.table_obj.get_ar_guid_by_batch_id(batch_id)
```

test/test_bq_billing_ar_batch.py
illusional: Or this, but it might confuse people haha
```suggestion
        *_, ar_guid = await self.table_obj.get_ar_guid_by_batch_id(batch_id)
```

web/src/pages/billing/components/BatchGrid.tsx
```diff
@@ -260,7 +260,7 @@ const BatchCard: React.FC<{ item: AnalysisCostRecordBatch }> = ({ item }) => {
 
                     <DisplayRow label="Cost">
                         {formatMoney(item.cost, 4)}{' '}
-                        {item.jobs?.length > 0 && <em>- across {item.jobs.length} job(s)</em>}
+                        {(item?.jobs_cnt || 0) > 0 && <em>- across {item.jobs_cnt} job(s)</em>}
```
illusional, line 263: Oh yep, I may have changed this in a few places ...

illusional (approved):

test/test_bq_billing_ar_batch.py
milo-hyben: Good point, changed to _,_, to not confused people :-)

test/test_bq_billing_ar_batch.py
milo-hyben: As above

db/python/tables/bq/billing_ar_batch.py
milo-hyben: Good point. I have changed the code accordingly.
There are some jobs within batches, where is missing ar_guid, so when querying the materlized view we can get 2 records per one batch, one is valid ar-guid and the other is NULL as ar-guid. The ORDER BY and LIMIT will ignore that NULL and only choose the valid ar-guid if present.
I had removed the limit and check if the second record is not None, which would be a problem.
If second records is None, then this is fine as we have batches where some its job has no ar-guid set.

There are 8 batches like that in db so far:
SELECT batch_id, ar_guid
FROM `billing-admin-290403.billing_aggregate.aggregate_job_details` 
where batch_id in ('432372', '431078', '431923', '431174', '431590', '430639', '432673', '430868')

For instance, job 60 has no ar-guid
https://batch.hail.populationgenomics.org.au/batches/432673/jobs/60

However the rest of jobs of the same batch have it set, e.g. job#1
https://batch.hail.populationgenomics.org.au/batches/432673/jobs/1

It is not necessary an issue/bug we want to report, we can safely ignore it.

If there is more than one valid ar-guid then we have a problem and the new change would throw Exception. 
Atm we do not have records like that in database.

PR #740, opened 2024-04-23
Refactor families + add family participants to graphql

illusional:
I got carried away, but broadly:

- Refactor the family layer + table to support the newer flavour of querying (+ cleanup a few methods)
- Add family participants to the graphql schema
- Remove the type: ignore for the graphql schema, so there's better type checking there.

Will self-review, then assign out soon.


api/graphql/schema.py
```diff
@@ -1,5 +1,3 @@
-# type: ignore
```
illusional, line 1: This is the big one for fixing type suggestions for this file

api/graphql/schema.py
```diff
@@ -144,11 +157,19 @@ async def pedigree(
     async def families(
         self,
         info: Info,
-        root: 'Project',
+        root: 'GraphQLProject',
+        id: GraphQLFilter[int] | None = None,
+        external_id: GraphQLFilter[str] | None = None,
     ) -> list['GraphQLFamily']:
         connection = info.context['connection']
```
illusional, line 164:
```suggestion
        # don't need a data loader here as we're presuming we're not often running 
        # the "families" method for many projects at once. If so, we might need to fix that
        connection = info.context['connection']
```

api/graphql/schema.py
```diff
@@ -700,7 +792,7 @@ class Query:
     """GraphQL Queries"""
 
     @strawberry.field()
-    def enum(self, info: Info) -> GraphQLEnum:
+    def enum(self, info: Info) -> GraphQLEnum:  # type: ignore
```
illusional, line 795: This could probably be an line comment, but GraphQLEnum is a function call, and python type inferencing doesn't like that, but strawberry needs it (because it resolves this at runtime)

db/python/layers/family.py
```diff
@@ -427,61 +155,64 @@ async def update_family(
...
 
-        return rows
+        mapped_rows: list[dict[str, str | int | None]] = []
+        for r in rows:
+            mapped_rows.append(
+                {
+                    'family_id': fmap.get(r.family_id, str(r.family_id)),
+                    'individual_id': pmap.get(r.individual_id, empty_participant_value),
+                    'paternal_id': pmap.get(r.paternal_id, empty_participant_value),
+                    'maternal_id': pmap.get(r.maternal_id, empty_participant_value),
```
illusional, line 205:
```suggestion
                    'individual_id': pmap.get(r.individual_id, r.individual_id) or empty_participant_value,
                    'paternal_id': pmap.get(r.paternal_id, r.paternal_id) or empty_participant_value),
                    'maternal_id': pmap.get(r.maternal_id, r.maternal_id) or empty_participant_value),
```

db/python/tables/family_participant.py
```diff
@@ -233,7 +199,7 @@ async def delete_family_participant_row(self, family_id: int, participant_id: in
 
         _update_before_delete = """
         UPDATE family_participant
-        SET audit_log_id = :audit_log_id
+        set audit_log_id = :audit_log_id
```
illusional, line 202:
```suggestion
        SET audit_log_id = :audit_log_id
```

db/python/layers/family.py
```diff
@@ -570,9 +304,9 @@ async def import_pedigree(
                             reported_sex=row.sex,
                         )
                     )
-                    external_participant_ids_map[
-                        row.individual_id
-                    ] = upserted_participant.id
+                    external_participant_ids_map[row.individual_id] = (
+                        upserted_participant.id
```
EddieLF, line 308: I prefer the old formatting because I thought the upserted_participant.id value was wrapped in a tuple for a minute. NBD though.

EddieLF (commented):
It looks like there's a lot of cleaning up of the db typehints and changes to the graphql filters & loaders - how much do these  reviewed? I can see the tests cover a fair chunk of it, but obviously Codecov has noted there are aspects missing from tests.

Otherwise, this is my understanding: 
The entries in the family_participant table will now be query-able through GraphQL, either as a child of a query for `participant` or `family`. They won't be query-able at the root level, because a family_participant function hasn't been added to `api/graphql/schema.py` under the `Query` class. 

When queried, a FamilyParticipant node will have the affected status as an integer field under `affected` (alongside `notes`, `participant_id`, `family_id`). 

The family layers/tables updates also let you query families and filter the results by family IDs and/or family external IDs. 

Is my understanding correct? And what do you think about the test coverage? 

All in all, this looks pretty good! 👍

EddieLF (approved):
Looks good to me, nice that you improved the test coverage and also updated the participant layer `typing` typehints to align with the other typehint changes ➕ 👍

Issue #743, opened 2024-04-24
Query fails in strange way when query returns nothing 


vivbak:
A query like this 
```
GET_ANALYSES_QUERY = gql(
    """
        query AnalysesQuery($metamist_proj: String!, $analysis_type: String!, $analysis_status: AnalysisStatus!) {
            project(name: $metamist_proj) {
                analyses (active: {eq: true}, type: {eq: $analysis_type}, status: {eq: $analysis_status}) {
                    id
                    type
                    meta
                    output
                    status
                    sequencingGroups {
                        id
                    }
                }
            }
        }
        """,
)
``` 
(As seen in production pipelines) 

Fails, with a strange error 
```
`raise TransportQueryError(
gql.transport.exceptions.TransportQueryError: {'message': '(1064, "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near \')\' at line 3")', 'locations': [{'line': 3, 'column': 5}], 'path': ['project', 'analyses']} ` 
```

The query should return nothing. When I populated the data s.t. the query should return something it stopped failing as above. 

A failure in general is no big deal, but this isn't very graceful and is hard to catch explicitly.

PR #744, opened 2024-04-24
Fix GenericFilter creating conditional with empty list

illusional:
Closes https://github.com/populationgenomics/metamist/issues/743

Adds a specific check to the GenericFilter for the value for `in` being empty, which means the rest of the query should resolve to False. Also allow a handy way to check if this is the case, and then you can just return no results (eg: implemented in the sequencing group query method.


nevoodoo (approved):
Looks good to me!

PR #748, opened 2024-04-30
Fixing custom audience attribute for cloud run service.

milo-hyben:
Background notes:
To be able to have custom domain linked to a cloud run, property 'custom_audiences' needs to be set.
However there is an issue with pulumi and gcp.
After trial and error I have found they don't support this field when cloud function gets created via gcp.cloudfunctionsv2.Function and gcp.cloudfunctionsv2.FunctionServiceConfigArg.

Solution is to create cloud service separately via gcp.cloudrunv2.Service, which has property 'custom_audiences'.
The only ugly bit is cloud function creates cloud run service by default and there seems to be no way to skip it.
So we would have 2 cloud runs linked to a one function, which is not a problem, just unnescessary extra cloud run service.
It would not cost anything extra as both services have cpu set to idle.

if custom_audience_list is present for the function in the etl configuration, it creates a new custom service with postfix '-external'.

### DIFF
--- a/metamist_infrastructure/driver.py
+++ b/metamist_infrastructure/driver.py
@@ -605,19 +605,13 @@ class MetamistInfrastructure(CpgInfrastructurePlugin):
     def _etl_external_function(
         self,
         f_name: str,
-        fxn: gcp.cloudfunctionsv2.Function,
+        docker_image_url: str,
         sa: gcp.serviceaccount.Account,
         custom_audiences: list[str] | None
     ):
-        docker_image_url = pulumi.Output.all(
-            self.config.gcp.region,
-            self.config.metamist.gcp.project,
-            fxn.name,
-        ).apply(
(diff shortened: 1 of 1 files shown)


metamist_infrastructure/driver.py
```diff
@@ -727,6 +775,23 @@ def _etl_function(
...
         )
 
+        if custom_audience_list:
+            # prepare docker image url
+            docker_image_url = pulumi.Output.all(
+                self.config.gcp.region,
+                self.config.metamist.gcp.project,
+                fxn.name,
+            ).apply(
+                lambda args: f"{args[0]}-docker.pkg.dev/{args[1]}/gcf-artifacts/{args[2].replace('-','--')}:latest"
```
illusional, line 785: Why a double `--` here?

illusional (approved):

PR #750, opened 2024-05-01
OurDNA: Dashboard Backend

nevoodoo:
## Description

This PR adds the backend logic that aggregates the data needed for the OurDNA Dashboard UI.

See: https://docs.google.com/spreadsheets/d/1crEwZhFMD-m0sO4TGm-8FcwsFxGhg_IIyPB5yH2O7Q0/edit?usp=sharing

## Changes
- Added initial layer logic at `db/python/layers/ourdna/dashboard.py` that controls the aggregation logic.
- Added initial schema changes to `api/graphql/schema.py` that would offer a new field to query this data from.


api/graphql/schema.py
```diff
@@ -1074,6 +1075,31 @@ async def analysis_runner(
...
+            )
+            project_name_map = {p.name: p.id for p in projects}
+
+        filter_ = SampleFilter(
+            project=(
+                project.to_internal_filter(lambda pname: project_name_map[pname])
+                if project
+                else None
+            ),
+        )
```
illusional, line 1100: `project` is always going to have a value here, so you can skip the following nullable checks. Also, do you want this to be multiple projects, or just a single?

api/graphql/schema.py
```diff
@@ -1074,6 +1075,31 @@ async def analysis_runner(
...
+        return [await OurDnaDashboardLayer(connection).query(filter_)]
```
illusional, line 1101: This reads a bit awkward, do you actually want to return multiple values here?

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __init__(self, connection: Connection):
+        super().__init__(connection)
+
+        # TODO initialize other layers here as needed for aggregation
+        # self.at = AnalysisRunnerTable(connection)
+
+        self.sample_table = SampleTable(connection)
+
+    async def get_collection_to_process_end_time(self, sample: Sample) -> int | None:
```
illusional, line 22: FWIW, if you don't expect this to be nullable, I'd probably not mark it as so. It can give a false illusional it's making a async check.

Also, definitely not expecting these to already be here, but these methods are very test-able. Also consider the cases where the values are not in the expected format.

illusional (commented):
Looking great so far! I don't know how this is recorded in the user stories, but you may want to think carefully about the relationship between samples + participant stats, and make sure any language around it is super clear and we're trying to display what's actually important to someone.

EG: _90% of samples processed within 24 hours_ may be deceiving if we split samples further if it's delivered under 24 hours, hence it may only be 20% of participant's have a sample that was processed in 24 hours.

api/graphql/schema.py
nevoodoo: Mmmm, GraphQL refused to accept anything but an iterable here :(

db/python/layers/ourdna/dashboard.py
nevoodoo: I think where `process_end_time` is not available yet because the sample isn't processed, or it is lost, we wouldn't be able to make this calculation. And so we have two cases here to be handled: not processed/lost OR processed. Thoughts?

illusional: Yeah totally agree, just make sure it's handled as strptime will raise a ValueError.
(I meant marking this as async, not nullable, got my words all mixed up)

api/graphql/schema.py
vivbak: We were thinking the project here would just be 'OurDNA'. Is there a scenario where another project is valid here?

db/python/layers/ourdna/dashboard.py
nevoodoo: Removed the async :) I think the handling should be working fine but will include this check in testing later.

nevoodoo:
@vivbak @illusional I have moved the `ourdna` field to the `GraphQLProject` type, and removed the nullable checks like was suggested, and there is a project permissions check being made by the layer so I have removed it from the schema level for now. Keen to hear what you think, I have tested to make sure the query works as intended locally :)

api/graphql/schema.py
```diff
@@ -205,6 +206,16 @@ async def analysis_runner(
         analysis_runners = await alayer.query(filter_)
         return [GraphQLAnalysisRunner.from_internal(ar) for ar in analysis_runners]
 
+    @strawberry.field
+    async def ourdna_dashboard(
+        self,
+        info: Info,
+        root: 'Project'
+    ) -> list[strawberry.scalars.JSON]:
```
illusional, line 214:
```suggestion
    ) -> strawberry.scalars.JSON:
```

api/graphql/schema.py
```diff
@@ -205,6 +206,16 @@ async def analysis_runner(
...
+        connection = info.context['connection']
+        ourdna_layer = OurDnaDashboardLayer(connection)
+        return [await ourdna_layer.query(SampleFilter(project=root.id), project_id=root.id)]
```
illusional, line 217:
```suggestion
        return await ourdna_layer.query(SampleFilter(project=root.id), project_id=root.id)
```

vivbak:
To Action: 
https://github.com/populationgenomics/metamist/pull/766#discussion_r1604073456

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+        self.sample_table = SampleTable(connection)
+        self.participant_table = ParticipantTable(connection)
+
+    @staticmethod
+    def get_meta_property(sample: Sample, property_name: str) -> Any:
+        """
+        Get a property from the meta field of a sample, accounting for hyphenated property names
+        or underscores in the property name
+        """
+        return sample.meta.get(property_name) or sample.meta.get(property_name.replace('-', '_'))
```
illusional, line 33: Are these lines not formatted to 88, they feel longer?

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+        processing_times_by_site: dict[str, dict[int, int]] = defaultdict(
+            lambda: defaultdict(int)
+        )  # {site: {hour_bucket: count}}
+        total_samples_by_collection_event_name: dict[str, int] = defaultdict(int)
+        samples_lost_after_collection: dict[str, dict[str, Any]] = {}
+        samples_concentration_gt_1ug: dict[str, float] = {}
+        participants_consented_not_collected: list[int] = []
+        participants_signed_not_consented: list[int] = []
+
+        # TODO We should figure out if we need to handle more than a single project_id
```
illusional, line 124: I don't think you need to handle more than a single project here, at least not for now

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+            _process_start_time, '%Y-%m-%d %H:%M:%S'
+        ) - datetime.strptime(_collection_time, '%Y-%m-%d %H:%M:%S')
+
+        return int(time_taken.total_seconds())
+
+    async def query(
+        self,
+        filter_: SampleFilter,
+        check_project_ids: bool = True,
+        project_id: int = None,
```
illusional, line 102: This parameter is 
```suggestion
        project_id: ProjectId = None,
```

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+    ) -> dict:
+        """Get dashboard data"""
+        projects: set[int]
```
illusional, line 105:
```suggestion
        projects: set[ProjectId]
```

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+        participants_consented_not_collected: list[int] = []
+        participants_signed_not_consented: list[int] = []
+
+        # TODO We should figure out if we need to handle more than a single project_id
+        # for project_id in project_ids:
+        participants.extend(
+            await self.participant_table.get_participants(project=project_id)
+        )
+
+        projects, samples_internal = await self.sample_table.query(filter_=filter_)
```
illusional, line 130: Call the SampleLayer query method, and then you don't need to check projects

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+        total_samples_by_collection_event_name: dict[str, int] = defaultdict(int)
+        samples_lost_after_collection: dict[str, dict[str, Any]] = {}
+        samples_concentration_gt_1ug: dict[str, float] = {}
+        participants_consented_not_collected: list[int] = []
+        participants_signed_not_consented: list[int] = []
+
+        # TODO We should figure out if we need to handle more than a single project_id
+        # for project_id in project_ids:
+        participants.extend(
+            await self.participant_table.get_participants(project=project_id)
```
illusional, line 127: - Why do you take a sample filter, and get all participants? Do you just want participants for the fetched samples, or actually want all. Can you add a comment either way.
- (pending the previous dot point): you can run both the participant and sample query together, in an `asyncio.gather`.
- Better to call a participant layer's `query` method, rather than interact with tables if you can avoid

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+        projects, samples_internal = await self.sample_table.query(filter_=filter_)
+        samples = [s.to_external() for s in samples_internal]
+
+        if check_project_ids:
+            await self.ptable.check_access_to_project_ids(
+                self.author, projects, readonly=True
+            )
+
+        for participant in participants:
+            samples_for_participant = [
```
illusional, line 139: This is an O(n**2) check, could you consider a `group by` on the participant id to reduce the complexity here

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+                sample for sample in samples if sample.participant_id == participant.id
+            ]
+
+            # Create a new participant object with the samples for this participant
```
illusional, line 143: Do you need to copy, if so why?

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+            # Get the participants that have signed but not consented
+            if participant.meta.get('consent') is None:
+                participants_signed_not_consented.append(participant.id)
+
+        # TODO Add logic here to query and aggregate the data
+
+        # go through all samples and get the meta and do something with it
+        # if we want stats for samples by participant, move this logic up before participant and make use of
+        # the filtered samples to map to each participant
+        for sample in samples:
```
illusional, line 168: Could you move this to a separate function, and return some data structure that contains all these stats? There's a lot of logic in here that IMO could also be broken into separate function.

A small note, there's very little computational difference between these two, so consider that when this gets refactored:

```python
for sample in samples:
    if sample.condition_1:
        list1.append(sample)
    if sample.condition_12:
        list2.append(sample)

# AND
for sample in samples:
    if sample.condition_1:
        list1.append(sample)

for sample in samples:
    if sample.condition_12:
        list2.append(sample)
```

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,269 @@
...
+        )
+
+        collection_to_process_end_time_statistics['max'] = (
+            max(collection_to_process_end_time.values())
+            if collection_to_process_end_time
+            else None
+        )
+
+        return {
+            'collection_to_process_end_time': collection_to_process_end_time,
```
illusional, line 260: Use a data structure here

test/test_ourdna_dashboard.py
```diff
@@ -0,0 +1,397 @@
...
+    @run_as_sync
+    async def setUp(self) -> None:
+        # don't need to await because it's tagged @run_as_sync
+        super().setUp()
+        self.odd = OurDnaDashboardLayer(self.connection)
+        self.sl = SampleLayer(self.connection)
+        self.pl = ParticipantLayer(self.connection)
+
+        participants = await self.pl.upsert_participants(
+            [
```
illusional, line 36: Can you move these static variables to be defined as constants to make it easier to scan the setup function

test/test_ourdna_dashboard.py
```diff
@@ -0,0 +1,397 @@
...
+                    reported_sex=2,
+                    karyotype='XX',
+                    meta={'consent': True, 'field': 3},
+                ),
+            ],
+        )
+
+        self.participants_external_objects = [participant.to_external() for participant in participants]
+
+        samples_data = [
```
illusional, line 60: Why not construct the objects here directly rather than matching keys you need to select for, and instead keep them nested in the `ParticipantUpsert` (under the samples attribute) that will do all this inserting for you, eg: https://github.com/populationgenomics/metamist/blob/392d08191a79fbdeb2ac10f6a36c058111a007dc/test/test_graphql.py#L27-L31

illusional (changes requested):

db/python/layers/participant.py
```diff
@@ -981,3 +981,9 @@ async def update_participant_family(
                 maternal_id=fp_row.maternal_id,
                 affected=fp_row.affected,
             )
+
+    async def get_participants_and_samples_meta_by_project(
+        self, project: ProjectId
+    ) -> list[tuple[int, dict, dict]]:
+        """Get participants who have consented but not collected samples"""
```
illusional, line 988: This docstring doesn't match

models/models/ourdna.py
```diff
@@ -0,0 +1,73 @@
...
+            collection_to_process_end_time_24h=collection_to_process_end_time_24h,
+            processing_times_by_site=processing_times_by_site,
+            total_samples_by_collection_event_name=total_samples_by_collection_event_name,
+            samples_lost_after_collection=samples_lost_after_collection,
+            samples_concentration_gt_1ug=samples_concentration_gt_1ug,
+            participants_consented_not_collected=participants_consented_not_collected,
+            participants_signed_not_consented=participants_signed_not_consented,
+        )
+
+    def to_dict(self) -> dict:
```
illusional, line 59: Can we return the actual object to GraphQL instead of to a generic dict, that gives you much better typing, plus you can use the experimental pydantic operator: https://strawberry.rocks/docs/integrations/pydantic.

That will give you a strongly types graphql query when you run `npm run compile`, and make the tests a bit more concrete

db/python/layers/ourdna/dashboard.py
```diff
@@ -95,129 +101,231 @@ def get_collection_to_process_start_time(self, sample: Sample) -> int | None:
 
         return int(time_taken.total_seconds())
 
+    def fetch_key_from_meta(self, key: str, meta: str | dict[str, Any]) -> bool:
+        """
+        Fetches a key from the given meta, if it exists, and checks if it is not False or None
+        """
+        if isinstance(meta, str):
+            meta = json.loads(meta)
```
illusional, line 109: Does this happen often? Feels performant dangerous to potentially always deserialise a JSON string

db/python/layers/ourdna/dashboard.py
```diff
@@ -95,129 +101,231 @@ def get_collection_to_process_start_time(self, sample: Sample) -> int | None:
...
+        if isinstance(meta, dict):
+            if key in meta:
+                value = meta.get(key)
+                if value is not False and value is not None:
+                    return True
+        return False
```
illusional, line 115: Do you want to throw an exception if the type isn't recognised?

db/python/layers/ourdna/dashboard.py
```diff
@@ -95,129 +101,231 @@ def get_collection_to_process_start_time(self, sample: Sample) -> int | None:
...
+
     async def query(
         self,
         filter_: SampleFilter,
-        check_project_ids: bool = True,
-        project_id: int = None,
-    ) -> dict:
+        project_id: ProjectId = None,
```
illusional, line 120: I think project_id here should be required, because it's used for queries below

models/models/ourdna.py
```diff
@@ -0,0 +1,73 @@
...
+        lambda: defaultdict(int)
+    )
+    total_samples_by_collection_event_name: dict[str, int] = defaultdict(int)
+    samples_lost_after_collection: dict[str, dict[str, Any]] = {}
+    samples_concentration_gt_1ug: dict[str, float] = {}
+    participants_consented_not_collected: list[int] = []
+    participants_signed_not_consented: list[int] = []
+
+    @staticmethod
+    def from_sample(d: dict) -> 'OurDNADashboard':
```
illusional, line 23: I'm not really understanding the method name.
Is this method required, it obscures the names of the variables, where you can construct directly, and let pydantic do that heavy lifting for you. Pydantic will also throw an error if the types aren't right which I think is preferred behaviour.

If there are specific transformations you need to do here (it doesn't look like it), then for sure, but I'd collect with `**kwargs`, and only modify the fields you need modifying. Maybe worth us chatting about this when we catch-up!

test/test_ourdna_dashboard.py
```diff
@@ -39,124 +39,112 @@ async def setUp(self) -> None:
...
+        self.sample_external_objects.extend(
+            sample.to_external() for sample in self.sample_internal_objects
+        )
 
     @run_as_sync
     async def test_get_dashboard(self):
         """Test get_dashboard"""
         sample_filter = SampleFilter(project=GenericFilter(eq=self.project_id))
-        dashboard = await self.odd.query(sample_filter)
+        dashboard = (await self.odd.query(sample_filter)).to_dict()
```
illusional, line 147: Following on from broken conversation above, not running to_dict here means the static type linting is as useful as running the test.

illusional (changes requested):

test/test_ourdna_dashboard.py
```diff
@@ -0,0 +1,412 @@
...
+        dashboard = await self.odd.query(project_id=self.project_id)
+        total_samples_by_collection_event_name = (
+            dashboard.total_samples_by_collection_event_name
+        )
+
+        # Check that total_samples_by_collection_event_name is not empty and is a dict
+        assert total_samples_by_collection_event_name
+        assert isinstance(total_samples_by_collection_event_name, dict)
+
+        sample_tally: dict[str, int] = OrderedDict()
```
illusional, line 277: This is a test, so I don't thing it matters too much, but I think you could simplify this by using a `defaultdict` here, and then using unittests `self.assertDictEqual`, rather than a straight assert.

In tests, use assert when you want to do something with the typing, otherwise generally encourage using the [unittests methods](https://docs.python.org/3/library/unittest.html#assert-methods) as it provides pretty good support for telling you the value, vs differences are.

db/python/layers/ourdna/dashboard.py
```diff
@@ -0,0 +1,371 @@
...
+        return sample.meta.get(property_name) or sample.meta.get(
+            property_name.replace('-', '_')
+        )
+
+    def get_collection_to_process_end_time(self, sample: Sample) -> int | None:
+        """
+        I want to know how long it took between blood collection and sample processing - SAMPLE TABLE
+        @fields: collection-time, process-end-time
+        """
+        _collection_time = self.get_meta_property(
```
illusional, line 42: I think this logic would be become simpler if there's a pre-processing step that pulls out these fields as pre-parsed dates, and then these values can be methods on that class. Pulling various tags from meta continuously, and parsing the dates each time feels a little tedious.

Maybe run a pre-processing step to get:

```python
class SampleProcessMeta:
    collection_time: datetime | None
    process_site: str
    process_start_time: datetime | None
    process_end_time: datetime | None

    @staticmethod
    def try_parse_datestr(datestr):
        try:
            return datetime.strptime(
                datestr, '%Y-%m-%d %H:%M:%S'
            )
        except:
            return None

    @staticmethod
    def from_sample(sample: Sample):
        collection_time_str = get_meta_property(sample=sample, property_name='collection-time')
        try:
        
        # ...
        return SampleProcessMeta(...)

    def get_collection_to_process_end_time_seconds(self) -> int:
        if not self.collection_time or not self.process_end_time:
            return None
        return int(self.process_end_time - self.collection_time)
```

What do you think?

illusional (commented):
Hey @nevoodoo, this is super close! On looking again, I think there's a way to simplify the logic of the layer, and nicely encapsulate that logic.

Could you also switch to using [unittest's assert methods](https://docs.python.org/3/library/unittest.html#assert-methods) in the test, they give better logging of values when something isn't equal - and you can use dicts without requiring them to be ordered (with dictEqual, eg)

test/test_ourdna_dashboard.py
nevoodoo: Great suggestion! I've ported most of the references where I used `assert` except where I need it for typing.

db/python/layers/ourdna/dashboard.py
```diff
@@ -15,93 +15,106 @@
...
-    """Layer for analysis logic"""
+class SampleProcessMeta:
+    """Helper class to encapsulate sample metadata properties and calculations."""
 
-    def __init__(self, connection: Connection):
-        super().__init__(connection)
-
-        self.sample_layer = SampleLayer(connection)
-        self.participant_layer = ParticipantLayer(connection)
+    meta: dict
```
illusional, line 21: Hmm, doesn't this end up as a static variable on the class? It's a different pattern to what we usually do. Can you just make this a regular class that gets constructed on input?

db/python/layers/ourdna/dashboard.py
```diff
@@ -15,93 +15,106 @@
...
-        if _collection_time is None or _process_start_time is None:
-            return None
+    @staticmethod
+    def from_sample(sample: Sample) -> 'SampleProcessMeta':
+        """Create a SampleProcessMeta instance from a Sample instance."""
+        instance = SampleProcessMeta()
+        instance.meta = sample.meta
+        return instance
+
+    @property
```
illusional, line 52: Can use cached_property to reduce boilerplate
```suggestion
    @cached_property
```

db/python/layers/ourdna/dashboard.py
```diff
@@ -15,93 +15,106 @@
...
+            self._collection_to_process_start_time = (
+                self.get_time_difference_in_seconds(
+                    'collection-time', 'process-start-time'
+                )
+            )
+        return self._collection_to_process_start_time
+
+    @property
+    def get_lost_sample_properties(self) -> dict[str, Any]:
+        """Returns the properties to report for a sample that has been lost"""
```
illusional, line 85:
```suggestion
        """Returns the normalised properties to report for a sample that has been lost"""
```

db/python/layers/ourdna/dashboard.py
```diff
@@ -15,93 +15,106 @@
...
-        )
-        _process_start_time = self.get_meta_property(
-            sample=sample, property_name='process-start-time'
-        )
-        if _collection_time is None or _process_start_time is None:
-            return None
+    @staticmethod
+    def from_sample(sample: Sample) -> 'SampleProcessMeta':
+        """Create a SampleProcessMeta instance from a Sample instance."""
+        instance = SampleProcessMeta()
```
illusional, line 48: EG: it should be as simple as this
```suggestion
        instance = SampleProcessMeta(meta=sample.meta)
```

Or just change this whole thing to an init:

```python
def __init__(self, sample: Sample):
    self.meta = sample.meta
```

db/python/layers/ourdna/dashboard.py
```diff
@@ -273,71 +289,27 @@ def process_samples_lost_after_collection(
...
         for sample in samples:
-            time_to_process_start = self.get_collection_to_process_start_time(sample)
-
-            if (
-                time_to_process_start is not None
-                and time_to_process_start > 72 * 60 * 60
-            ):
+            processed_meta = SampleProcessMeta.from_sample(sample)
+            time_to_process_start = processed_meta.collection_to_process_start_time
+            if time_to_process_start and time_to_process_start > 72 * 60 * 60:
```
illusional, line 294: Put this logic into the SampleProcessMeta, as `is_lost`

db/python/layers/ourdna/dashboard.py
```diff
@@ -15,93 +15,106 @@
...
+        if self._collection_to_process_start_time is None:
+            self._collection_to_process_start_time = (
+                self.get_time_difference_in_seconds(
+                    'collection-time', 'process-start-time'
+                )
+            )
+        return self._collection_to_process_start_time
+
+    @property
+    def get_lost_sample_properties(self) -> dict[str, Any]:
```
illusional, line 84: If you store a reference to the sample, you can return a `OurDNALostSample` and get init checking through linting, make your life easy here!

illusional (changes requested):
Hey @nevoodoo, this is really close! Great that more logic is encapsulated in that class. 

Though, you can encode more logic and reduce the amount of key accesses here. I've less direct in earlier reviews, but I'll be more prescriptive, encode those key accesses as actual fields. 

- You're already encoding the field names as strings, simplify the amount of string constants here!
- we're parsing dates every time we want to use it. In the below equations, we're parsing everything twice.
- there's no error handling here for bad date formats (want to add a test for this?)
- You can unit test this class without inserting anything into the database


Here's my direct suggestion:

```python
@staticmethod
def try_parse_datetime(d) -> datetime | None:
    try:
        return datetime.strptime(d, '%Y-%m-%d %H:%M:%S')
    except:
        return None

@cached_property
def process_start_time(self) -> datetime | None:
    return self.try_parse_datetime(self.get_property('process-start-time'))

@cached_property
def processing_time(self) -> tuple[str | None, int | None]:
    """Get processing times and site for a sample."""
    if not (self.process_start_time and self.process_end_time):
        return None
    return (self.process_end_time - self.process_start_time).total_seconds

@cached_property
def processing_time_by_site(self) -> tuple[str | none, int | None]:
    return self.get_property('processing-site'), self.processing_time

illusional (approved):
Amazing, thanks for actioning my annoyingly progressive feedback!

Issue #746, opened 2024-04-29
CohortCriteria is bananas! 


vivbak:
The CohortCriteria model, and indeed other models in metamist, seemingly allow you to pass whatever params you like to it. 

This seems problematic -- one for discussion w/ the team?

MattWellie:
Fav issue title to date 🍌

jmarshall:
Here's the example that led to this:

```diff
-- a/scripts/create_custom_cohort.py
++ b/scripts/create_custom_cohort.py
@@ -23,12 +23,13 @@ def main
     capi = CohortApi()
     cohort_criteria = CohortCriteria(
         projects=projects or [],
         sg_ids_internal=sg_ids_internal or [],
         excluded_sgs_internal=excluded_sg_ids or [],
         sg_technology=sg_technologies or [],
         sg_platform=sg_platforms or [],
         sg_type=sg_types or [],
         sample_type=sample_types or [],
+        fruit=['banana'],
     )
```

Here the `CohortCriteria` object acquired a `fruit` field even though no such field appears in its definition.

And previously we accidentally had `sample_types=sample_types or []`, whereby the `CohortCriteria` silently accepted the misnamed field and its intended `sample_type` field was always left unset/empty.

jmarshall:
Turns out this is a [configuration setting in pydantic](https://docs.pydantic.dev/1.10/usage/model_config/#options) for which the default is to ignore unknown fields — which is a bit surprising as we seemed to be getting the `allow` behaviour, as the fruit was preserved.

PR #752 sets `forbid` mode and works through the consequences.

jmarshall:
That PR is merged now, so our model classes will now insist on parameters being real field names rather than typos.

(…or **most** of them will — see the late-breaking comment on the PR…)

PR #752, opened 2024-05-01
Forbid fields other than those defined by each model class

jmarshall:
This PR activates field strictness, as identified in #746. This may or may not be a good idea, but it identifies a few existing problems hence so far I like it… :smile:

This would have detected the `sample_type`/`sample_types` problem (https://github.com/populationgenomics/metamist/pull/615#issuecomment-2071594382) that took a bit of debugging at the end of custom-cohorts.

~~It identifies where we've left a real-world field that exists in the database out of `SequencingGroupInternal` though apparently we don't use it at present (as the omission went unnoticed)~~[Resolved a different way — see below]. It also identifies that `ProjectSummary.links` was probably never being set.

The second commit's fixes are enough to make the tests pass again. There may possibly be other models needing adjustments that have been missed because they are not exercised by tests…

### DIFF
--- a/db/python/tables/sample.py
+++ b/db/python/tables/sample.py
@@ -49,7 +49,6 @@ class SampleTable(DbBase):
         'active',
         'type',
         'project',
-        'audit_log_id',
     ]
 
     # region GETS
--- a/models/models/web.py
+++ b/models/models/web.py
@@ -117,8 +117,3 @@ class ProjectSummary(SMBase):
     seqr_sync_types: list[str]
 
     links: PagingLinks | None
-
-    class Config:
-        """Config for ProjectSummaryResponse"""
-
-        fields = {'links': '_links'}
--- a/db/python/tables/assay.py
+++ b/db/python/tables/assay.py
@@ -549,6 +549,7 @@ class AssayTable(DbBase):
                 drow['external_ids'] = {drow.pop('name'): external_id}
             else:
                 drow['external_ids'] = {}
+                del drow['name']
 
             sequencing_group_id = drow.pop('sequencing_group_id')
             projects.add(drow.pop('project'))
--- a/db/python/tables/sequencing_group.py
+++ b/db/python/tables/sequencing_group.py
@@ -59,7 +59,6 @@ class SequencingGroupTable(DbBase):
         'sg.technology',
         'sg.platform',
         'sg.meta',
-        'sg.author',
         'sg.archived',
     ]
     common_get_keys_str = ', '.join(common_get_keys)
--- a/models/base.py
+++ b/models/base.py
@@ -1,4 +1,4 @@
-from pydantic.main import BaseModel
+from pydantic.main import BaseModel, Extra
 
 # annotate any external objects that must be instantiated with this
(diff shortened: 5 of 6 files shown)


models/models/web.py
```diff
@@ -117,8 +117,3 @@ class ProjectSummary(SMBase):
     seqr_sync_types: list[str]
 
     links: PagingLinks | None
-
-    class Config:
-        """Config for ProjectSummaryResponse"""
-
-        fields = {'links': '_links'}
```
jmarshall, line 124: This change I am the least sure about. I think this is left over from draft versions of this code — `_links` doesn't appear anywhere else (in metamist) any more.

db/python/tables/sample.py
```diff
@@ -49,7 +49,6 @@ class SampleTable(DbBase):
         'active',
         'type',
         'project',
-        'audit_log_id',
```
jmarshall, line 52: An alternative would be to add an `audit_log_id` field to `SampleInternal`. However most other models don't bother having such a field, so I went with leaving it off and we can add it later in the event we ever want to manipulate it.

jmarshall:
I updated _db/python/tables/sequencing_group.py_ so that `sg.author` is omitted from `SequencingGroupInternal` et al. I checked the three places where `common_get_keys_str` is used:

* two pour the resulting rows into `SequencingGroupInternal.from_db()` so it is obviously correct to omit `sg.author` from the SELECT list if you don't want it in the resulting `SequencingGroupInternal`;
* the other, `query()` is more complex but the change still has no unfortunate effects because `SequencingGroupFilter` does not allow filtering by author and there is no `foo`: `sg.author` in `sql_overrides` — so we can drop `sg.author` from the SELECT list as we can be sure it won't appear in the constructed WHERE conditions.

db/python/tables/sample.py
illusional: Na this is pretty reasonable, I'll need to have a better think about audit logs, I think returning the "last edited author" is a bit silly in a way.

models/models/web.py
illusional: Yeah fun, happy for this change :) we were going to go links, but added page tokens (which is effectively just the sample ID)

illusional (approved):

models/models/web.py
jmarshall: The `links` field should still be there and the code initialising it is still there; this removal just removes the aliasing of it so that it's really called `_links` — which wasn't working because nothing else thinks it's called `_links`. Or something like that — I didn't find the pydantic field aliasing documentation very clear.

But anyway, sounds like this field doesn't matter too much => even better!

jmarshall:
Thanks for reviewing. I'll merge it and start taking advantage of it. And hopefully nothing breaks 🤞

jmarshall:
I forgot to mention one possible issue flying just below the radar: some of the Family and Analysis models (and some model classes over in _api/routes/*.py_) inherit from pydantic's BaseModel instead of our SMBase, so they won't get this SMBase.Config setting to set `forbid` mode.

I'm happy to have just hit most of the model classes here, but… what's up with those classes inheriting directly from BaseModel?

PR #755, opened 2024-05-02
Release

milo-hyben:


illusional (approved):
You'll need to update the cpg-infra pointer to deploy metamist changes right?

milo-hyben:
> You'll need to update the cpg-infra pointer to deploy metamist changes right?

Yes, that is correct.

PR #758, opened 2024-05-03
Rolling back max_delivery_attempts to 5.

milo-hyben:
Rolling back max_delivery_attempts=5, gcp throwing error when deploying:
`The value for max_delivery_attempts is too small. You passed 3 in the request, but the minimum value is 5.`


illusional (approved):
Annoying!!

PR #765, opened 2024-05-07
Bump version: 6.10.1 → 6.10.2

vivbak:
Forgot to bump version


illusional (approved):
FWIW, you don't need a bumpversion here, because you're not changing the schema of the API (the graphql interface / api interface). Note this doesn't include the handling code of the API.

PR #766, opened 2024-05-08
OurDNA: Dashboard Frontend

nevoodoo:
[Link to Figma Mockup](https://www.figma.com/file/EXQC3S37utlGqqglGkctPC/OurDNA-Dashboard?type=design&node-id=1%3A4&mode=design&t=bX0SoD1nWa72ZNTa-1)

https://docs.google.com/presentation/d/1MkRjXH6mOX5qQIUjKOwocAZk1friEf48tquLA3cFEP0/edit#slide=id.g2dd7580e6f0_0_26


illusional (changes requested):
We talked in person about these changes, we'll continue to chat about semantic-ui. Great to loop @milo-hyben in here about the d3 vs charts-js.

Also mentioned about pulling out font changes into a different PR.

web/src/pages/ourdna/OurDnaDashboard.tsx
```diff
@@ -0,0 +1,237 @@
...
+    })
+    const tileSizesColumn = useBreakpointValue({
+        base: 'repeat(2, 1fr)',
+        sm: 'repeat(1, 1fr)',
+        md: 'repeat(1, 1fr)',
+        lg: 'repeat(2, 1fr)',
+    })
+
+    const samplesLostAfterCollections = {}
+    data &&
```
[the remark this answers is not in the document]
dancoates: Alternatively if the `if (!data) return /* empty muck component *` early return mentioned below is hoisted above this then you wouldn't need to check for data.

web/src/pages/ourdna/OurDnaDashboard.tsx
```diff
@@ -0,0 +1,237 @@
...
+        Object.keys(data.project.ourdnaDashboard[0].samples_lost_after_collection).forEach(
+            (key) => {
+                samplesLostAfterCollections[key] =
+                    data.project.ourdnaDashboard[0].samples_lost_after_collection[key]
+                        .time_to_process_start / 3600
+            },
+            []
+        )
+
```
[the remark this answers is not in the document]
illusional: I got badly burnt by this in my own side app with very large lists, it caused a huge amount of object creations and changing to a straight update (or maybe lodash key by) reduced it from minutes to seconds.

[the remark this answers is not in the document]
dancoates: Yeah that's a fair call when you're hitting performance problems, I'd still mutate within the reduce rather than mutating outside of it if the performance isn't good enough with the spread operator.

```ts
const samplesLostAfterCollections = Object
    .entries(data.project.ourdnaDashboard[0].samples_lost_after_collection)
    .reduce((rr: Record<string, number>, [key, sample]) => {
        rr[key] = sample.time_to_process_start / 3600;
        return rr;
    }, {})

```

web/package.json
```diff
@@ -18,11 +19,13 @@
         "@types/react-dom": "^17.0.11",
         "@types/react-router-dom": "^5.3.3",
         "axios": "^0.24.0",
+        "chart.js": "^4.4.2",
```
vivbak, line 22: I think we can remove this now :)

web/package.json
```diff
@@ -18,11 +19,13 @@
...
         "d3": "^7.6.1",
         "graphql": "^16.6.0",
         "lodash": "^4.17.21",
         "prop-types": "^15.8.1",
         "react": "^17.0.2",
+        "react-chartjs-2": "^5.2.0",
```
vivbak, line 28: And this :)

web/src/Routes.tsx
```diff
@@ -2,6 +2,8 @@ import * as React from 'react'
 
 import SwaggerUI from 'swagger-ui-react'
 import { Routes as Switch, Route } from 'react-router-dom'
+import { ChakraProvider } from '@chakra-ui/react'
```
vivbak, line 5: And this :)

web/src/cpg-logo-rgb.svg
vivbak, line 1: This is my bad, but we should rename this to our-dna-logo-rgb.svg

web/src/cpg-white.svg
vivbak, line 1: Similarly

web/src/index.css
```diff
@@ -1,4 +1,5 @@
 @import url('https://fonts.googleapis.com/css2?family=Raleway:wght@400&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Plus+Jakarta+Sans:ital,wght@0,200..800;1,200..800&display=swap');
```
vivbak, line 2: Do we still use this? I think we can scrap :)

web/src/shared/components/OurDNALogo.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import ourDNA from '../../cpg-logo-rgb.svg'
```
vivbak, line 2: We need to be sure to edit here if we change the names :)

web/src/shared/components/ourdna/BarChart.tsx
```diff
@@ -0,0 +1,251 @@
+import * as React from 'react'
+import { Card, Container } from 'semantic-ui-react'
+import * as d3 from 'd3'
```
vivbak, line 3: Goodness D3 is so clunky 😂 
@milo-hyben, just pulling you here for a consult to see if you have any advice on how to make this a little sleeker

web/src/shared/components/ourdna/TableTile.tsx
```diff
@@ -0,0 +1,64 @@
...
+    columns: Array<string>
+    icon: React.ReactNode
+}
+
+const TableTile: React.FC<TableTileProps> = ({ header, data, columns, icon }) => (
+    <Card
+        fluid
+        style={{
+            backgroundColor: 'var(--color-bg-card)',
+            boxShadow: 'rgba(0, 0, 0, 0.24) 0px 3px 8px',
```
vivbak, line 16: just a nit, but do you think we should keep it consistent and pull the shadow as a var defined in the css

web/src/shared/components/ourdna/Tile.tsx
```diff
@@ -0,0 +1,46 @@
...
+    description: string
+    icon: React.ReactNode
+}
+
+const Tile: React.FC<TileProps> = ({ header, stat, units, unitsColour, description, icon }) => (
+    <Card
+        fluid
+        style={{
+            backgroundColor: 'var(--color-bg-card)',
+            boxShadow: 'rgba(0, 0, 0, 0.24) 0px 3px 8px',
```
vivbak, line 18: And here

web/src/shared/components/ourdna/BarChart.tsx
milo-hyben: Yes, there are few ways to use d3. Most React like is implemented in the StackedAreaByDateChart. Then second close is StackedBarChart.
The rest of charts are more of pure JS than React and we should refactor at some stage.
Basically whenever we need to clear content / reseting SVG in the code, it is a bit of smell not using it React way.
Happy to jump on call and explain in details.

web/src/shared/components/ourdna/BarChart.tsx
```diff
@@ -0,0 +1,251 @@
...
+            .attr('font-size', 10)
+            .attr('text-anchor', 'start')
+            .attr('transform', `translate(${dimensions.width - margin.right + 10},${margin.top})`)
+            .selectAll('g')
+            .data(transformedData.map((d) => d.site).filter((v, i, a) => a.indexOf(v) === i))
+            .enter()
+            .append('g')
+            .attr('transform', (d, i) => `translate(0,${i * 20})`)
+
+        // Append X axis
```
dancoates, line 174: Not sure if you saw but d3 has a helper lib for building axes, https://d3js.org/d3-axis. This is fine for now though

dancoates (approved):

web/src/shared/components/ourdna/BarChart.tsx
dancoates: lol, sorry, just realised you're already doing this, ignore me

PR #767, opened 2024-05-08
Metamist Infrastructure Documentation

milo-hyben:
I have added a simple diagram explaining Metamist Infrastructure ETL flow:

https://github.com/populationgenomics/metamist/tree/metamist-etl-docs/metamist_infrastructure


illusional (approved):
Ooft this did sound simpler in my head, but drawn out like this 🤯

PR #771, opened 2024-05-14
Updated family view

illusional:
More information, more easily available, with the pedigree. Very early draft


illusional:
Show pedigree rows to start
image

illusional:
Family view:

image

Participant View:

image

Analysis View: 
image

web/src/pages/family/FamilyView.tsx
```diff
@@ -1,229 +1,359 @@
...
+            id: number
+            timestampCompleted?: any | null
+            type: string
+            sgs: string[]
+            meta?: any | null
+            output?: string | null
+        }
+    } = {}
+
+    for (const fp of data?.family?.familyParticipants) {
```
dancoates, line 131: This probably needs to be coalesced to something iterable, at the moment if data or family don't exist this'll error out

web/src/pages/family/FamilyView.tsx
```diff
@@ -1,229 +1,359 @@
...
-import Pedigree from '../../shared/components/pedigree/Pedigree'
 import LoadingDucks from '../../shared/components/LoadingDucks/LoadingDucks'
 
 import { gql } from '../../__generated__/gql'
-import FamilyViewTitle from './FamilyViewTitle'
 
-import iconStyle from '../../shared/iconStyle'
-import SeqPanel from '../../shared/components/SeqPanel'
-import SampleInfo from '../../shared/components/SampleInfo'
+import _ from 'lodash'
```
dancoates, line 11: would be good to import the individual tools here

web/src/pages/participant/ParticipantViewContainer.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    if (loading) {
+        return <LoadingDucks />
+    }
+    if (error || !data) {
+        return <Message error>{error?.message || 'Participant was not found'}</Message>
+    }
+
+    const participant = data?.participant
+    const analyses = participant.samples.flatMap((s) =>
```
dancoates, line 71: This is a little thing, but it's a bit more readable to write these nested flatmaps as:

```ts
participant.samples
    .flatMap(s => s.sequencingGroups)
    .flatMap(sg => sg.analyses)
    .flatMap(an => ({
        ...an,
        sgs: an.sequencingGroups.map(sg => sg.id)
    }))
```

rather than calling flatMap in the predicate of each function

web/src/index.css
```diff
@@ -490,3 +498,7 @@ html[data-theme='dark-mode'] .ui.table {
 .ui.attached.inverted.menu {
     border: 1px solid #555 !important;
 }
+
+.dimmed.dimmable > .ui.modals.dimmer.visible {
+    display: flex !important;
```
dancoates, line 503: Not sure I approve of this haha, but also not too fussed as it seems like we've been fairly liberal with !important in the past

web/src/shared/components/pedigree/TangledTree.tsx
```diff
@@ -1,29 +1,774 @@
 /* Inspired by https://observablehq.com/@nitaku/tangled-tree-visualization-ii */
```
dancoates, line 1: I haven't reviewed this file in too much detail but happy for it to go in as-is. My feeling is that it could probably be simplified by moving away from SVG and relying on the DOM layout engine to organize the nodes but I also suspect I'm underestimating the complexity. Will have a bit of a better look at it at a later date.

dancoates (commented):
Pretty happy with this generally, have made a few very minor comments, but generally happy to approve. Will have another look once you've made the changes you wanted.

illusional:
Thanks @dancoates, I've:

- applied your suggestions
- added a quick route for an analysis page, 
- updated a few places to use a link instead of an anchor. Seems to work okay from my local testing!
- Used word-breaks to force a wrapping to make the views appear better (including replacing `<pre>` with a `<span style="font-family: monospace">`)

dancoates (approved):

PR #773, opened 2024-05-14
Release

EddieLF:
No bumpversion needed because no API changes right?


illusional (approved):
Yep, no bumpversion required

PR #774, opened 2024-05-14
Fix individual metadata for seqr layer headers

EddieLF:
OK really silly bug that I missed when I changed the code to use walrus at the last minute before submitting my previous PR.

Basically we need an extra set of parentheses to define the `header` variable in the loop, or else the `headers` list is just full of `True`.

```python
list1 = ['a','b','c']
list2 = ['a','c','d']

# header defined as "item in list2"
# if condition evaluates `header=True`
for item in list1:
  if header := item in list2:
    print(header)
>True
>True

# header defined as "item"
# if condition evaluates `header in list2`
for item in list1:
  if (header := item) in list2:
    print(header)
>a
>c
```

### DIFF
--- a/db/python/layers/participant.py
+++ b/db/python/layers/participant.py
@@ -767,7 +767,8 @@ class ParticipantLayer(BaseLayer):
         rows = [{h: r.get(h) for h in set_headers if h in r} for r in rows]
         headers = []  # get ordered headers if we have data for it
         for h in SeqrMetadataKeys.get_ordered_headers():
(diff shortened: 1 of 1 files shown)


illusional (approved):
Hmmm, I think this starts to border on confusing Python to intepret. I think clarity wise it would be better to have it in different statements, thoughts?

EddieLF:
@illusional sure, I've split the header definition in the loop onto an extra line for clarity.

PR #777, opened 2024-05-15
Add pseudo contributor role w/o readonly reqs, but write access

illusional:
Short term hack for: https://github.com/populationgenomics/metamist/pull/703

1. Check readonly
2. Then update the connection to make it non-readonly. 

Though this gently points out that someone could update the connection.readonly attribute to False anywhere in the code, maybe this should be frozen.

### DIFF
--- a/api/utils/db.py
+++ b/api/utils/db.py
@@ -120,7 +120,7 @@ async def HACK_dependable_contributor_project_connection(
     connection = await ProjectPermissionsTable.get_project_connection(
         project_name=project,
         author=author,
-        readonly=False,
+        readonly=True,
         on_behalf_of=None,
         ar_guid=ar_guid,
         meta=meta,
--- a/api/utils/db.py
+++ b/api/utils/db.py
@@ -114,7 +114,7 @@ async def HACK_dependable_contributor_project_connection(
     if extra_values:
         meta.update(extra_values)
 
-    meta = {'role': 'contributor'}
+    meta['role'] = 'contributor'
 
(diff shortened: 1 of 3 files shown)


api/utils/db.py
```diff
@@ -103,6 +103,35 @@ async def dependable_get_write_project_connection(
...
+    author: str = Depends(authenticate),
+    ar_guid: str = Depends(get_ar_guid),
+    extra_values: dict | None = Depends(get_extra_audit_log_values),
+) -> Connection:
+    """FastAPI handler for getting connection WITH project"""
+    meta = {}
+    if extra_values:
+        meta.update(extra_values)
+
+    meta = {'role': 'contributor'}
```
jmarshall, line 117: Probably wants to go above the `meta.update(extra_values)` bit?

illusional: I didn't want any other values to potentially update it?

jmarshall: Fair, but then it would be better as

```
meta['role'] ='contributor'
```

At the moment, it wipes out anything passed in as an argument…

illusional: Oh wow, I'm losing my marbles two ways 😓

dancoates (approved):
Looks good, just a weird unrelated lint error that needs fixing by the looks of it

jmarshall (approved):
Weird that the linter didn't complain about `has_participant_join` in the PR in which it was recently added, but :shrug: …

PR #778, opened 2024-05-15
Release

illusional:
- Fix individual metadata for seqr layer headers (#774)
- Add pseudo contributor role w/o readonly reqs, but write access (#777)


dancoates (approved):
👍

Issue #780, opened 2024-05-15
Cohort Endpoint returns all Cohorts, not just the ones in the specified project. 


vivbak:
This leads to:

A) Inaccurate outputs for the project.
B) Permissions error if a cohort has been created in a project that a user does not have access to.

PR #781, opened 2024-05-16
Custom Cohorts Fixes

vivbak:
Fixes https://github.com/populationgenomics/metamist/issues/780 

The issues were:
1- Project was not being passed to the filter, so the query would return everything (if you enter via project)
2- The template project was not being stored on the template model 
3- Cohort criteria was being returned as the model, but needs to be in dict form to display on UI, so required a to_external method.

### DIFF
--- a/models/models/cohort.py
+++ b/models/models/cohort.py
@@ -52,6 +52,28 @@ class CohortCriteriaInternal(SMBase):
     sg_type: list[str] | None = None
     sample_type: list[str] | None = None
 
+    def to_external(self) -> dict:
+        """
+        Convert to external model
+        """
+        return {
+            'projects': [str(p) for p in self.projects] if self.projects else [],
+            'sg_ids_internal': (
+                sequencing_group_id_format_list(self.sg_ids_internal_raw)
+                if self.sg_ids_internal_raw
+                else []
+            ),
+            'excluded_sgs_internal': (
+                sequencing_group_id_format_list(self.excluded_sgs_internal_raw)
+                if self.excluded_sgs_internal_raw
+                else []
+            ),
+            'sg_technology': self.sg_technology if self.sg_technology else [],
+            'sg_platform': self.sg_platform if self.sg_platform else [],
+            'sg_type': self.sg_type if self.sg_type else [],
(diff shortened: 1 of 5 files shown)


models/models/cohort.py
```diff
@@ -52,6 +52,28 @@ class CohortCriteriaInternal(SMBase):
     sg_type: list[str] | None = None
     sample_type: list[str] | None = None
 
+    def to_external(self) -> dict:
+        """
+        Convert to external model
+        """
+        return {
+            'projects': [str(p) for p in self.projects] if self.projects else [],
```
jmarshall, line 60: Should this be translating from numeric project id to textual project name?

jmarshall:
Codecov is right — the new `to_external()` method deserves a test. There might be somewhere natural in the existing tests to exercise it, or might want a new test case. I'm happy to do that if you want?

vivbak:
If you were able to test it that would be great @jmarshall!

models/models/cohort.py
vivbak: 🤦‍♀️ Yes, thank you. I just converted 1 to "1" 😂  Let me fix that!!

vivbak: Ok I've pushed a solution to this, but I'm not convinced that it is a very elegant approach so happy to take feedback if you can see a better way.

models/models/cohort.py
```diff
@@ -52,6 +52,28 @@ class CohortCriteriaInternal(SMBase):
     sg_type: list[str] | None = None
     sample_type: list[str] | None = None
 
+    def to_external(self, project_names: list[str]) -> dict:
```
jmarshall, line 55: Should probably be returning a `CohortCriteria` object rather than a raw dict?

vivbak: ah, i did it because I was getting this 'Object is not serialisable' error. (thinking...)

vivbak: I probably just need to create a strawberry type for it instead, will do that now

jmarshall:
Tests added — you'll need to change the asserts (to the commented out ones) if you manage to change the return type to `CohortCriteria`.

jmarshall (approved):

PR #784, opened 2024-05-16
fix for analysis runner entries that don't have a config path

dancoates:
this was breaking the api/ui when trying to display analysis runner entries for projects where the config path was not set.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "metamist",
-    "version": "6.8.0",
+    "version": "6.10.2",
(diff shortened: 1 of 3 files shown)


web/package-lock.json
illusional, line 1: Did you mean to commit this? I guess not bad though

illusional (approved):

web/package-lock.json
dancoates: yeah, I think someone forgot to run npm install after bumping the package version, figured it would be good to sneak it in

PR #786, opened 2024-05-17
Botocore/Boto3 relax pin

milo-hyben:
I made a silly mistake, based my branch on 'main' and not 'dev'.
Will close this one and create the new branch based on dev. Unless there is better way of doing it.


milo-hyben:
Close this one as this was a mistake :-)

PR #787, opened 2024-05-17
Relaxing botocore/boto3 pins.

milo-hyben:
And we need to relax Boto pins in metamist both as well.


milo-hyben:
Will bump version as well.

illusional (approved):
No need for bumpversion, this is a server only change. The setup.py controls the installed versions, and no issue there AFACT

milo-hyben:
> No need for bumpversion, this is a server only change. The setup.py controls the installed versions, and no issue there AFACT

Ok, good to know. Thank you!

PR #791, opened 2024-05-21
Analysis query fixes

### LINKED ISSUE #790: Querying analyses with many sequencing groups and large meta causes non-responsive queries
A query like:

```graphql
query MyQuery {
  project(name: "project-name") {
    analyses(type: {eq: "sv"}) {
      id
      output
    }
  }
}
```

will produce a SQL query that takes a very long time to resolve:

```sql
SELECT a.id as id, a.type as type, a.status as status,
        a.output as output, a_sg.sequencing_group_id as sequencing_group_id,
        a.project as project, a.timestamp_completed as timestamp_completed,
        a.active as active, a.meta as meta, a.author as author
FROM analysis a
LEFT JOIN analysis_sequencing_group a_sg ON a.id = a_sg.analysis_id
WHERE a.project = 6 AND a.type = 'sv' AND a.status = 'completed'
```

This is because each analysis has many sequencing groups and the large meta object needs to be replicated for each row produced by the join. 

We can potentially resolve this by moving the joining of the meta into the python code rather than having it in the DB.

dancoates:
Hopefully fixes #790 

This changes from querying the cross product join of analyses and sequencing groups to using a group by and JSON_ARRAYAGG to reduce the number of rows that the large meta object is propagated to.

Also removes a broken and unused function and fixes a few places that were still referring to sample ids on analyses.


db/python/tables/analysis.py
```diff
@@ -295,20 +295,19 @@ async def query(self, filter_: AnalysisFilter) -> List[AnalysisInternal]:
         else:
             _query = f"""
             SELECT a.id as id, a.type as type, a.status as status,
```
illusional, line 297: Can we potentially unify these two branches with JSON_ARRAYAGG, bevause we're currently not returning cohorts here (if you don't include a cohort in the filter), and there's a separate call to get sequencing-group-ids. 

IMO we should always LEFT JOIN on cohorts AND sequencing-groups, and then always JSON_ARRAYAGG so we're correctly returning the right metadata. We should probably also change the methods below to also return cohorts, otherwise I bet we're going to have an issue with "this endpoint is returning no cohorts when it should have them".

illusional (commented):
This fix is fine, but it highlights an actual issue where we're not correctly returning cohorts on one branch, and it looks as if it can be unified.

db/python/tables/analysis.py
dancoates: It actually does return cohorts, just a bit further down in a separate query:
https://github.com/populationgenomics/metamist/pull/791/files#diff-64b0b1278d5707365163edea17225c7b615255c0f4262d1e477b8456fc0f40bcL314

You're definitely right that it would be cleaner to be combined into one query though, I'll make that update

dancoates:
I've updated to combine the queries, I found that GROUP_CONCAT actually works better than JSON_ARRAYAGG as it doesn't include nulls and makes it easier to split back out to ids without having to parse the json for each row

db/python/tables/analysis.py
```diff
@@ -347,7 +294,7 @@ async def get_latest_complete_analysis_for_type(
 
         _query = f"""
 SELECT a.id as id, a.type as type, a.status as status,
-        a.output as output, a_sg.sample_id as sample_id,
+        a.output as output, a_sg.sequencing_group_id as sequencing_group_id,
```
illusional, line 297: Same agg treatment here might avoid copying the a.meta multiple time in this query?

db/python/tables/analysis.py
```diff
@@ -397,24 +344,24 @@ async def get_incomplete_analyses(
         """
         Gets details of analysis with status queued or in-progress
         """
+
         _query = """
 SELECT a.id as id, a.type as type, a.status as status,
-        a.output as output, a_sg.sequencing_group_id as sequencing_group_id,
+        a.output as output, JSON_ARRAYAGG(a_sg.sequencing_group_id) as _sequencing_group_ids,
```
illusional, line 350: Can we return cohort_ids to this query, probably should've been done in the initial custom cohorts PR

illusional (approved):

db/python/tables/analysis.py
jmarshall: The two branches was an unfortunate hack: if cohorts aren't part of the query, then joining against the cohort table would seem to ensure zero rows returned. So we made two branches as an expedient solution. If we had spent longer on it we would have preferred implementing this as a single query, and look forward to seeing this update.

db/python/tables/analysis.py
dancoates: Good point, I realised this can just be updated to use the query method too, saves a bunch of duplicated code

db/python/tables/analysis.py
dancoates: Having a `LEFT JOIN` should mean that analyses are returned even if there are no cohorts or sequencing groups in the join tables

dancoates:
Also probably worth noting that cohorts aren't actually in the graphql schema for analyses so even though this returns cohort ids you can only get them through the rest api, not sure if it would be worth adding to the schema, but probably out of scope of this change as it would require adding a new loader for cohorts

nevoodoo:
I think this should also close #798 once it is merged into main :)

PR #776, opened 2024-05-14
Update FastAPI + Pydantic

illusional:
- Spiritual successor to #643
- Add a hack to make the OpenAPI 3.1 appear like 3.0 to appease the open-api-generator
- FastAPI new lifecycle method: https://fastapi.tiangolo.com/advanced/events/
- Compile requirements and requirements-dev so it's easier to track dependency changes
    - Drop `python-multipart`, but it's present in the requirements.txt
- Add SMBase.model_config to restrict extra params
- Make GenericFilter inherit from SMBase (which accidentally carried through, and could revert if it makes it easier) - it's actually the change I wanted to make initially.
- Fix the CORS configuration option

Substantially changes the following methods, which requires a `bumpversion major`.

- getAssaysByCriteria ([cs](https://github.com/search?type=code&q=org%3Apopulationgenomics+get_assays_by_criteria))
- getParticipants ([cs](https://github.com/search?type=code&q=org%3Apopulationgenomics+get_participants))
- getSamples ([cs](https://github.com/search?type=code&q=org%3Apopulationgenomics+get_samples), w/ two usages: [analysis-runner:copy_family_to_test](https://github.com/populationgenomics/analysis-runner/blob/3ba1e315d01c9cef676a0a8c4f3f9ad0ccae0d29/scripts/copy_family_to_test.py#L84), [joint-calling:sm_utils](https://github.com/populationgenomics/joint-calling/blob/0520237cbc78a59fca407b4d2781b53344365c2e/joint_calling/sm_utils.py#L290))

PR #792, opened 2024-05-21
Pin requests to 2.31.0 due to breaking change

nevoodoo:
## Description
Unit tests are failing due to the latest release of `requests==2.32.0` which isn't compatible with `docker-py`. This pins the package to `2.31.0`.

## Notes
- The unit tests pass on this branch, unsure if we should pin in the `setup.py` file too?


illusional:
Closing due to #776 being deployed

PR #793, opened 2024-05-22
Assorted Cohort Fixes 

vivbak:
- Cohort -> Project and CohortTemplate -> Project were previously not working. This PR defines the project_id on the GraphQLCohort and GraphQLCohortTemplate objects, so that it can be correctly passed to the loader. It also creates the project field on the GraphQLCohortTemplate. 
- You could not correctly filter for cohort template ids, due to a type mismatch. This PR fixes the type defined for the template_id.


vivbak:
FYI @MattWellie

vivbak:
Note: Unit tests are currently failing due to a known issue https://github.com/populationgenomics/metamist/pull/792

milo-hyben (approved):
This PR is looking good to me! Great work on fixing this one.

I only have two comments not regarding this change, but more in general:
1.  'project' vs 'project_id' 
I really like this change call it what it is, 'project_id', 
'connection.project' expecting 'project_id' that is very confusing legacy. We should refactor connection at some stage.

2. template_id, it seems it can be str or int, without knowing much of the reasons I think it should be a string all the way up to a db table, would make it easier I think, 
Here is what confuses me:

CohortInternal.template_id is a int
https://github.com/populationgenomics/metamist/blob/879e945967397752e9a09657b73042c16376c95e/models/models/cohort.py#L20

however CohortBody.template_id is a string:
https://github.com/populationgenomics/metamist/blob/879e945967397752e9a09657b73042c16376c95e/models/models/cohort.py#L114

PR #799, opened 2024-05-24
Update SG meta typing to dict[str, Any]

EddieLF:
This updates the sequencing group model to allow the `meta` dict to take any type of value, not just strings. 

While most of the other models with `meta` fields (e.g. [`participant`](https://github.com/populationgenomics/metamist/blob/dev/models/models/participant.py#L23), [`sample`](https://github.com/populationgenomics/metamist/blob/dev/models/models/sample.py)) are typed with `meta: dict`, the `meta` dict for the model [`assay`](https://github.com/populationgenomics/metamist/blob/dev/models/models/assay.py#L13) is typed: `dict[str, Any]`. This would match that typing for sequencing group `meta`. 

If we should limit sequencing group meta dicts to only contain string values it would be good to know why. 
Also not sure if this would require a bumpversion to release.

--- 

I encountered this when trying to upsert a sequencing group with a meta dict that included `{"haplotag": true}`. The HTTP request returned a 422 exception:

```
File ".../venv/lib/python3.11/site-packages/metamist/rest.py", line 225, in request
    raise ApiException(http_resp=r)
metamist.exceptions.ApiException: (422)
...
HTTP response body: {"detail":[
    {
     "type":"string_type",
     "loc":["body",0,"samples",0,"sequencing_groups",0,"meta","dict[str,str]","haplotag"],
     "msg":"Input should be a valid string",
     "input":true
    }, ...
]
```


illusional (approved):
Thanks @EddieLF!
Theoretically yes, and probably functionally yes, depends if the generated API is validating types on the way in (probably not), and we don't check it on response in the client.

Issue #814, opened 2024-06-11
Create Test Subset Fails When epid != esid 


vivbak:
At the moment, if the external participant id does not match the external sample id, the script will fail with a key error. 

```
 File "/Users/vivian/Desktop/Code/sample-metadata/scripts/create_test_subset.py", line 461, in transfer_analyses
    new_sequencing_group_id = get_new_sg_id(
  File "/Users/vivian/Desktop/Code/sample-metadata/scripts/create_test_subset.py", line 425, in get_new_sg_id
    sample_to_sg_attribute_map[peid_sid_key].get(new_sg_attributes)
KeyError: ('NA12878', 'XPGLCL85') 
```

It makes sense that this went unnoticed, because often these values are the same in production.

PR #815, opened 2024-06-11
Fix mismatch between epid and esid in create_test_subset 

vivbak:
Closes https://github.com/populationgenomics/metamist/issues/814

When the sample_to_sg_attribute_map is created, the keys are created like so: 

`exid_old_sid = (s['participant']['externalId'], s['id']) `

When the code later references this map, it creates the keys like so 

`seid_sid_key = (new_sample['externalId'], old_sid)` 

Using the externalID from the participant when setting up the map, and using the externalID from the sample when accessing it later. 

We often use the same name for both, so this can go relatively unnoticed in production.


michael-harper (approved):
Ooo annoying how this went unnoticed, looks like using sample external ID is more robust way to approach 👏

PR #816, opened 2024-06-11
Update generic_parser from_json return after ingestion

EddieLF:
This change makes the `generic_parser.from_json` function return the full response it receives from Metamist. 

When the parser is used to ingest from a manifest, the parser is instantiated:
```python
parser = GenericParser(
    project=dataset,
    search_paths=search_paths,
)
```
We then call `parser.from_manifest_path`
```python
response = await parser.from_manifest_path(manifest=manifest_path, dry_run=dry_run)
```
This returns `parser.parse_manifest`, which in turn returns `parser.from_json`. Currently, `from_json` has no returned value unless the `--dry-run` flag is set. 
This means that nothing can be done to analyse the response beyond what gets logged by the generic parser. 
```python
response = await parser.from_manifest_path(manifest=manifest_path, dry_run=dry_run)
print(response)
>>> None
```

By adding a `return result` to `from_json`, the result/response can be printed or analysed by any custom parser implementation. 
```python
response = await parser.from_manifest_path(manifest=manifest_path, dry_run=dry_run)
print(json.dumps(response, indent=2))
>>> { Actual JSON response }
```


illusional (approved):
Seems sensible!

PR #742, opened 2024-04-24
OurDNA: Improve project table view and filtering

illusional:
OurDNA RFC: https://docs.google.com/document/d/1bG8oZQdhA3lEeqlvnCZ1n_P1OqaqDBPlMPB7oz275ns/edit#heading=h.sxmryphgwrnn

Aims:
- Improve the filtering and usability of the table grid on the project page
- Make this page generally more useful

Consider:
- How to hide / show columns for specific projects
- Exporting data to CSV / etc

Actual changes:

- [large changes]:
    - Split the project summary from participant grid
    - Further decompose component structure for participant grid
    - Use generic filters for participant grid instead of custom query format
    - Add ability to export changes
    - Added ability to write more expressive queries
    - Encode more params on the URL to make sharing better
    - Migrate in support for multiple external IDs for participants / samples.

- [unrelated changes]
    - Display improvements to ErrorBoundary
    - Upgrade Vite
    - Add startswith GenericFilter operator
    - Improve participant query, by allowing for more nested objects to query against
    - Add participant tests

PR #823, opened 2024-06-13
Improved table filtering

illusional:
- Move filtering classes to own namespace
- Add more nested fields to sample / sequencing group filtering

NB: This is built on #742, so there will be some strange diffs, because some of the participant query changes were created there.


milo-hyben (approved):
From the quick review and as per our quick chat, I would only rename 'db_filters' to 'filters'
This reads better:
from db.python.filters import GenericFilter
than:
from db.python.db_filters import GenericFilter

Will wait till you merge both PRs into one.

PR #824, opened 2024-06-13
Hotfix for external IDs on the Project Grid

illusional:
Views like this:

image

### DIFF
--- a/web/src/pages/project/ProjectGrid.tsx
+++ b/web/src/pages/project/ProjectGrid.tsx
@@ -307,7 +307,7 @@ const ProjectGrid: React.FunctionComponent<ProjectGridProps> = ({
                                 // debugger
                                 return (
                                     <SUITable.Row
-                                        key={`${p.external_id}-${s.id}-${sg.id}-${assay.id}`}
+                                        key={`${p.id}-${s.id}-${sg.id}-${assay.id}`}
                                     >
                                         {isFirstOfGroup && (
                                             <SUITable.Cell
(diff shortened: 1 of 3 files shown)


web/src/pages/project/ProjectGrid.tsx
```diff
@@ -300,9 +306,7 @@ const ProjectGrid: React.FunctionComponent<ProjectGridProps> = ({
                                 // const border = '1px solid'
                                 // debugger
                                 return (
-                                    <SUITable.Row
-                                        key={`${p.external_id}-${s.id}-${sg.id}-${assay.id}`}
-                                    >
+                                    <SUITable.Row key={`${p.id}-${s.id}-${sg.id}-${assay.id}`}>
```
jmarshall, line 309: Does this need any corresponding updates elsewhere (e.g. in code that tries to link to this row key)?

illusional: Na, [keys](https://legacy.reactjs.org/docs/lists-and-keys.html#keys) are just for react to tell what's changed between renders

jmarshall: Excellent.

jmarshall (approved):
LGTM. I can think of maybe slightly less gross ways to do it (e.g. rename `sanitiseValue` to `displayValue`, maybe give it `k` as a parameter too, and teach it how to display `external_ids` specially; or alternatively have `formatExternalIds` able to format single ids too) but this is great as a hot fix.

illusional:
Yeah agreed :) I've completely rewritten the page in #742, so no point doing anything super crazy here

PR #826, opened 2024-06-17
Release

jmarshall:
Small hot-fix release (without version bump) to fix:

* Handle multiple participant external_ids in create_test_subset.py (#825)


jmarshall:
We could also slide #821 in (can't remember if we decided to merge that imminently too), but as it's a source refactoring only it doesn't really matter when it hits **main**, only **dev**.

illusional (approved):

PR #827, opened 2024-06-17
Proof of concept: Allow selecting of slices of assay meta in graphql

dancoates:


db/python/tables/assay.py
```diff
@@ -83,15 +86,30 @@ async def query(
...
 
         conditions, values = filter_.to_sql(sql_overides)
-        keys = ', '.join(self.COMMON_GET_KEYS)
+        keys = ', '.join(
+            [k for k in self.COMMON_GET_KEYS if include_meta or k != 'a.meta']
+        )
+        meta_slice_keys = (
+            ','.join(
+                [
+                    f'JSON_VALUE(a.meta, \'{m["path"]}\') as _meta_{m["alias"]}'
```
dancoates, line 95: definitely a sql injection issue here, would need fixing if we want to do this for real

api/graphql/schema.py
```diff
@@ -928,6 +946,13 @@ async def sample(self, info: Info, root: 'GraphQLAssay') -> GraphQLSample:
         sample = await loader.load(root.sample_id)
         return GraphQLSample.from_internal(sample)
 
+    @strawberry.field
+    async def metaValue(
+        self, info: Info, root: 'GraphQLAssay', path: str
+    ) -> strawberry.scalars.JSON:
+        alias = info.selected_fields[0].alias
```
dancoates, line 953: would also need to handle un-aliased fields here

db/python/tables/assay.py
dancoates: Should also use JSON_EXTRACT rather than JSON_VALUE

PR #828, opened 2024-06-18
Get_cram_paths external IDs fix

EddieLF:
A quick fix for the `get_sample_cram_path_map_for_seqr` Analysis table function to allow it to work with the recent external ID changes. Now joins to the `participant_external_id` table to get the external ID.


db/python/tables/analysis.py
```diff
@@ -431,16 +433,16 @@ async def get_sample_cram_path_map_for_seqr(
             filters.append('JSON_VALUE(a.meta, "$.sequencing_type") ' + seq_check)
 
         if participant_ids:
-            filters.append('p.id IN :pids')
+            filters.append('peid.participant_id IN :pids')
```
jmarshall, line 436: This field needs to be changed to `participant_id` (as per the new table) here too. And a minor renaming of the table alias.

test/test_analysis.py
```diff
@@ -167,3 +172,22 @@ async def test_get_analysis(self):
...
+            [
+                ParticipantUpsertInternal(
+                    external_ids={PRIMARY_EXTERNAL_ORG: 'PEXT1'},
+                    meta={},
+                    samples=[SampleUpsertInternal(id=self.sample_id)]
+                ),
+            ],
+        )
+
+        id_map = await self.al.get_sample_cram_path_map_for_seqr(self.project_id, ['blood'], [part[0].id])
```
jmarshall, line 192: Added a basic test case so that this code gets exercised. There's not enough data set up to get a non-empty `id_map` back, so we don't check the result very much.

db/python/tables/analysis.py
```diff
@@ -413,12 +414,13 @@ async def get_sample_cram_path_map_for_seqr(
...
 
-        values: dict[str, Any] = {'project': project}
+        values: dict[str, Any] = {'project': project, 'PRIMARY_EXTERNAL_ORG': PRIMARY_EXTERNAL_ORG}
         filters = [
             'a.active',
             'a.type = "cram"',
             'a.status = "completed"',
-            'p.project = :project',
+            'peid.project = :project',
+            'peid.name = :PRIMARY_EXTERNAL_ORG',
```
jmarshall, line 423: At least for now, maintained the status quo by only returning a row for the primary external id rather than several, one for each external id. This is a technique I used in several of the “higher level” queries, and will want to be revisited eventually.

jmarshall (approved):
Thanks for the fix and apologies for the inconvenience!

I've made a couple of additions, and commented on the significant ones.

LGTM — The fix should be good to go now.

EddieLF:
Thanks for the help @jmarshall, appreciate your input and the inclusion of a basic test case

jmarshall:
Thanks for finding the problem…

If that test case had been there previously, this wouldn't have been missed out in the multiple-ext-ids PR because even this basic test would have failed. Turns out test coverage is a good thing!

PR #829, opened 2024-06-18
Metamist permissions system update

dancoates:
## Context

Metamist currently has a fairly simple permission system. Each project defines a **read group** and a **write group**, these two groups define a list of **group members**. The members of the read group will have readonly access to the project, and the members of the write group have full unrestricted write access to the project. 

Write actions within metamist are primarily performed programmatically via the API, but with some upcoming new features there will soon be the need to allow human users to perform a limited set of __write__ actions. Some examples of these new features are:

- __Custom cohorts__, where we will want to allow users to create cohorts
- __Attachments__, where we will want to allow users to attach comments or other attachment types to various entities within metamist.

We do not want to grant full write access to allow users to perform these actions, so it is necessary to create a more nuanced role based permissions system that allows limited access to certain write operations but restricts access to anything destructive. 

## Solution

Moving away from storing the majority of permissions in groups to creating a direct relationship between users and projects. There is a lot of unnecessary redundancy from the way permissions are stored currently. By convention a `read` group will be suffixed with `-read` and similar for a write group. These are then referenced in the `read_group_id` and `write_group_id` columns in the `project` table. To continue extending this would result in a complex system with lots of columns on the `project` table and potentially a situation where the group suffixed with `-read` ends up being the write group.

This update introduces a new table `project_member`. This table holds the relationship between a member (represented by an email) and the project. In addition it includes a `role` column which specifies a role that the user has within that project. Each member can have multiple roles within a project by there being multiple rows for them in the `project_member` table for any given project. 

In addition to the table updates, this PR also moves much of the permissions checking logic away from the `tables/project.py` file which acts as the interface for projects in the database. Instead this logic is now on the Connection class. This class is available pretty much everywhere throughout the project so it makes it much easier to quickly check permissions. This change also mitigates a class of performance problems that was starting to occur in the graphql resolvers where permissions checks were not being cached between resolvers. The Connection object is constructed once at the beginning of every request and all permission checks from then on are synchronous and should be very fast.

A permission check looks like:

```python
self.connection.get_and_check_access_to_projects_for_ids(
    project_ids=[1,2,3],
    allowed_roles=ReadAccessRoles,
)
```

Where `ReadAccessRoles` is a set of the project member role enums. If the current user has any one of the specified roles in all of the specified projects the permission check will pass.


db/python/tables/project.py
```diff
@@ -465,209 +266,71 @@ async def delete_project_data(
...
-        """
-
-        all_projects = await self._get_project_rows_internal()
-        seqr_projects = [p for p in all_projects if p.meta.get('is_seqr')]
-        return seqr_projects
-
-    # gruo
-
-
-class GroupTable:
```
dancoates, line 522: I've removed all these methods, as they weren't exposed via any routes and groups are very minimal now, there should only ever be the two admin groups. I'm thinking that we may rethink or remove those admin groups entirely in the future too. But I'm happy to add these back in we reckon they'll be useful in the future.

db/python/tables/project.py
dancoates, line 1: This diff is pretty nasty, probably would be easier to look at as a deletion of the old file and creation of a new file. There isn't much left from the old file.

test/test_project_groups.py
dancoates, line 1: Should add a test in here for the logic allowing a data manager to delete a test project

docs/installation.md
```diff
@@ -288,20 +288,27 @@ export SM_DEV_DB_PORT=3306 # or 3307
...
-# skips permission checks in your local environment
-export SM_ALLOWALLACCESS=true
 # uses your username as the "author" in requests
 export SM_LOCALONLY_DEFAULTUSER=$(whoami)
 ```
 
+To allow the sytem to be bootstrapped and create the initial project, you'll need to add yourself to the two admin groups that allow creating projects and updating project members:
+
+```sql
+INSERT INTO group_member(group_id, member)
```
illusional, line 306: Potentially give the mariadb --execute "..." command with the whoami substitution?

test/data/generate_data.py
```diff
@@ -76,6 +77,14 @@ async def main(ped_path=default_ped_location, project='greek-myth'):
         await papi.create_project_async(
             name=project, dataset=project, create_test_project=False
         )
+        default_user = os.getenv('SM_LOCALONLY_DEFAULTUSER')
```
illusional, line 80: Some defensiveness here would be valuable - I don't usually have this environment variable set in my bash environments

test/data/generate_seqr_project_data.py
```diff
@@ -422,9 +454,20 @@ async def main():
             await papi.create_project_async(
                 name=project, dataset=project, create_test_project=False
             )
+
+            default_user = os.getenv('SM_LOCALONLY_DEFAULTUSER')
```
illusional, line 458: Ditto

db/python/tables/project.py
```diff
@@ -465,209 +266,71 @@ async def delete_project_data(
...
 
-    def __init__(self, connection: Database, allow_full_access: bool = None):
-        if not isinstance(connection, Database):
-            raise ValueError(
-                f'Invalid type connection, expected Database, got {type(connection)}, '
-                'did you forget to call connection.connection?'
+            # Get existing rows so that we can keep the existing audit log ids
+            existing_rows = await self.connection.fetch_all(
+                """
+                select project_id, member, role, audit_log_id
```
illusional, line 286: Upper case here for consistency

db/python/tables/project.py
```diff
@@ -465,209 +266,71 @@ async def delete_project_data(
...
-        _query = """
-            INSERT INTO `group` (name, audit_log_id)
-            VALUES (:name, :audit_log_id)
-            RETURNING id
-        """
-        return await self.connection.fetch_val(
-            _query, {'name': name, 'audit_log_id': audit_log_id}
-        )
+        if self._connection:
+            await self._connection.refresh_projects()
```
illusional, line 334: This is pretty circular haha, but makes sense that we're caching this information on the connection now

illusional (changes requested):
This is looking fab! I have a few requests, but all-in-all, a very worthy upgrade. 

What's your sense with how consistent our terminology is for referring to an authenticated "customer". We use author a lot, but I wonder if that's lazy, or something more accurate.

I'm _slightly_ concerned by introducing a stronger concept of test projects here, metamist currently has practically no (except on creation) link between test and _full_ projects. My default position is cpg-infra should do any of that logic for us, and metamist just follows whatever list is produced (hence, it's clear where any permission change actually needs to originate from).

FWIW, prior to https://github.com/populationgenomics/metamist/pull/568, things were cached at the application level through static properties, so it allowed for lazier patterns, certainly required when looking up [values in secrets](https://github.com/populationgenomics/metamist/blob/4460b88ad2668dca840d9be7f2f0843ff7ce63c8/db/python/tables/project.py#L183) which despite it's name was [not locally cached](https://github.com/populationgenomics/cpg-utils/blob/42f1e5339258bca9f1e229d665971fda30552531/cpg_utils/membership.py#L8).

https://github.com/populationgenomics/metamist/blob/4460b88ad2668dca840d9be7f2f0843ff7ce63c8/db/python/tables/project.py#L53-L59

api/routes/project.py
```diff
@@ -56,61 +69,85 @@ async def create_project(
...
     Requires READ access + project-creator permissions
     """
     ptable = ProjectPermissionsTable(connection)
-    p_obj = await ptable.get_and_check_access_to_project_for_name(
-        user=connection.author, project_name=project, readonly=False
+
+    assert connection.project
+
+    # Allow data manager role to delete test projects
+    data_manager_deleting_test = (
```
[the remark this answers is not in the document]
dancoates: Sorry, I'm not quite following how we would do this from cpg-infra? For context - this change is related to the discussion where Amy mentioned it would be handy for data managers to be able to clear out test projects https://centrepopgen.slack.com/archives/C020PDNH3N0/p1717389879026539

api/routes/project.py
```diff
@@ -56,61 +69,85 @@ async def create_project(
...
     Not that this is protected by access to a specific access group
     """
     ptable = ProjectPermissionsTable(connection)
-    await ptable.set_group_members(
-        group_name=ptable.get_project_group_name(project, readonly=readonly),
-        members=members,
-        author=connection.author,
-    )
+
+    await ptable.check_member_admin_permissions(author=connection.author)
```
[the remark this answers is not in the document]
dancoates: Whoops, yep, it actually is already on the connection, this line can just be deleted, thanks

api/utils/db.py
```diff
@@ -59,6 +60,12 @@ def authenticate(
     If a token (OR Google IAP auth jwt) is provided,
     return the email, else raise an Exception
     """
+
+    if default_user := get_default_user():
```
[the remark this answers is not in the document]
dancoates: Yeah I don't love it either, the problem is that the generated API by default will get your local GCP credentials. So when running scripts like `generate_data.py`, `x_goog_iap_jwt_assertion` will exist and the default user won't be used, this causes problems because the authenticated user doesn't have permissions. I think it is pretty safe as `get_default_user()` has an explicit environment check.

db/project.xml
```diff
@@ -1420,4 +1420,69 @@
...
+			<column name="member"/>
+		</createIndex>
+
+		<sql>ALTER TABLE project_member ADD SYSTEM VERSIONING;</sql>
+
+
+		<!-- Copy data across to new table -->
+
+		<sql>
+			-- copy across read and write roles from existing project table
```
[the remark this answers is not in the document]
dancoates: Sorry, not sure I understand, could you expand a bit please?

db/python/tables/project.py
dancoates: Yeah this is really only for the tests, seeing as the connection is constructed per-request it generally isn't necessary to refresh them. It's only the tests where we create a project and then do something else

docs/installation.md
dancoates: I think there's maybe too many permutations and combinations of possible setups to provide an exact command? Personally I don't have the mariadb executable available because I run it in docker, so would require a docker command, but that wouldn't work for others.

api/settings.py
```diff
@@ -60,22 +59,11 @@
 
 def get_default_user() -> str | None:
     """Determine if a default user is available"""
-    if is_all_access() and _DEFAULT_USER:
+    if SM_ENVIRONMENT == 'local' and _DEFAULT_USER:
```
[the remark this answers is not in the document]
dancoates: Yeah I like that, we don't really use env checks anywhere in the tests but I've updated to set it to `test` in the tests so we could in the future

db/python/connect.py
```diff
@@ -50,31 +58,155 @@ class Connection:
...
+        project_name_map: dict[str, Project],
         author: str,
         on_behalf_of: str | None,
-        readonly: bool,
         ar_guid: str | None,
         meta: dict[str, str] | None = None,
     ):
         self.connection: databases.Database = connection
-        self.project: int | None = project
+        self.project: Project | None = project
```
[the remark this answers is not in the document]
dancoates: Good idea, I'm not 100% on the best way to protect class variables in python but I've had a go at it, let me know if there's a better way

dancoates:
> What's your sense with how consistent our terminology is for referring to an authenticated "customer". We use author a lot, but I wonder if that's lazy, or something more accurate.

Yeah I mused on this a bit when working on the changes. In the past I've used the concept of a "viewer" which is defined to mean the currently active or viewing user. I think that name was a bit of a quirk of my last job though, I haven't seen it used elsewhere. "AuthenticatedUser" would be a more verbose but clearer name. At the moment the `Connection` class is performing that role, we could probably change that name as we kind of have two concepts of a connection - an API connection and a database connection, leading to a few places where you have `connection.connection`.

api/routes/project.py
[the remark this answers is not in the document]
dancoates: Oh I think I might have just understood, we could give "data-managers" as defined in cpg-infrastructure some sort of project admin role for test projects. Rather than having a data manager role in metamist. That does sound like a better way of handling it. At the moment project admin in kind of a pseudo-role that is set by being in the `project-creators` group, but it would be possible to make it a role that can be set in the project_members table too.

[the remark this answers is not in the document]
illusional: It's creeping outside the scope a bit, but I wonder if we should change it in the generated API to not provide an auth token if local / test? Then we wouldn't need to re-arrange here, and would probably speed up local testing by not always generating this token which does take some amount of time.

https://github.com/populationgenomics/metamist/blob/846c6b80006e0264c1153998bb825f0804bd68c9/openapi-templates/configuration.mustache#L502-L503

api/utils/db.py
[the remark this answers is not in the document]
illusional: It's creeping outside the scope a bit, but I wonder if we should change it in the generated API to not provide an auth token if local / test? Then we wouldn't need to re-arrange here, and would probably speed up local testing by not always generating this token which does take some amount of time.

https://github.com/populationgenomics/metamist/blob/846c6b80006e0264c1153998bb825f0804bd68c9/openapi-templates/configuration.mustache#L502-L503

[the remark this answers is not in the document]
dancoates: Just checked this and as long as you have the SM_ENVIRONMENT=local set when running the generate data script it works fine, I must have not had it set when I was testing. I've reordered the checks back to what they were

ckrishk (approved):

api/settings.py
```diff
@@ -9,9 +9,8 @@
 LOG_DATABASE_QUERIES = (
     os.getenv('SM_LOG_DATABASE_QUERIES', 'false').lower() in TRUTH_SET
 )
-_ALLOW_ALL_ACCESS: bool = os.getenv('SM_ALLOWALLACCESS', 'n').lower() in TRUTH_SET
 _DEFAULT_USER = os.getenv('SM_LOCALONLY_DEFAULTUSER')
-SM_ENVIRONMENT = os.getenv('SM_ENVIRONMENT', 'local').lower()
+SM_ENVIRONMENT = os.getenv('SM_ENVIRONMENT', 'production').lower()
```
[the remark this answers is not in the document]
dancoates: yeah, it's better to default to production as the local env removes auth protections, this way if we forget to set the environment it defaults to a safe value.

api/routes/project.py
```diff
@@ -56,61 +69,71 @@ async def create_project(
 @router.get('/seqr/all', operation_id='getSeqrProjects')
 async def get_seqr_projects(connection: Connection = get_projectless_db_connection):
     """Get SM projects that should sync to seqr"""
-    ptable = ProjectPermissionsTable(connection)
-    return await ptable.get_seqr_projects()
+    projects = connection.all_projects()
```
illusional, line 72: Could we actually return all projects, filter for seqr and check if the user has access to them all?
For example, we _always_ want to joint-call all seqr projects, and if the service-account doesn't have access to a seqr project, then we should fail here.

illusional (approved):
Awesome, super clean! 

Not blocking this PR, but can you write up a bit of documentation for how this works, and how developers should interact with it? For example, how project_admin permission works, when that role isn't given by cpg-infra (I see how it works in code, but worth writing down). And any handy methods people should take note of :)

dancoates:
> Awesome, super clean!
> 
> Not blocking this PR, but can you write up a bit of documentation for how this works, and how developers should interact with it? For example, how project_admin permission works, when that role isn't given by cpg-infra (I see how it works in code, but worth writing down). And any handy methods people should take note of :)

Yeah, good thinking. I'll definitely write up some docs

api/routes/project.py
dancoates: Thanks, that's a great pick up, I didn't realise that is what this endpoint was used for, I've updated the code to raise a forbidden error if the user doesn't have access to list all seqr projects

PR #831, opened 2024-06-19
Update fix_pbmc_sample_participant.py script for multiple external ids

jmarshall:
I didn't previously audit the ad-hoc scripts directory too carefully… having now done so, alongside _create_test_subset.py_ I think this is the only one needing adjustments.

Expand `sample_map_by_external_id` to hold all the external_ids for all the samples. For samples with an external_id containing '-PBMC', use `next(...)` to pick the first such external_id — generally there'll surely only be one anyway.

### DIFF
--- a/scripts/fix_pbmc_sample_participant.py
+++ b/scripts/fix_pbmc_sample_participant.py
@@ -19,12 +19,12 @@ def main():
         )
     )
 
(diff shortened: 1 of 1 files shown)


jmarshall:
Incorporated into PR #866.

PR #836, opened 2024-06-24
Workaround for graphql performance issue when using generic input types

dancoates:
At the moment there is a issue with strawberry graphql which causes performance issues in cases where the field type for a large result set uses a generic input type. There is an outline of this issue here: https://github.com/strawberry-graphql/strawberry/issues/3544 and a stripped down reproduction case here https://github.com/dancoates/strawberry-generic-input-repro/blob/main/schema.py.

This PR introduces a workaround to this issue by making input types concrete before they are used in the GraphQL schema. This does introduce some code duplication, but IMO is worth it for the speed up, in my testing it has made requests 3-5x faster.


dancoates:
Closing in favour of #841

PR #840, opened 2024-06-25
Add support for nested samples

illusional:
- Adds support for nested samples (up to SW limited 10 depth on upsert)
- DB model links to parent AND root sample IDs, which makes it easier to query
- Add `isnull` generic filter (turns into `{field} IS NULL` / `{field IS NOT NULL}`
- Update OurDNA dashboard to look for root samples (where root_sample_id isnull)
- Add these new sample root / parent IDs to participant + sample filter
- Temporarily Remove foreignkey checks when clearing test database (+ add comment about truncates)
- Upgrades liquibase because I suspected an error in recognising `cascadeDelete`

Design doc: https://docs.google.com/document/d/1km3MU2NhlP0bv2uafnhDOGGqmK7vPluzThfLl9xmKVE/edit#heading=h.82ap8j490z81


test/testbase.py
```diff
@@ -110,8 +110,8 @@ async def setup():
                 if am_i_in_test_environment:
                     db_prefix = '../db'
 
-                con_string = db.get_connection_url()
-                con_string = 'mysql://' + con_string.split('://', maxsplit=1)[1]
+                # con_string = db.get_connection_url()
```
milo-hyben, line 113: I think those 2 commented lines can be removed.

milo-hyben (approved):
Looking good to me, just that small thing regarding commented lines.

PR #841, opened 2024-06-26
Upgrade to dev version of strawberry graphql to improve performance of resolvers with generic inputs

dancoates:
At the moment there is a issue with strawberry graphql which causes performance issues in cases where the field type for a large result set uses a generic input type. There is an outline of this issue here: https://github.com/strawberry-graphql/strawberry/issues/3544 and the PR with the fix is here https://github.com/strawberry-graphql/strawberry/pull/3549

This PR upgrades to the new dev version of strawberry which version includes a fix for the issue. Once an official release is out we can upgrade again.


illusional (approved):

patrick91:
This has been released in Strawberry 0.235.1 😊

PR #843, opened 2024-06-26
Bump to latest Strawberry

patrick91:


illusional:
Thanks @patrick91, really appreciate the fix here! We've been using the strawberry pre-release, and seems to be working great. Thanks for compiling our all dependencies too :D

illusional (approved):

PR #849, opened 2024-07-01
Participant / Sample External IDs in parser

illusional:
Add support for multiple participant and external IDs in the parsers, in two ways:

- Allow some subclass to provide an implementation for `get_{participant/sample}_external_ids`.
- Add an implementation here for the `GenericMetadataParser`.
- Switching matching participants / samples to look through ALL IDs.
- Add a docstring to better describe parameters.
- Fix a bug in update parent / root for nested samples.
-


db/python/layers/sample.py
```diff
@@ -243,8 +243,8 @@ async def upsert_sample(
                         participant_id=s.participant_id,
                         type_=s.type,
                         active=s.active,
-                        sample_parent_id=sample_parent_id,
-                        sample_root_id=sample_root_id,
+                        sample_parent_id=r.parent.id if r.parent else sample_parent_id,
```
illusional, line 246: Not strictly relevant to this PR, but a bug from #840

metamist/parser/generic_parser.py
```diff
@@ -963,33 +988,52 @@ async def validate_rows(self, rows: list[SingleRow]):
...
+    def get_reported_gender(self, row: GroupedRow) -> str | None:
         """Get reported gender from grouped row"""
         return None
 
-    def get_karyotype(self, row: GroupedRow) -> Optional[str]:
+    def get_karyotype(self, row: GroupedRow) -> str | None:
         """Get karyotype from grouped row"""
         return None
 
     # @abstractmethod
```
illusional, line 1033:
```suggestion
```

metamist/parser/generic_parser.py
```diff
@@ -190,6 +184,22 @@ def process_fieldname(self, fieldname: str) -> str:
         return fieldname.strip()
 
 
+def prepare_external_ids(
+    primary_external_id: str, external_ids: dict[str, str]
```
jmarshall, line 188: If we need `or {}` below, then this probably wants to be `dict[str, str] | None`

metamist/parser/generic_parser.py
```diff
@@ -198,15 +208,17 @@ def __init__(
...
+        reported_sex: int | None,
+        reported_gender: str | None,
+        karyotype: str | None,
+        external_pids: dict[str, str] | None = None,
     ):
         self.rows = rows
 
         self.internal_pid = internal_pid
         self.external_pid = external_pid
+        self.external_pids = external_pids or {}
```
jmarshall, line 221: Might deserve a comment noting that — unlike other layers — the primary one isn't in `external_pids` in this class.

metamist/parser/generic_parser.py
```diff
@@ -239,13 +251,16 @@ def __init__(
...
-        meta: Dict[str, Any] | None = None,
+        meta: dict[str, Any] | None = None,
+        external_sids: dict[str, str] | None = None,
     ):
         self.participant = participant
         self.rows = rows
 
         self.internal_sid = internal_sid
         self.external_sid = external_sid
+        self.external_sids = external_sids or {}
```
jmarshall, line 262: Ditto re comment

metamist/parser/generic_parser.py
```diff
@@ -667,9 +682,9 @@ async def from_json(self, rows, confirm=False, dry_run=False):
 
     def _get_dict_reader(self, file_pointer, delimiter: str):
         """
-        Return a DictReader from file_pointer
+        Return a dictReader from file_pointer
```
jmarshall, line 685: stet here and on the line a couple below

db/python/layers/sample.py
jmarshall: Fair enough!

jmarshall (approved):
Some minor comments, otherwise looks good to me.

Are the codecov notes complaining about missing test coverage on the new `get_participant_external_ids()` etc functions accurate?

PR #850, opened 2024-07-01
Fix get sequencing groups by analysis IDs

illusional:
- Add test to model failure case
- Change inner to left join to catch (most common) case where sequencing groups doesn't have any external ID.

### DIFF
--- a/test/test_analysis.py
+++ b/test/test_analysis.py
@@ -189,5 +189,71 @@ class TestAnalysis(DbIsolatedTest):
             ],
         )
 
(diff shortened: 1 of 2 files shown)


test/test_analysis.py
```diff
@@ -189,5 +189,71 @@ async def test_get_sample_cram_path_map_for_seqr(self):
             ],
         )
 
-        id_map = await self.al.get_sample_cram_path_map_for_seqr(self.project_id, ['blood'], [part[0].id])
+        id_map = await self.al.get_sample_cram_path_map_for_seqr(
```
illusional, line 192: This is still on me for not checking why my vscode settings are out of whack with CI

dancoates (approved):

PR #852, opened 2024-07-02
Restrict create_test_subset to GVCFs and CRAMs

illusional:
- Also sneaky fix for graphql analysis.project.name


michael-harper (approved):
Thanks @illusional !

PR #855, opened 2024-07-04
Performance profiling improvements

dancoates:
- Allow outputting html, json, or text reports
- Json reports can be viewed in https://www.speedscope.app/ for detailed flame charts
- Write file based profiles to a profile folder with a timestamp filename, so that they can be kept over time
- Add documentation about how to enable profiling
- ~Disable file watching unless running locally~


api/server.py
```diff
@@ -186,6 +224,5 @@ async def exception_handler(request: Request, e: Exception):
         'api.server:app',
         host='0.0.0.0',
         port=int(os.getenv('PORT', '8000')),
-        # debug=True,
-        reload=True,
+        reload=SM_ENVIRONMENT == 'local',
```
illusional, line 227: My understanding is this doesn't run in production, uvicorn selects the right app to deploy with:
```
uvicorn api.server:app
```

illusional (approved):
Great addition, thanks @dancoates!

api/server.py
dancoates: Ah yeah, of course! Thanks for pointing out - I've reverted this bit of the change

PR #856, opened 2024-07-05
First Class Files

nevoodoo:
# Description

Currently within metamist, we record paths in a few different ways, but inherently they’re stored in a text field. This means it’s difficult to work out:

- Where files are used in the centre,
- Metadata about specific files

The way we represent analysis and outputs has also evolved through the usage of metamist, and this is a good time to reflect on our future usage.

This replaces the current `Analysis` model to use a proper `OutputFile` class for storing output paths instead of a `str`. This will allow us to support multiple outputs in a structured way along with useful file stats. 

This issue has also been raised before with some prior work being done on it a while ago on #376 and more recently #657

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Backwards compatible change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [x] This requires a migration of some/all of the current relevant objects

# How Has This Been Tested?

The tests I've run so far:

- [x] Testing existing str path - I tested our current way of using the `output` field and maintained how it works. Ideally for the time being, we should be able to pass a str file path, and receive a str file path. This should also create a file object that is returned via the `outputs` field, if the path is valid and file meta can be obtained for it.
- [x] Testing single file object - We should also be able to pass input as a str to the `outputs` field (`output` only accepts str to deter it from being used for dict input), and a file object should be created for it. In the future, this field may only accept dict values.
- [x] Testing file with secondary files - We should now be able to pass a dict object containing multiple files, and additional secondary files to each file. This input structure can be maintained, and when data is returned, it follows the same structure - with meta properties add for both the file, and its secondary files.

# Checklist:

- [ ] Tested all `layers` methods (works with `get_analysis_by_id()`, others methods have the same code)
- [x] Added migration scripts
- [ ] Update UML diagrams
- [x] Added unit tests for this feature


models/models/output_file.py
```diff
@@ -0,0 +1,216 @@
...
+    ) -> dict | None:
+        """Get file info for file at given path"""
+        try:
+            file_checksum = None
+            valid = False
+            size = 0
+            # try:
+            if isinstance(file_obj, GSPath) and client:
+                if '/' in file_obj.blob:
+                    prefix = '/'.join(file_obj.blob.split('/')[:-1]) + '/'
```
dancoates, line 79: If I'm reading this right, it seems that this is listing all the blobs in the folder of the blob you are looking for, I just did a quick test and it seems to work fine if you just specify the full filename as the prefix, that should result in quite a lot fewer files being listed.

models/models/output_file.py
```diff
@@ -0,0 +1,216 @@
...
+                delimiter = '/'
+                blobs = client.list_blobs(
+                    bucket_or_name=str(file_obj.bucket),
+                    versions=False,
+                    prefix=prefix,
+                    delimiter=str(delimiter),
+                )  # pylint: disable=E1101
+                for blob in blobs:
+                    if blob.name == file_obj.blob:
+                        if file_obj.suffix != '.mt':
```
dancoates, line 91: Would be good to have a note here explaining why we only set checksums when it is a .mt file

models/models/output_file.py
```diff
@@ -0,0 +1,216 @@
...
+            nameroot=self.nameroot,
+            nameext=self.nameext,
+            file_checksum=self.file_checksum,
+            size=self.size,
+            meta=self.meta,
+            valid=self.valid,
+        )
+
+    @staticmethod
+    async def get_file_info(
```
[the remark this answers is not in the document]
dancoates: in fact, should this be returning the OutputFile model?

models/models/output_file.py
```diff
@@ -0,0 +1,216 @@
...
+                else just the output string based on the analysis_outputs table.
+
+        Returns:
+            dict: Should return the JSON structure based on the input data.
+        """
+        root: dict = {}
+        if isinstance(data, str):
+            return data
+
+        def add_to_structure(current, path, content):
```
dancoates, line 123: Would be good to have some more types here, and probably some comments. It is pretty tricky to tell what is going on.

db/python/tables/analysis.py
```diff
@@ -270,6 +287,79 @@ async def query(self, filter_: AnalysisFilter) -> List[AnalysisInternal]:
...
+            if file_id:
+                # Building OutputFileInternal object with secondary files if available
+                file_internal = OutputFileInternal.from_db(**dict(row))
+                # If no json_structure, just set to the output.
+
+                if row['json_structure'] is None:
+                    analysis_files[row['analysis_id']]['output'] = row['path']
+                else:
+                    analysis_files[row['analysis_id']]['output'] = ''
+                analysis_files[row['analysis_id']]['outputs'].append((file_internal, row['json_structure']))  # type: ignore [union-attr]
```
dancoates, line 319: This line goes over 88 cols for me, might be something up with the formatter?

db/python/tables/analysis.py
```diff
@@ -270,6 +287,79 @@ async def query(self, filter_: AnalysisFilter) -> List[AnalysisInternal]:
...
+            a_id: {
+                'output': files['output'],
+                'outputs': OutputFileInternal.reconstruct_json(files['outputs']),
+            }
+            for a_id, files in analysis_files.items()
+        }
+
+        return analysis_output_files  # type: ignore [return-value]
+
+    async def get_secondary_files_for_file_output(
```
dancoates, line 336: This method isn't called anywhere, is it needed for something in the future? Or would it be ok to remove?

db/python/tables/analysis.py
```diff
@@ -255,8 +248,32 @@ async def query(self, filter_: AnalysisFilter) -> List[AnalysisInternal]:
...
         for row in rows:
-            analysis = AnalysisInternal.from_db(**dict(row))
+
+            analysis_data = dict(row)
+            analysis_output_for_id = analysis_outputs_by_aid.get(
+                analysis_data['id'], None
+            )
+            analysis_data['output'] = (
+                analysis_output_for_id.get('output', None)
+                if analysis_output_for_id
```
dancoates, line 267: The AnalysisInternal model from_db method sets the defaults for these, so you could get away with a single 

```python
if analysis_output_for_id:
```

statement eg. 

```python

if analysis_output_for_id:
    analysis_data['output'] = analysis_output_for_id.get('output', None)
    analysis_data['outputs'] = analysis_output_for_id.get('outputs', {})

```

db/python/tables/analysis.py
```diff
@@ -308,12 +398,29 @@ async def get_latest_complete_analysis_for_type(
         rows = await self.connection.fetch_all(_query, values)
         if len(rows) == 0:
             raise NotFoundError(f"Couldn't find any analysis with type {analysis_type}")
-        a = AnalysisInternal.from_db(**dict(rows[0]))
+        latest_analysis = rows[0]
+        latest_analysis_data = dict(latest_analysis)
+        analysis_outputs_by_aid = await self.get_file_outputs_by_analysis_ids(
```
dancoates, line 403: This bit of code to attach the outputs to the analysis is repeated a few times in this file, would be good to break it out into a function to reduce the repetition.

db/python/tables/analysis.py
```diff
@@ -395,11 +524,27 @@ async def get_analysis_by_id(
 
         project = rows[0]['project']
 
-        a = AnalysisInternal.from_db(**dict(rows[0]))
+        analysis_data = dict(rows[0])
+        analysis_outputs_by_aid = await self.get_file_outputs_by_analysis_ids(
+            [analysis_data['id']]
+        )
+
+        # TODO: Instead of fetching the same value for both, output should have a str or None, and outputs should have a dict or None
```
dancoates, line 532: I think this todo is todone? Though I might be reading it wrong.

db/python/tables/output_file.py
```diff
@@ -0,0 +1,266 @@
...
+        cs_keys = ', '.join(keys)
+        cs_id_keys = ', '.join(f':{k}' for k in keys)
+        non_pk_keys = [k for k in keys if k != 'path']
+        update_clause = ', '.join(
+            [f'{k} = VALUES({k})' for k in non_pk_keys]
+        )  # ON DUPLICATE KEY UPDATE {update_clause}
+
+        async with self.connection.transaction():
+            _query = dedent(
+                f"""INSERT INTO output_file ({cs_keys}) VALUES ({cs_id_keys}) ON DUPLICATE KEY UPDATE {update_clause} RETURNING id"""
```
dancoates, line 97: can run this on two lines to keep within the line length limit

db/python/tables/output_file.py
```diff
@@ -0,0 +1,266 @@
...
+        ]
+
+        kv_pairs = [(k, v) for k, v in kv_pairs if v is not None]
+        keys = [k for k, _ in kv_pairs]
+        cs_keys = ', '.join(keys)
+        cs_id_keys = ', '.join(f':{k}' for k in keys)
+        non_pk_keys = [k for k in keys if k != 'path']
+        update_clause = ', '.join(
+            [f'{k} = VALUES({k})' for k in non_pk_keys]
+        )  # ON DUPLICATE KEY UPDATE {update_clause}
```
dancoates, line 93: Don't need this comment, it is added below

db/python/tables/output_file.py
```diff
@@ -0,0 +1,266 @@
...
+
+        async with self.connection.transaction():
```
dancoates, line 95: No need for a transaction if there's only one query

dancoates: Oh and this is already happening in a transaction which is started in the `create_or_update_analysis_output_files_from_output` method, meaning this would be creating checkpoints which I don't think are needed

db/python/tables/output_file.py
```diff
@@ -0,0 +1,266 @@
...
+        _update_query = dedent(
+            """
+            DELETE ao FROM analysis_outputs ao
+            WHERE (ao.analysis_id = :analysis_id)
+            AND (ao.file_id NOT IN :file_ids)
+            """
+        )
+
+        if file_ids:
+            await self.connection.execute(
```
dancoates, line 207: I think it'd be good if this was in the same transaction as the create/update, otherwise the updates won't be rolled back if the deletion fails

dancoates (changes requested):
This is looking solid @nevoodoo, most of my notes are pretty small things. Happy to chat about any of them if you'd like.

models/models/output_file.py
```diff
@@ -0,0 +1,216 @@
...
+            valid = False
+            size = 0
+            # try:
+            if isinstance(file_obj, GSPath) and client:
+                if '/' in file_obj.blob:
+                    prefix = '/'.join(file_obj.blob.split('/')[:-1]) + '/'
+                else:
+                    prefix = ''
+                delimiter = '/'
+                blobs = client.list_blobs(
```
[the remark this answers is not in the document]
nevoodoo: I just wrapped it with this utility: `from fastapi.concurrency import run_in_threadpool` should be good now :)

db/python/tables/output_file.py
```diff
@@ -0,0 +1,266 @@
...
+        analysis_id: int,
+        json_dict: RecursiveDict | str,
+    ) -> None:
+        """
+        Create analysis files from JSON
+        """
+        files = await self.find_files_from_dict(json_dict=json_dict)
+        file_ids: list[int] = []
+
+        if os.environ.get('SM_ENVIRONMENT', 'local').lower() in (
```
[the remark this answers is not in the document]
nevoodoo: There seems to be a weird behaviour with this in tests because despite having `os.environ['SM_ENVIRONMENT'] = 'test'` in `testbase.py`, the value returned from `api.settings` is `production` and this causes my tests to fail. I've left this as is but worth investigating later on

models/models/output_file.py
nevoodoo: Good point, I've patched this now!

db/python/tables/analysis.py
nevoodoo: Hmmm, I ran `black` manually on this file at 79 chars and it still refused to format it.

models/models/output_file.py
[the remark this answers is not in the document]
nevoodoo: It could, but I use the values from the dict to just write to the database, and so it seemed pointless creating an OutputFile object. Is there any benefit from making this change otherwise?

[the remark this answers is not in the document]
dancoates: I think so, at the moment return type of this function is unknown, so if you were to change the code here to change what keys the dict returns the type checker wouldn't pick up that the code in `create_or_update_output_file` needs to be updated. The best thing about type safety is that you can know the effect area that your code changes have.

db/python/tables/analysis.py
nevoodoo: Good point!

db/python/tables/analysis.py
nevoodoo: I will address this in a separate PR, as I noticed a lot of code in this file was reusable so it might be worth addressing it all at once :) Super useful suggestion, thanks!

models/models/output_file.py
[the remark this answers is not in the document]
nevoodoo: Makes sense, thanks for explaining!

dancoates (approved):
This is looking good to me, nice stuff!

test/testbase.py
```diff
@@ -153,7 +156,10 @@ async def setup():
                     *('--defaultsFile', db_prefix + '/liquibase.properties'),
                     *('--url', lcon_string),
                     *('--driver', 'org.mariadb.jdbc.Driver'),
-                    *('--classpath', db_prefix + '/mariadb-java-client-3.0.3.jar'),
+                    *(
+                        '--classpath',
+                        db_prefix + '/mariadb-java-client-3.0.3.jar',
+                    ),
```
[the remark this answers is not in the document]
nevoodoo: Quite possible :) I think I'll leave it for now, I'm hoping once I've setup a fresh environment, this should be sorted.

PR #860, opened 2024-07-08
Add new BBV fields to the dashboard

nevoodoo:
Just need to amend one field:

`collection-event-name` changes to `collection-event-type`.

We already had `processing-site` as is, so we don't need to modify this.


milo-hyben (approved):
Looks good to me!

PR #861, opened 2024-07-08
Fix OpenApi type generation for project roles

dancoates:
Looks like pydantic/fastapi don't correctly handle usage of field serializers for int based enums being converted to strings, this switches to using a StrEnum instead.


milo-hyben (approved):
Looking good!

PR #863, opened 2024-07-09
OurDNA Dashboard Enhancements

nevoodoo:


db/python/layers/ourdna/dashboard.py
```diff
@@ -80,6 +80,11 @@ def processing_time_by_site(self) -> tuple[str | None, int | None]:
         """Get processing times and site for a sample."""
         return self.get_property('processing-site'), self.processing_time
 
+    @cached_property
+    def processing_time_by_collection_site(self) -> tuple[str | None, int | None]:
+        """Get processing times and site for a sample."""
+        return self.get_property('collection-lab'), self.processing_time
```
milo-hyben, line 86: There are 2 attributes 'collection-lab' and 'collection-site', get_property call is not matching the function/property name, a bit confusing :-)

milo-hyben (commented):
Just a small thing regarding property name, otherwise looking good!

db/python/layers/ourdna/dashboard.py
nevoodoo: Just renamed the function name :) Good spot!

milo-hyben (approved):
Looking good!

PR #865, opened 2024-07-09
Update css for project grid table layout

EddieLF:
In the project insights PR #533 a piece of css was added which impacted semantic-ui Table components. The css set `table-layout: fixed`, meaning all Table columns are rendered to fit on the users screen, compressing all the data cells too. 

This fix sets the layout to `auto`, allowing the columns to relax and use the width they need so the grid can actually be read.


illusional (approved):
🙏

PR #866, opened 2024-07-10
Multiple external IDs in GraphQL and other front-end code

jmarshall:
Leaves the `externalId` field as is, for existing queries and those who just want to query the primary one.


illusional (approved):
Awesome, thanks @jmarshall!

PR #868, opened 2024-07-11
Get OpenAPI Spec without spinning up server

illusional:
Great shout @dancoates .

This technically allows us to unbundle our 3.1 -> 3.0 OpenAPI schema version hack, and code has been set-up to support that, we basically just need to check that the Swagger pages (both FastAPI / node version) support 3.1 before doing this.

If you hide whitespace changes, this is mostly just deletes.


nevoodoo (approved):
Breeeezy

PR #870, opened 2024-07-15
Create or validate md5s script

EddieLF:
Removes the existing (and broken) `check_md5s.py` script, and adds a way to check md5s with the existing `create_md5s.py` script, now renamed to `create_or_validate_md5s.py`

Script will now create md5s for files in the dir that don't have one, and either validate those that do have md5s, or overwrite old md5s with new md5s if the `--force-recreate-existing` flag is present.


scripts/create_or_validate_md5s.py
```diff
@@ -0,0 +1,155 @@
...
+    validate_files: bool,
+    billing_project: str,
+    driver_image: str,
+    storage_client: storage.Client
+):
+    """
+    Create MD5s for files in the provided gs directory, skipping files with certain extensions.
+    If the MD5 already exists, validate it instead, unless force_recreate is set.
+    """
+    b = get_batch(f'Create or validate md5 checksums for files in {gs_dir}')
```
illusional, line 80: Nitpick, but set the create / validate based on what's asked.

scripts/create_or_validate_md5s.py
```diff
@@ -0,0 +1,155 @@
...
+    files = get_blobs_in_directory(gs_dir, billing_project, storage_client)
+    for filepath in files:
+        if filepath.endswith('.md5') or filepath.endswith(skip_filetypes):
+            continue
+        if f'{filepath}.md5' in files:
+            if force_recreate:
+                print('Re-creating md5 for', filepath)
+                job = b.new_job(f'Create {os.path.basename(filepath)}.md5')
+                create_md5(job, filepath, billing_project, driver_image)
+            elif validate_files:
```
illusional, line 91:
```suggestion
        if f'{filepath}.md5' in files and not force_recreate:
            if validate_files:
```

illusional (changes requested):
The structure looks pretty good! The default is to always create MD5s for missing files, even in a validation mode, which is a surprising default to me. 

I don't know if I have a better suggestion, but maybe requiring the user to see what they want to do?

1. `mode=validate`
2. `mode=create` (with an optional `--force-recreate`)

EddieLF:
@illusional thanks for the feedback. I've implemented the `mode` based on if a user sets `--validate-existing-only` or if they set `--create-only`. 

Originally I thought it might be nice if we created MD5s where they didn't exist and validated them where they did - all in a single pass. 
But it probably makes more sense to split this out into two possible options. Either create the md5s OR validate them, don't try and do both in one run.

scripts/create_or_validate_md5s.py
```diff
@@ -0,0 +1,159 @@
...
+        raise ValueError(f'Expected GS directory, got: {gs_dir}')
+
+    if not billing_project:
+        billing_project = config_retrieve(['workflow', 'dataset_gcp_project'])
+    driver_image = config_retrieve(['workflow', 'driver_image'])
+
+    if validate_existing_only and create_only:
+        raise ValueError('Only one of --validate-existing-only or --create-only can be set.')
+
+    mode = 'validate' if validate_existing_only else 'create'
```
illusional, line 142: Directly ask the user for the mode as a click argument with only two choices (`create` / `validate`), rather than inferring it through a set of command line options that are unbound.

scripts/create_or_validate_md5s.py
```diff
@@ -0,0 +1,159 @@
...
+
+    storage_client = storage.Client()
+
+    create_and_validate_md5s_for_files_in_directory(
+        gs_dir,
```
illusional, line 147: Nitpick: use the keyword=value param here to stability, especially if the order of arguments change. It makes it easier to review too.

scripts/create_or_validate_md5s.py
```diff
@@ -0,0 +1,159 @@
...
+    Streams the file with gsutil and calculates the md5 checksum,
+    then compares it to the stored checksum in the .md5 file.
+    """
+    copy_common_env(job)
+    job.image(driver_image)
+    md5_filepath = f'{filepath}.md5'
+    job.command(
+        f"""\
+    set -euxo pipefail
+    gcloud -q auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS
```
illusional, line 48: In unrelated news, you don't actually need to do this anymore with a recent Hail Batch update.

scripts/create_or_validate_md5s.py
```diff
@@ -0,0 +1,159 @@
...
+    calculated_md5=$(gsutil -u {billing_project} cat {filepath} | md5sum | cut -d " " -f1)
```
illusional, line 49: Nitpicky / petty, but good to use shlex.quote these params, ie: if a user provides the billing_project `"help); gsutil rm -rf gs://bucket; calculated_md5=$(gsutil -u seqr"`, this could cause a lot of problems.

illusional (commented):
I think the current flow makes this feel a bit confusing with a set of continues, you can make this easier.

```
if create:
    if force or not exists:
        create
if validate:
    if exists:
        validate
```

illusional (approved):
Thanks for making those changes!

PR #872, opened 2024-07-15
When updating Analysis record only set timestamp_completed value if it's not already set.

milo-hyben:
This PR is fixing the issue with timestamp_completed being updated even if it has been set prior.
This is causing issues with seqr aggregation as it can not find any completed cram types prior to May 18, 2022
With more details reported in the close PR:
https://github.com/populationgenomics/metamist/pull/871

There are 5614 records with wrong timestamp_completed.

There are 3 steps to fix this issue:
1. Apply this PR
2. Backup the records to be fixed with this sql:

```
CREATE TABLE analysis_backup_prior_timestamp_completed_update AS
SELECT analysis.*
FROM analysis
INNER JOIN (
SELECT id, MIN(timestamp_completed) as timestamp_completed 
FROM analysis FOR SYSTEM_TIME ALL 
WHERE timestamp_completed IS NOT NULL
GROUP BY id
) as s
on s.id = analysis.id AND analysis.timestamp_completed <> s.timestamp_completed;
```

3. Update existing records with the sql:

```
CREATE TABLE tmp_analysis_new_timestamp_completed AS
SELECT s.id, s.timestamp_completed as new_timestamp_completed
FROM analysis a
INNER JOIN (
SELECT id, MIN(timestamp_completed) as timestamp_completed 
FROM analysis FOR SYSTEM_TIME ALL 
WHERE timestamp_completed IS NOT NULL
GROUP BY id
) as s
on s.id = a.id AND a.timestamp_completed <> s.timestamp_completed;

UPDATE analysis
INNER JOIN tmp_analysis_new_timestamp_completed s
on s.id = analysis.id AND analysis.timestamp_completed <> s.new_timestamp_completed
SET analysis.timestamp_completed = s.new_timestamp_completed
WHERE analysis.id in (SELECT id from tmp_analysis_new_timestamp_completed);


```

### DIFF
--- a/db/python/tables/sequencing_group.py
+++ b/db/python/tables/sequencing_group.py
@@ -343,7 +343,7 @@ class SequencingGroupTable(DbBase):
         _query = """
         SELECT sg.id, min(s.row_start)
         FROM sequencing_group sg
-        INNER JOIN sample s ON s.id = sg.sample_id
+        INNER JOIN sample FOR SYSTEM_TIME ALL s ON s.id = sg.sample_id
         WHERE sg.id in :sgids
         GROUP BY sg.id
         """
--- a/db/python/tables/analysis.py
+++ b/db/python/tables/analysis.py
@@ -187,7 +187,10 @@ VALUES ({cs_id_keys}) RETURNING id;"""
 
         if status == AnalysisStatus.COMPLETED:
             fields['timestamp_completed'] = datetime.datetime.utcnow()
-            setters.append('timestamp_completed = :timestamp_completed')
+            # only set timestamp_completed if it's not already set
+            setters.append(
+                'timestamp_completed = CASE WHEN timestamp_completed IS NULL THEN :timestamp_completed ELSE timestamp_completed END',
+            )
 
         if output:
             fields['output'] = output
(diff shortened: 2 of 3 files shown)


db/python/tables/sequencing_group.py
```diff
@@ -343,7 +343,7 @@ async def get_samples_create_date_from_sgs(
         _query = """
         SELECT sg.id, min(s.row_start)
         FROM sequencing_group sg
-        INNER JOIN sample s ON s.id = sg.sample_id
+        INNER JOIN sample FOR SYSTEM_TIME ALL s ON s.id = sg.sample_id
```
milo-hyben, line 346: Hey @illusional I am still getting 'AssertionError: No ratio found for date 2022-01-19' for SAMPLE_CREATE_DATE after updating DB with the correct time_completed.
I reckon this is the corporate, can't really test unless we deploy. Without 'FOR SYSTEM_TIME ALL' it only return last updated timestamp and not create_date as function name suggest.

illusional: Oh yep, sounds good, thanks @milo-hyben!
Let's give deploying a crack. 

I think this is probably the true fix.

illusional (approved):

PR #875, opened 2024-07-18
Custom-cohort creation script patch

MattWellie:
### Needed

- dry-run doesn't work. The boolean is captured appropriately, but when sent to `create_cohort_from_criteria` it's ignored. **This PR has not resolved that**
- project, name, description made mandatory (?)

### QOL
- projects changed from `type=str, nargs='*'` to a `nargs='*', default=[]` - the type here was misleading as we want a list?
- multiple args changed from `required=False, type=list[str]` to `nargs='*', default=[]`
   - these are all lists of strings with none-to-many elements (though typically 1)
   - if the flag is not used, the value becomes an empty list
   - argparse's `type` doesn't work with nested types (afaik, the passed type is used as a callable), so `list[str]` is just `list`
   - argparse's `type` also casts values to the given type, so `--sg_platform illumina` becomes `['i', 'l', 'l', 'u', 'm', 'i', 'n', 'a']`, instead of `['illumina']` (the result of `list(input)` instead of `[input]`

### Strict Checking
- where there's a strict enumeration of valid values in metamist I've applied these using `choices` - could bin this off, or query metamist for the available choices at runtime.

Challenge: 
```bash
python scripts/create_custom_cohort.py --project matt --name MATT --description "Matt desc" --projects matt not_matt --sg_technology short-read --sg_platform illumina --sg_type genome --sg_ids_internal CPGA --sample_type blood
```

Current Result: (I couldn't find any way to use multiple CPG IDs, so just using a list of one in this toy example)
```txt
{'project': 'matt', 'name': 'MATT', 'description': 'Matt desc', 'template_id': None, 'projects': ['matt', 'not_matt'], 'sg_ids_internal': ['C', 'P', 'G', 'A'], 'excluded_sgs_internal': None, 'sg_technology': ['s', 'h', 'o', 'r', 't', '-', 'r', 'e', 'a', 'd'], 'sg_platform': ['i', 'l', 'l', 'u', 'm', 'i', 'n', 'a'], 'sg_type': ['g', 'e', 'n', 'o', 'm', 'e'], 'sample_type': ['b', 'l', 'o', 'o', 'd'], 'dry_run': False}
```

New Result:
```txt
{'project': 'matt', 'name': 'MATT', 'description': 'Matt desc', 'template_id': None, 'projects': ['matt', 'not_matt'], 'sg_ids_internal': ['CPGA'], 'excluded_sgs_internal': [], 'sg_technology': ['short-read'], 'sg_platform': ['illumina'], 'sg_type': ['genome'], 'sample_type': ['blood'], 'dry_run': False}
```


vivbak (commented):
Huge thanks for this @MattWellie! 
I'm just trying to think of a simple way to make codecov happy, presumably by testing that the parser is working correctly (?) 

Not a super clear part of the code to test, to me.

MattWellie:
> I'm just trying to think of a simple way to make codecov happy, presumably by testing that the parser is working correctly (?)

Yeah, there's a couple of ways to solve that. I'll have a tinker this morning!

PR #882, opened 2024-07-31
Move to use query syntax for multiple query params

illusional:
Fix the query params on family endpoint + other endpoints

Docs: https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#query-parameter-list-multiple-values


api/routes/analysis.py
```diff
@@ -238,16 +238,16 @@ async def query_analyses(
 
 @router.get('/analysis-runner', operation_id='getAnalysisRunnerLog')
 async def get_analysis_runner_log(
-    project_names: list[str] = Query(None),  # type: ignore
+    project_names: Annotated[list[str], Query()] = [],  # type: ignore
```
jmarshall, line 241: Can we remove `type: ignore` here?

jmarshall (changes requested):
It would be good to have a link to how this annotation changes the fastapi encoding and solves the problem.

The linters are complaining that the `= []` default argument is the Python problem of a shared mutable default argument… so that probably needs a dummy None default that we change to `[]` or an annotation comment to shut it up…

api/routes/analysis.py
illusional: Switched to `noqa` due to the mutable list thing. I added a doc link above

jmarshall (approved):

PR #885, opened 2024-08-02
Add participant filtering guide + UX improvements

illusional:
- add a `readonly` option to the CodeEditor
- I've verified that the user guide is at least partly understable
- Added `Select all | Select none` options
- Split non-meta and meta fields in ProjectColumnOptions
- Added dark mode support to the modal


nevoodoo (approved):
LGTM!

PR #886, opened 2024-08-06
[test] Silence port errors + simplify test db construction

illusional:
~~I don't know why this isn't working ...~~
Looks like it more just accidentally always worked.

This PR:
- When searching for a free port, explicitly disconnect before giving to the database (rather than the accidental garbage collection free up)
- Reduces the number of containers used, by creating multiple databases on the same MariaDB instance.

### DIFF
--- a/db/python/connect.py
+++ b/db/python/connect.py
@@ -369,9 +369,7 @@ class SMConnections:
             if log_database_queries is not None
             else LOG_DATABASE_QUERIES
         )
-        return databases.Database(
-            config.get_connection_string(), echo=LOG_DATABASE_QUERIES
-        )
+        return databases.Database(config.get_connection_string(), echo=_should_log)
 
     @staticmethod
     async def connect():
--- a/db/python/connect.py
+++ b/db/python/connect.py
@@ -359,9 +359,16 @@ class SMConnections:
         return SMConnections._credentials
 
     @staticmethod
-    def make_connection(config: DatabaseConfiguration):
+    def make_connection(
+        config: DatabaseConfiguration, log_database_queries: bool | None = None
+    ):
         """Create connection from dbname"""
         # the connection string will prepare pooling automatically
+        _should_log = (
+            log_database_queries
+            if log_database_queries is not None
+            else LOG_DATABASE_QUERIES
+        )
         return databases.Database(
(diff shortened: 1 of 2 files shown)


db/python/connect.py
```diff
@@ -359,9 +359,16 @@ def _get_config():
         return SMConnections._credentials
 
     @staticmethod
-    def make_connection(config: DatabaseConfiguration):
+    def make_connection(
+        config: DatabaseConfiguration, log_database_queries: bool | None = None
+    ):
         """Create connection from dbname"""
         # the connection string will prepare pooling automatically
+        _should_log = (
```
dancoates, line 367: I think this is intended to be passed to the `echo` param below

jmarshall:
I think the problem here is that `find_free_port()` creates a socket that it doesn't use at all, just to get an otherwise unused port. It then leaks the local variable `s` without closing it, leading to the warning (which gets printed immediately, before the next line at the call site, `db.with_bind_ports(…)` — apparently this garbage collection ~~is~~ can be instantaneous!).

So the following would fix it:

```diff
@@ -51,7 +51,9 @@ def find_free_port():
     """Find free port to run tests on"""
     s = socket.socket()
     s.bind(('', 0))  # Bind to a free port provided by the host.
-    return s.getsockname()[1]  # Return the port number assigned.
+    port = s.getsockname()[1]
+    s.close()
+    return port  # Return the port number assigned.
```

(Or maybe some `try` … `finally` construct would be more elegant.)

Or ideally one could nuke `find_free_port()` entirely, tell the `db` to choose a port to bind to (maybe via `db.with_bind_ports(db.port, 0)`? I couldn't find much documentation…), and then ask it what port it got via some other API in the documentation I didn't find yet…

dancoates (approved):
nice! just one thing to fix, otherwise should be a nice improvement

illusional:
Thanks @jmarshall! Funny that it's accidentally worked for the past 2 years - that creating the connection, then accidentally leaking it causing it to be immediately freed up to be used. I've reinstated that function, claim the port, and then immediately release it port to be used by the testcontainers database.

This PR also reduces the number of mysql containers spun up by the test suite (going for one _database_ in MariaDB per test class).

PR #887, opened 2024-08-06
Lock openapi schema version for swagger UI to 3.0.2

dancoates:
Not quite sure why this suddenly broke, but this seems to fix it

### DIFF
--- a/api/utils/openapi.py
+++ b/api/utils/openapi.py
@@ -78,7 +78,7 @@ def get_openapi_schema_func(app, version):
             version=version,
             routes=app.routes,
             # update when FastAPI + swagger supports 3.1.0
-            # openapi_version='3.1.0'
+            openapi_version='3.0.2',
         )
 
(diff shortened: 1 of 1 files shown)


illusional (commented):
I thought swagger UI would support 3.1.0 fine? Or if it's the react flavour, should update swagger on the front-end.

The actual openapi generator should not use this function anymore.

dancoates:
Yeah fair point, I was tossing up whether to upgrade the swagger UI, it's a major version upgrade so could break things but I'll give it a shot

dancoates:
@illusional I had a crack at updating to the new swagger ui react to support openapi def 3.1.0, but it is a bit of a rabbit hole, v5 of swagger UI requires v9 of react-redux which requires v18 of react, and upgrading to v18 of react requires updating most of our component libraries, I'm not even sure all of them will support react 18. I think it'd be good to get temporary fix out and look at upgrading react separately.

illusional (approved):
Ooft, what a rabbit hole.

PR #892, opened 2024-08-12
Front end dependency cleanup

dancoates:
- Remove unused dependencies
- Move dev dependencies to dev dependency section
- Fix lodash imports so that the whole library isn't included in bundle (reduces bundle size by 500kb)


illusional (approved):
👌

PR #893, opened 2024-08-12
Nested sample fixes

illusional:
Participant ID not set on updates, which is causing samples to disappear on project grid


silkm:
Looks good, nice catch

silkm (approved):

PR #895, opened 2024-08-12
Use ruff format instead of black

illusional:
Same treatment as cpg-infra, reduce future instability between ruff format and check, and black. Simplify dev setups.
Include formatting files inside the metamist directory, and exclude specific subfiles, instead of the previous blanket exclude.


dancoates (approved):
Nice! Excited for how much faster this'll be

PR #896, opened 2024-08-12
Multiple external IDs for families

jmarshall:


db/python/tables/family.py
```diff
@@ -167,14 +179,60 @@ async def get_family_external_ids_by_participant_ids(
...
+
+            if to_update:
+                _query = 'SELECT project FROM family WHERE id = :id'
+                row = await self.connection.fetch_one(_query, {'id': id_})
+                project = row['project']
+
+                _update_query = """
+INSERT INTO family_external_id (project, family_id, name, external_id, audit_log_id)
+VALUES (:project, :id, :name, :external_id, :audit_log_id)
+ON DUPLICATE KEY UPDATE external_id = :external_id, audit_log_id = :audit_log_id
```
dancoates, line 222: I think this will update the audit log id even if the `external_id` isn't changed. It could be worth making that update conditional so that it only gets updated if it is different, ie.

```sql
ON DUPLICATE KEY UPDATE
    audit_log_id = IF(
        external_id = VALUES(external_id),
        audit_log_id, -- Return the existing value if unchanged
        VALUES(audit_log_id) -- Return the new value.
    ),
    external_id = IF(
        external_id = VALUES(external_id),
        external_id,
        VALUES(external_id)
    )
```

I haven't tested if this works, but I think should be ok in theory - what I'm not sure about is whether it'll still create a entry in the system versioned tables if the `IF` statement returns the exact existing value. Even if it does I think it'd still be good cos the audit_log_id wouldn't be updated.

dancoates: Alternatively could filter out unchanged external ids before doing the update

db/python/tables/family.py
```diff
@@ -167,14 +179,60 @@ async def get_family_external_ids_by_participant_ids(
...
+            to_update = {k.lower(): v for k, v in external_ids.items() if v is not None}
+
+            if to_delete:
+                # Set audit_log_id to this transaction before deleting the rows
+                _audit_update_query = """
+UPDATE family_external_id
+SET audit_log_id = :audit_log_id
+WHERE family_id = :id AND name IN :names
+                """
+                await self.connection.execute(
```
dancoates, line 201: Should all these queries be happening in a transaction?

db/python/tables/family.py
```diff
@@ -217,7 +275,25 @@ async def create_family(
 RETURNING id
         """
 
-        return await self.connection.fetch_val(_query, updater)
+        new_id = await self.connection.fetch_val(_query, updater)
```
dancoates, line 278: I think this and the external id creation should happen in a transaction

db/python/tables/family.py
```diff
@@ -288,19 +397,20 @@ async def get_id_map_by_external_ids(
     async def get_id_map_by_internal_ids(
         self, family_ids: List[int], allow_missing=False
     ):
-        """Get map of {external_id: internal_id} for a family"""
+        """Get map of {internal_id: external_id} for a family"""
         if len(family_ids) == 0:
             return {}
-        _query = 'SELECT id, external_id FROM family WHERE id in :ids'
+        # FIXME needs to be uniqueified!!
```
dancoates, line 403: Does this FIXME still need to be FIXME'd?

dancoates (changes requested):
This is looking good, just a few comments/questions

db/python/tables/family.py
jmarshall: Yes indeed!

db/project.xml
```diff
@@ -1606,4 +1606,81 @@
...
+
+		<sql>ALTER TABLE family_external_id ADD SYSTEM VERSIONING;</sql>
+
+		<!-- Migrate existing external_ids to the new tables, keyed by PRIMARY_EXTERNAL_ORG, i.e. '' -->
+		<sql>INSERT INTO audit_log (author, on_behalf_of, ar_guid, comment, auth_project, meta)
+		     VALUES ('liquibase', NULL, NULL, 'family external_id migration', NULL, NULL)
+		     RETURNING @audit_log_id := id;
+
+		     INSERT INTO family_external_id (project, family_id, name, external_id, audit_log_id)
+		       SELECT project, id, '', external_id, @audit_log_id
```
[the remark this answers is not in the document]
jmarshall: I was pretty sure there had to be some way to do it, but it took quite a while to [find MariaDB's incantation](https://mariadb.com/kb/en/user-defined-variables/) for it!

db/python/tables/family.py
```diff
@@ -167,14 +179,60 @@ async def get_family_external_ids_by_participant_ids(
...
+        values: Dict[str, Any] = {'audit_log_id': audit_log_id}
+
+        if external_ids:
+            to_delete = [k.lower() for k, v in external_ids.items() if v is None]
+            to_update = {k.lower(): v for k, v in external_ids.items() if v is not None}
+
+            if to_delete:
+                # Set audit_log_id to this transaction before deleting the rows
+                _audit_update_query = """
+UPDATE family_external_id
```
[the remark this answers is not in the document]
jmarshall: :shrug: There's a bit of a mixture of at-the-left-margin and inline-with-surrounding-code SQL in these files.

But mostly I was being lazy and trying to avoid excess line-wrapping… will revisit these and look at what's in the nearby code.

db/python/tables/family.py
jmarshall: Good point

db/python/tables/family.py
jmarshall: Technically probably yes

db/python/tables/family.py
```diff
@@ -167,14 +179,60 @@ async def get_family_external_ids_by_participant_ids(
...
+SET audit_log_id = :audit_log_id
+WHERE family_id = :id AND name IN :names
+                """
+                await self.connection.execute(
+                    _audit_update_query,
+                    {'id': id_, 'names': to_delete, 'audit_log_id': audit_log_id},
+                )
+
+                _delete_query = """
+DELETE FROM family_exernal_id
```
jmarshall, line 207: Turns out I didn't have test coverage on this function 🤣 

```suggestion
DELETE FROM family_external_id
```

db/python/tables/family.py
jmarshall: I checked the two places where this function is invoked, which turn out to mostly just want a single external ID for use in a pedigree or similar. Hence the best approach, at least in the short term, is indeed to uniqueify this by returning only the **primary** external ids — which is now implemented.

db/python/tables/family.py
jmarshall: After consideration, I'm going to call this a **feature** :smile:

When updating, you have the option of leaving items out of `external_ids` if you don't want them changed. So if you do specify `'foo': 'bar'` even though its value is already `'bar'`, I'm going to call it a feature that that causes this record to be bumped in the audit log, and `foo`'s value to be “freshened”.

jmarshall:
(The _db/project.xml_ conflict is trivial — just need to concatenate the changesets from both sides — and will be resolved during merge.)

db/python/tables/family.py
dancoates: No worries, sounds fine. I think there's plenty of places where the audit log is a tad noisier than it could be so we'll probably need to implement some de-duplication if we ever want to display in the UI/API anyway

dancoates (approved):
Changes look good! happy for this to be merged from my perspective

PR #897, opened 2024-08-13
Better ETL accessors 

illusional:
- (soon to action, analysis group)


milo-hyben (approved):
ETL extract which used by BBV has no code change, so merge or deployment should not have any impact on BBV team.
Would be nice if we can test these changes outside the production :-)

PR #898, opened 2024-08-13
Metamist Comments Backend

dancoates:
Implementation of comments for metamist.
See RFC for details: https://docs.google.com/document/d/1S_LFxLHfVMGPL7mJ3O6ozPjSDWMkWtlk5QX0wIcOAq4/edit?usp=sharing


api/graphql/schema.py
```diff
@@ -90,7 +107,7 @@ async def m(info: Info[GraphQLContext, 'Query']) -> list[str]:
 
 GraphQLEnum = strawberry.type(type('GraphQLEnum', (object,), enum_methods))
 
-GraphQLAnalysisStatus = strawberry.enum(AnalysisStatus)
+GraphQLAnalysisStatus = strawberry.enum(AnalysisStatus)  # type: ignore
```
ckrishk, line 110: nit: does this mean the type is ignored?

db/python/tables/comment.py
```diff
@@ -0,0 +1,551 @@
...
+            WITH root_comment as (
+                SELECT COALESCE(parent_id, id) as comment_id
+                FROM comment
+                WHERE id = :comment_id
+            ) {join_table_query}
+        """
+
+        rows = await self.connection.fetch_all(query, {'comment_id': comment_id})
+
+        if len(rows) == 0:
```
ckrishk, line 427: We can create ticket in the backlog to add tests that codecov is suggesting

db/python/layers/family.py
```diff
@@ -338,10 +338,6 @@ async def import_pedigree(
 
         return True
 
-    async def update_family_members(self, rows: list[PedRowInternal]):
```
ckrishk, line 341: is this change expected?

api/graphql/schema.py
dancoates: yeah this was a weird one, I didn't touch this line but it started erroring out in mypy, the ignore just silences the type error on this line but the type checking still works elsewhere

ckrishk (approved):

nevoodoo (approved):

jmarshall:
Do we need to DELETE FROM these tables in _db/python/tables/project.py_'s `delete_project_data()`?

dancoates:
> Do we need to DELETE FROM these tables in _db/python/tables/project.py_'s `delete_project_data()`?

Ooh that's a good point, I think we might need to rethink those project deletes a bit – I suspect it is currently broken as we missed that in the first class files PR too, so the `analysis_outputs` table FK would be breaking deletes. 

That probably also means that our test coverage for that route isn't good. I'll add a issue to the tracker to come back around and see if we can improve tests and make it a bit harder to miss adding tables to the project deletes.

Issue #795, opened 2024-05-22
Analysis update endpoint doesn't allow updating `active`


dancoates:
The `AnalysisUpdateModel` specifies that active is a valid param to pass through but it isn't sent on to the layer or table methods:

https://github.com/populationgenomics/metamist/blob/dev/api/routes/analysis.py#L136

dancoates:
This was fixed by #899

PR #903, opened 2024-08-20
Fix reading analysis.active database field and add test case

jmarshall:
This is the final (?) part of the #795 / #899 mystery: even after deploying that PR, the `active` field was always being shown as True in GraphQL.

This field is reported by the database as `b'\x00'` or `b'\x01'`, both of which `bool(…)` returns as True. Use `parse_sql_bool()` instead; and I've verified that the `from_db()` accessors for all other BOOLEAN or BIT fields use `parse_sql_bool()` too.
    
Also adds a test case exercising the `update_analysis()` API fixed in PR #899 and checking that active is returned from the database correctly, checking both directly and via the query API.

Also some drive-by trivia: Fix `nullIfInactive` typo so grepping finds this comment. I also verified that changing this XML comment doesn't affect liquibase's hash for this changeSet.


dancoates (approved):
oof, that sql bool thing is a real footgun, almost seems like a bug in aiomysql

jmarshall:
Indeed. Joyously compounded in this case by the field in question actually being a BIT rather than an actual BOOLEAN, for some reason. But apparently BOOLEAN is also just a synonym for TINYINT(1) in MariaDB anyway…

dancoates:
Ah that makes more sense, just did a quick test and a `BOOLEAN` column does actually convert ok with just `bool()`, it's only `BIT` columns that seem to have this problem - we should probably avoid using them - not sure what the advantage is.

PR #916, opened 2024-09-03
Fix frontend linting and formatting

dancoates:
- Fixes (or ignores in some cases) ~500 odd linting errors
- Adds the linting check as part of CI
- Removes linting and formatting from pre-commit, they weren't working anyway
- Adds organizing of imports back into prettier config


web/src/pages/billing/components/BatchGrid.tsx
```diff
@@ -344,26 +343,6 @@ const BatchCard: React.FC<{ item: AnalysisCostRecordBatch }> = ({ item }) => {
 const BatchGrid: React.FunctionComponent<{
     data: AnalysisCostRecord
 }> = ({ data }) => {
-    const [openRows, setOpenRows] = React.useState<string[]>([])
```
ckrishk, line 347: Is this change expected?

ckrishk (approved):
Looks good!!

web/src/pages/billing/components/BatchGrid.tsx
dancoates: Good question! I think this code was left over from a previous version or something, none of it was actually used anywhere.

violetbrina (approved):

PR #922, opened 2024-09-09
Disabling the SonarQube workflow temporarily

nevoodoo:
While we wait for DNS to propagate, we want to make sure we can still merge PRs, and the workflow isn't blocking


nevoodoo:
We do not need to disable the workflow anymore 🥳

PR #929, opened 2024-09-16
Update sonar-project.properties for new project iD

nevoodoo:


sonar-project.properties
```diff
@@ -1,4 +1,3 @@
-sonar.projectKey=populationgenomics_metamist_AZHVO0XpdaNSak8FrtKn
-sonar.python.version=3.11
```
dancoates, line 2: Not sure we want to remove the python version?

dancoates (approved):

sonar-project.properties
nevoodoo: Sorry seems like I selected more than I intended to remove, good spot, fixed now!

PR #930, opened 2024-09-19
Insights details update

EddieLF:
Updates the Insights dashboard logic

- Collect the entire CRAM analysis entry for each sequencing group, not just TRUE/FALSE for CRAM existence
- Display the date the CRAM analysis was created from the timestamp, not just TRUE/FALSE
- Collect analysis outputs by joining to `analysis_outputs` and `output_file` tables and coalescing the output fields
- Sort the values that can be filtered for each column when they are displayed to the user
- Update the Project and Sequencing type selection to use a dropdown list for better UX


nevoodoo (approved):
Looks good! Just wondering if there is anything worth updating in the tests for this layer?

db/python/layers/project_insights.py
```diff
@@ -528,24 +537,24 @@ async def _crams_by_project_id_and_seq_fields(
...
+    LEFT JOIN sequencing_group sg ON sg.id = asg.sequencing_group_id
+WHERE
+    a.project IN :projects
+    AND sg.type IN :sequencing_types
+    AND a.type = 'CRAM'
+    AND a.status = 'COMPLETED'
+GROUP BY
+    a.project,
+    sg.type,
+    sg.technology;
```
nevoodoo, line 557: Might be better to move this to the `tables` submodule since most SQL logic for other entities resides there and the layer usually calls the table methods? The SQL was already here so I'm neutral about moving it, just my OCD ticking :)

db/python/layers/project_insights.py
```diff
@@ -559,6 +568,72 @@ async def _crams_by_project_id_and_seq_fields(
...
+        AND a.project IN :projects
+        GROUP BY asg.sequencing_group_id
+    ) max_timestamps ON asg.sequencing_group_id = max_timestamps.sequencing_group_id
+    AND a.timestamp_completed = max_timestamps.max_timestamp
+WHERE
+    a.project IN :projects
+    AND sg.type IN :sequencing_types
+    AND a.type = 'CRAM'
+    AND a.status = 'COMPLETED';
+        """
```
nevoodoo, line 608: Same SQL related comment as above ^

EddieLF:
@nevoodoo thanks for the review. For the SQL - I agree that having raw SQL in this layer module sticks out because the rest of the db layer modules avoid this. I guess this is because the SQL here is used to create a view by combining the results from a number of queries from a number of different tables.

If the SQL were to be moved into the tables modules, the various functions would probably be distributed into both the `sequencing_group` the `analysis` table modules. But most of the queries are quite specific to the project insights dashboards, so it could get confusing. 

Maybe there could be a new module "views" or something where these kinds of queries could be saved, like:
```
metamist/db/python/tables/views/project_insights.py
```

Moving the SQL out of the layer module would make things a lot nicer and would also mean that some of the mypy/pylint disable comments can be removed.

Let's earmark this in an issue for a future PR and just get this one merged for now to fix the dashboard.

PR #931, opened 2024-09-19
Configure Dependabot

nevoodoo:
This sets up an initial iteration of dependabot that proposes version upgrades, on a schedule every midnight

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,3 +17,7 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
+    groups:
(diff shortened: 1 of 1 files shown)


milo-hyben (approved):
Looking good, just wonder if every day is not a bit of overkill and if weekly run would be sufficient.

nevoodoo:
> Looking good, just wonder if every day is not a bit of overkill and if weekly run would be sufficient.

Yeah weekly probably would be good. I was thinking of trying daily for a week to being with, to see how noisy it gets, and to help debug if there are issues at all, then adjusting accordingly to a weekly schedule :)

PR #937, opened 2024-09-19
Update dependabot schedule

nevoodoo:
This changes the dependabot schedule to weekly and enforces a limit of 10 open PRs.


milo-hyben (approved):
Looking good!

PR #943, opened 2024-09-22
Patched FCF Analysis Update

nevoodoo:
There was a missing update stub for the `outputs` dict on the `update` route. This PR attempts to fix that.


jmarshall (approved):
LGTM (although I'd like to see tests covering it, but NBD).

milo-hyben (commented):
Looks good to me.

PR #944, opened 2024-09-22
Additional Dependabot Tweaks

nevoodoo:
This PR tries to:

- Condense the updates into `minor` and `major` version updates
- Use `pip-compile` to correctly compile the requirements files based on the changes being proposed

### DIFF
--- a/.github/workflows/compile_requirements.txt
+++ b/.github/workflows/compile_requirements.txt
@@ -1,0 +1,39 @@
+name: Compile Requirements with Docker
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/compile_requirements.txt
milo-hyben, line 1: Should not be this file with extension *.yaml ?

milo-hyben (approved):

.github/workflows/compile_requirements.txt
nevoodoo: I genuinely do not know how I missed that, but it explains why my formatter didn't work for YAML 🙈

PR #949, opened 2024-09-23
Added branch ref to Dependabot

nevoodoo:
This adds a branch ref to the workflow so that Dependabot can commit changes to it. The grouping is still problematic but this final fix should hopefully resolve it.


milo-hyben (approved):
Looks good to me!

PR #957, opened 2024-09-23
Uplift code quality in metamist repo

violetbrina:
This is the beginning of the code quality uplift in metamist. This PR addresses all of the security issues, hotspots and bugs.
A separate PR will be started to reduce the tech debt in the repo.


jmarshall:
Is it possible to go back to a previous analysis so we can see the reports you're addressing?

(Or I guess we could cause it to analyse **dev** or **main** again…)

violetbrina:
> Is it possible to go back to a previous analysis so we can see the reports you're addressing?
> 
> (Or I guess we could cause it to analyse **dev** or **main** again…)

That's an excellent question! I'm looking into that to figure out the best configuration for "New Code". This should allow us to compare main (or, in the case of metamist, probably dev) to the newest code that is analysed in order to compare.

db/deploy/Dockerfile
```diff
@@ -8,7 +8,37 @@ ENV LIQUIBASE_VERSION=4.26.0
...
+
+# Add any additional commands here, such as copying application files, installing Python dependencies, etc.
+# COPY requirements.txt .
+# RUN pip install -r requirements.txt
+
+# Entry point or command to run your application
+# ENTRYPOINT ["python", "your_script.py"]
 
 # Download and install Liquibase
 RUN wget https://github.com/liquibase/liquibase/releases/download/v${LIQUIBASE_VERSION}/liquibase-${LIQUIBASE_VERSION}.zip \
```
milo-hyben, line 44: Is this duplicating wget at LN 19?

db/deploy/Dockerfile
```diff
@@ -32,4 +62,4 @@ COPY . ./
 RUN python3 -m pip install --no-cache-dir --break-system-packages -r requirements.txt
 
 # Run the FastAPI app on container startup
-CMD ["uvicorn", "main:app", "--host", "0.0.0.0", "--port", "8080"]
+CMD ["uvicorn", "main:app", "--host", "0.0.0.0", "--port", "8000"]
```
milo-hyben, line 65: I wonder what is the reason behind 8000, 8080 is more of a web standard.

deploy/api/Dockerfile
```diff
@@ -7,15 +7,26 @@ ENV SM_ENVIRONMENT ${SM_ENVIRONMENT}
...
+# Change ownership of the application directory to the non-root user
+RUN chown -R appuser:appuser /app/sample_metadata/
+
+# Switch to the non-root user
+USER appuser
+
+EXPOSE $PORT
+
+# Command to run the FastAPI app
+CMD ["uvicorn", "api.main:app", "--host", "0.0.0.0", "--port", "8000"]
```
milo-hyben, line 32: should not "8000" be "${PORT}" ?

milo-hyben (commented):
Excellent work Sabrina!
I only have few questions as per my comments.

db/deploy/Dockerfile
violetbrina: Ah of course it is. I'll remove the dup. Thanks for that pick up.

db/deploy/Dockerfile
violetbrina: Fixed

deploy/api/Dockerfile
violetbrina: Done!

milo-hyben (approved):

PR #958, opened 2024-09-25
Fixed faulty FCF dictionary behaviour

nevoodoo:
There was an issue reported by @EddieLF where the dictionary passed to the `/update` endpoint was being discarded. I traced the bug to the GCS Client not listing blobs as expected. The behaviour seems to have changed recently and I have tried to fix this by using `blob_name` which does not have a `/` at the front. There are some additional changes being made:

- Unique constraint added to ensure we do not add the same file/output over and over again to the join table. There is a caveat here - MariaDB does not treat NULL values as the same, and so rows with NULL values in some columns look unique despite being the same. I have added a DISTINCT clause to my query but there could be a lot of duplicate inserts to this table. This is worth looking at, in more detail. Open for suggestions <3 
- Query should return all outputs (both files and plain strings), which it now does.

### DIFF
--- a/db/python/tables/output_file.py
+++ b/db/python/tables/output_file.py
@@ -121,7 +121,7 @@ class OutputFileTable(DbBase):
         """Add file to an analysis (through the join table)"""
         _query = dedent(
             """
-            INSERT INTO analysis_outputs
+            INSERT IGNORE INTO analysis_outputs
                 (analysis_id, file_id, json_structure, output)
             VALUES (:analysis_id, :file_id, :json_structure, :output)
         """
@@ -147,6 +147,7 @@ class OutputFileTable(DbBase):
         """
         files = await self.find_files_from_dict(json_dict=json_dict)  # type: ignore [arg-type]
         file_ids: list[int] = []
+        outputs: list[str] = []
 
         async with self.connection.transaction():
             if 'main_files' in files:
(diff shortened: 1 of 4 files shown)


db/python/tables/output_file.py
```diff
@@ -121,7 +121,7 @@ async def add_output_file_to_analysis(
         """Add file to an analysis (through the join table)"""
         _query = dedent(
             """
-            INSERT INTO analysis_outputs
+            INSERT IGNORE INTO analysis_outputs
```
milo-hyben, line 124: It feels like a code smell :-)
wonder if we should report the error IGNORE is trying to hide or calling code should be revisited.
Is this purely to prevent duplicates?

milo-hyben (commented):
Just that one comment, otherwise looks good to me.

db/python/tables/output_file.py
nevoodoo: Yep, purely to prevent duplicates, I could replace this with an `ON DUPLICATE` clause, but I don't quite understand/know if one is better or preferred over the other. Thoughts?

milo-hyben: There is a discussion regarding this:
https://stackoverflow.com/questions/548541/insert-ignore-vs-insert-on-duplicate-key-update

Someone mentioned ON DUPLICATE can cause mysql deadlocking.
I do not have preference/strong opinion on this as both are non standard SQL.
Let's stick with IGNORE as you had implemented. We can alway revisit if needed.

milo-hyben (approved):

PR #960, opened 2024-09-29
Setup Renovate

nevoodoo:
This adds the RenovateBot that uses `pip-compile` for now. Also removing Dependabot to avoid conflicts.  Also bumping `strawberry-graphql` because of a recent security patch.


milo-hyben (approved):
Looking good to me! Will be interesting to see how good is renovatebot.

PR #966, opened 2024-09-30
GraphQL Mutations

nevoodoo:
A continuation of #686 - created this fresh branch because there were too many merge conflicts to deal with. Feedback and changes requested in #686 should be incorporated here.


api/graphql/mutations/analysis_runner.py
```diff
@@ -0,0 +1,76 @@
...
+@strawberry.type
+class AnalysisRunnerMutations:
+    """Analysis Runner mutations"""
+
+    project_id: strawberry.Private[int]
+
+    @strawberry.mutation
+    async def create_analysis_runner_log(  # pylint: disable=too-many-arguments
+        self,
+        ar_guid: str,
```
dancoates, line 26: Rather than having so many args here, better to wrap them in an input type

api/graphql/types/__init__.py
```diff
@@ -0,0 +1,207 @@
+import strawberry
```
dancoates, line 1: I think for the most part these types would be better placed being alongside the mutations they are used in

api/graphql/mutations/sequencing_group.py
```diff
@@ -36,3 +59,30 @@ async def add_comment(
             content=content,
         )
         return GraphQLComment.from_internal(result)
+
+    @strawberry.mutation
+    async def update_sequencing_group(
+        self,
+        sequencing_group_id: str,
+        sequencing_group: strawberry.scalars.JSON,
```
dancoates, line 67: These inputs should follow the same pattern used elsewhere of having a input type with the id as part of it

api/graphql/mutations/sample.py
```diff
@@ -36,3 +63,70 @@ async def add_comment(
...
+        return [SampleUpsertType.from_upsert_internal(s) for s in upserted]
+
+    # endregion CREATES
+
+    # region OTHER
+
+    @strawberry.mutation
+    async def update_sample(
+        self,
+        id_: str,
```
dancoates, line 118: id is already included in `SampleUpsertInput` it shouldn't be a separate param as well

dancoates (changes requested):
I've left some specific comments but there are a few general things that I think will need updating in this too.
- I know it was my idea but I think wrapping the mutations in a  project to implement the permissions was actually not a great idea. It means that we can only restrict to full write permissions but some mutations like cohorts and comments actually allow for a lower level of permissions. After thinking about it some more, I think the best approach is to remove all ther permissions checks from graphql and only have them in layers. 
- The return values of mutations need to be consistent and should return the thing that was updated/created. This makes the mutations a ton easier to use in the frontend as you don't need to make an extra request to get the updated item. It's also important that they return the exact same type as if you were to request the item, not a type like `AssayUpsertType`. This is due to the way Apollo handles its normalized cache. For things like deletes we can probably just return True or None, just need to settle on a standard and stick with it.
- Also need to be consistent with inputs, it seems like in most cases the mutations are accepting a single input type, or in some cases the id and an input type. I think that's fine but we should be consistent and make sure that we don't have mutations that have a ton of inputs.

api/graphql/mutations/analysis.py
```diff
@@ -0,0 +1,98 @@
...
+
+        analysis_id = await atable.create_analysis(
+            AnalysisInternal.from_db(**strawberry.asdict(analysis)),
+            project=root.project_id,
+        )
+
+        return analysis_id
+
+    @strawberry.mutation
+    async def update_analysis(
```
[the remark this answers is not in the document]
nevoodoo: Hmm, not sure what the best approach is here then. Should we include `project` as an additional arg to the layer call? As it stands, it doesn't need a `project_id` which as you pointed out, defeats the purpose of putting it under the wrapper :/

api/graphql/mutations/analysis.py
```diff
@@ -0,0 +1,98 @@
...
+
+        if not analysis.sequencing_group_ids and not analysis.cohort_ids:
+            raise ValueError('Must specify "sequencing_group_ids" or "cohort_ids"')
+
+        analysis_id = await atable.create_analysis(
+            AnalysisInternal.from_db(**strawberry.asdict(analysis)),
+            project=root.project_id,
+        )
+
+        return analysis_id
```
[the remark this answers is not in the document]
nevoodoo: Would you reckon we just add the `get` layer call within the mutation or is it better to refactor all the way to the table calls to return this? The latter option would mean refactoring the `routes` too then.

api/graphql/mutations/analysis.py
```diff
@@ -0,0 +1,98 @@
...
+        atable = AnalysisLayer(connection)
+        await atable.update_analysis(
+            analysis_id,
+            status=analysis.status,
+            output=analysis.output,
+            outputs=analysis.outputs,  # type: ignore [arg-type]
+            meta=analysis.meta,  # type: ignore [arg-type]
+            active=analysis.active,
+        )
+        return True
```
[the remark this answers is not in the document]
nevoodoo: Same comment as above ^

api/graphql/mutations/analysis.py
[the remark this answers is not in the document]
dancoates: I think adding the get call here is fine

api/graphql/mutations/analysis.py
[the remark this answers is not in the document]
dancoates: Ah sorry you can probably ignore this comment, I made it before realising that the approach of wrapping in project is probably wrong.

api/graphql/mutations/sequencing_group.py
nevoodoo: Should this also return the sequencing group instead of a `bool`?

dancoates: yep!

api/graphql/mutations/cohort.py
```diff
@@ -0,0 +1,171 @@
...
+        cohort_criteria: CohortCriteriaInput | None = None,
+        dry_run: bool = False,
+    ) -> Annotated['GraphQLCohort', strawberry.lazy('api.graphql.schema')]:
+        """
+        Create a cohort with the given name and sample/sequencing group IDs.
+        """
+        from api.graphql.schema import GraphQLCohort
+
+        connection: Connection = info.context['connection']
+        connection.check_access(
```
dancoates, line 71: This check relies on the project being set on the connection, I think it will error out here

api/graphql/mutations/cohort.py
```diff
@@ -0,0 +1,171 @@
...
+        template: CohortTemplateInput,
+        info: Info,
+    ) -> Annotated['GraphQLCohortTemplate', strawberry.lazy('api.graphql.schema')]:
+        """
+        Create a cohort template with the given name and sample/sequencing group IDs.
+        """
+        from api.graphql.schema import GraphQLCohortTemplate
+
+        connection: Connection = info.context['connection']
+        connection.check_access(
```
dancoates, line 133: Same here, this won't work as the project won't be set on the connection

api/graphql/mutations/cohort.py
```diff
@@ -0,0 +1,171 @@
...
+            {ProjectMemberRole.writer, ProjectMemberRole.contributor}
+        )
+        clayer = CohortLayer(connection)
+
+        if not connection.project:
```
dancoates, line 138: This will always error out as the connection won't have a project, the same applies to all uses of `connection.project` and `connection.project_id` in this file.

api/graphql/mutations/project.py
```diff
@@ -34,3 +52,108 @@ async def add_comment(
...
+    @strawberry.mutation
+    async def update_project(
+        self,
+        project: str,
+        project_update_model: JSON,
+        info: Info,
+    ) -> JSON:
+        """Update a project by project name"""
+        connection = info.context['connection']
+        connection.check_access({ProjectMemberRole.project_admin})
```
dancoates, line 96: Same issue here, can't use `check_access` as it requires a project on the connection.

api/graphql/mutations/participant.py
```diff
@@ -34,3 +66,76 @@ async def add_comment(
...
+        info: Info,
+    ) -> list[Annotated['GraphQLParticipant', strawberry.lazy('api.graphql.schema')]]:
+        """
+        Upserts a list of participants with samples and sequences
+        Returns the list of internal sample IDs
+        """
+        from api.graphql.schema import GraphQLParticipant
+
+        connection: Connection = info.context['connection']
+        connection.check_access(FullWriteAccessRoles)
```
dancoates, line 105: Can't use `check_access` as it requires a project on the connection

api/graphql/mutations/project.py
```diff
@@ -34,3 +52,108 @@ async def add_comment(
...
+        info: Info,
+    ) -> JSON:
+        """
+        Delete all data in a project by project name.
+        """
+        connection = info.context['connection']
+        connection.check_access({ProjectMemberRole.project_admin})
+
+        ptable = ProjectPermissionsTable(connection)
+        assert connection.project
```
dancoates, line 120: There won't be a project on the connection

api/graphql/mutations/sequencing_group.py
```diff
@@ -36,3 +60,45 @@ async def add_comment(
...
+
+        # Should be moved to the sequencing group layer
+        connection.check_access_to_projects_for_ids(
+            [root.project_id], FullWriteAccessRoles
+        )
+
+        if not sequencing_group.id:
+            raise ValueError('Sequencing group ID must be provided for update')
+
+        # TODO: Review this against the route endpoint
```
dancoates, line 84: Does this TODO still need doing?

dancoates (changes requested):
Sorry for the (extremely) slow re-review. I've gone through and added some more comments. A couple of little things and two main things:

- Need to remove the high level permission checks on mutations entry points that restrict to FullWriteAccess as some mutations are possible with just contributor access. I think in doing this it might be worth removing the `project` wrapper around mutations too to simplify things, as it wouldn't be doing much other than passing down the project id at that point, and not all mutations actually need a project id.
- Need to update anything that is currently relying on having a project on the connection object as it won't exist unfortunately.

api/graphql/mutations/family.py
```diff
@@ -4,13 +4,25 @@
...
-    from api.graphql.schema import GraphQLComment
+    from api.graphql.schema import GraphQLComment, GraphQLFamily
+
+
+@strawberry.input
+class FamilyUpdateInput:
+    """Family update type"""
+
+    id: int
+    external_id: JSON | None = None
```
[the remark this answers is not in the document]
nevoodoo: Fixed!

api/graphql/mutations/__init__.py
```diff
@@ -1,56 +1,117 @@
...
     """Mutation class"""
 
+    # Project
+    @strawberry.field
+    def project(
+        self, name: str, info: Info[GraphQLContext, 'Mutation']
+    ) -> ProjectMutations:
+        """Project mutations"""
+        connection = info.context['connection']
+        projects = connection.get_and_check_access_to_projects_for_names(
```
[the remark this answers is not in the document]
nevoodoo: Yep, you're right, removed!

api/graphql/mutations/project.py
nevoodoo: Fixed!

api/graphql/mutations/participant.py
nevoodoo: Fixed!

api/graphql/mutations/project.py
nevoodoo: Fixed!

api/graphql/mutations/sequencing_group.py
nevoodoo: Just reviewed it, I probably should have left more notes in the TODO to understand why I was supposed to review it ;)

nevoodoo:
@dancoates this should be good for a final review now :) I've bumped test coverage on all the mutations, and I'm happy with this unless there's more things that you think I can polish/fix.

dancoates (approved):
This is looking good! Permissions is still a mess generally, but this works as well as possible within the existing system.

PR #979, opened 2024-10-28
Add copying of application code back into deploy dockerfile

dancoates:
This was accidentally removed in PR #957, causing the container to not start correctly due to missing application code

### DIFF
--- a/deploy/api/Dockerfile
+++ b/deploy/api/Dockerfile
@@ -19,6 +19,9 @@ RUN pip install --no-cache-dir --no-deps -r requirements.txt
 
 # Copy the application code
 COPY api api
+COPY db db/
+COPY models models/
+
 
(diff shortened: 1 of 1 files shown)


violetbrina (approved):

jmarshall:
Is it that this was inadvertently / incorrectly removed in PR #957?

(It'd be great to have that context — and what went wrong, if anything — in the commit message or as a comment here.)

Also they were previously in alphabetical order, so it'd help diffing a tiny bit if this put them back identically.

dancoates:
Thanks @jmarshall good points, I've added a reference to the PR in the commit, and fixed the ordering to match what it was previously for better diff/blame.

jmarshall (approved):
Perfect, thanks!

PR #984, opened 2024-11-04
register missing combiner analysis script

Alexander-Stuckey:
Does this even do what I want it to do? Kinda just winged it based off another script.

### DIFF
--- a/scripts/register_migging_combiner_analysis.py
+++ b/scripts/register_migging_combiner_analysis.py
@@ -1,3 +1,10 @@
+"""Register a missing combiner analysis in Metamist
+
+Ad-hoc script to register a missing combiner analysis in Metamist.
+Is supremely inflexible since there is a short timeframe for combining previous
+VDS' for the OurDNA browser release.
+"""
+
 from argparse import ArgumentParser
 
 from hail.vds import read_vds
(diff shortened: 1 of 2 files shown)


scripts/register_migging_combiner_analysis.py
```diff
@@ -0,0 +1,38 @@
...
+    return read_vds(to_path(vds)).variant_data.s.collect()
+
+
+def main(vds: str, dataset: str):
+    aapi = AnalysisApi()
+    am = Analysis(
+        type='combiner',
+        output=vds,
+        status=AnalysisStatus('completed'),
+        sequencing_group_ids=get_sg_ids(vds),
```
MattWellie, line 41: you'll want to import `cpg_utils.hail_batch.init_batch` and call that, you won't be able to call a Hail method without activating a hail runtime, and that helper method will deal with all the billing project/ref genome stuff

Alexander-Stuckey: I always hecking forget that

Alexander-Stuckey: Alright, that's fixed now

MattWellie (approved):
Looks good, happy to increase the flexibility later

PR #985, opened 2024-11-04
Minor fixes

Alexander-Stuckey:
Minor fixes to the combiner script, use better config settings to get the dataset name.

### DIFF
--- a/scripts/register_missing_combiner_analysis.py
+++ b/scripts/register_missing_combiner_analysis.py
@@ -9,7 +9,7 @@ from argparse import ArgumentParser
 
 from hail.vds import read_vds
 
-from cpg_utils.hail_batch import dataset_path, init_batch
+from cpg_utils.hail_batch import config_retrieve, dataset_path, init_batch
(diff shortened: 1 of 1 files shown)


scripts/register_missing_combiner_analysis.py
```diff
@@ -24,10 +23,10 @@ def get_sg_ids(vds: str) -> list[str]:
     Returns:
         list[str]: A list of sequencing group IDs.
     """
-    return read_vds(to_path(vds)).variant_data.s.collect()
+    return read_vds(dataset_path(vds)).variant_data.s.collect()
```
MattWellie, line 26: I'm personally more a fan of passing the full path to the analysis we want to register on the CLI, instead of this helper method taking half a path as an argument.

MattWellie (approved):

scripts/register_missing_combiner_analysis.py
Alexander-Stuckey: Fair enough, I can change that back.

PR #986, opened 2024-11-04
Update project_insights code to account for multiple family external ids changes

jmarshall:
When on the final phase of merging #896, I did not exhaustively recheck the code base for newly arrived references to the soon-to-be-obsolete `family.external_id` field. Since that PR had been begun, this code had been merged and uses this field. And unluckily the project_insights test cases tested only the summary not the details, so the old field access was not detected via the unit tests.

Fix this instance to use the new separate `family_external_id` table, and add a details test case that just causes the relevant code to be run.

Note that as for `pext.external_id` and `sext.external_id`, this might benefit from being aggregated (or alternatively maybe joining also on `fext.name = ''`) as at present multiple rows may be returned if an an item actually does have multiple external ids. But that can be adjusted as and when it actually occurs.

### DIFF
--- a/db/python/layers/project_insights.py
+++ b/db/python/layers/project_insights.py
@@ -750,7 +750,7 @@ SELECT
     sg.technology as sequencing_technology,
     s.type as sample_type,
     f.id as family_id,
-    f.external_id as family_external_id,
+    fext.external_id as family_external_id,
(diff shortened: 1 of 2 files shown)


ckrishk (approved):
LGTM

PR #988, opened 2024-11-13
Fix return type of enum create routes

dancoates:
The return type was list[str] but the table method actually just returned a string. This caused a 500 error on the response


jmarshall (approved):
LGTM.

PR #991, opened 2024-11-15
Script to upsert MGRB participants

michael-harper:
As per [this Slack thread](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1728524002090459).

MGRB data was ingested without participants. The rest of the data such as sample's sequencing_groups etc. was still ingested however they were never assigned to a participant. Not having participants breaks production-pipelines.

This PR creates new participants and adds the current sample data to them.

@milo-hyben and I have pair programmed and run the results on mgrb-test in metamist that has now fixed the data in metamist.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,7 +6,7 @@
 #
 aiohappyeyeballs==2.4.0
     # via aiohttp
-aiohttp==3.10.5
+aiohttp==3.10.*
     # via -r requirements.in
 aiomysql==0.2.0
     # via databases
@@ -479,5 +479,5 @@ wrapt==1.16.0
     # via
     #   deprecated
     #   testcontainers
-yarl==1.9.10
+yarl==1.12.*
     # via aiohttp
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -110,7 +110,7 @@ docker==7.1.0
     # via testcontainers
 email-validator==2.2.0
     # via fastapi
-fastapi[all]==0.112.2
+fastapi[all]==0.115.5
     # via
     #   -r requirements.in
     #   strawberry-graphql
@@ -413,7 +413,7 @@ sqlalchemy==2.0.34
     # via
     #   databases
     #   testcontainers
-starlette==0.40.0
+starlette==0.41.2
     # via
     #   fastapi
     #   strawberry-graphql
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -413,7 +413,7 @@ sqlalchemy==2.0.34
     # via
     #   databases
     #   testcontainers
-starlette==0.38.4
+starlette==0.40.0
     # via
     #   fastapi
     #   strawberry-graphql
@@ -471,7 +471,7 @@ watchfiles==0.24.0
     # via uvicorn
 websockets==13.0.1
     # via uvicorn
-werkzeug==3.0.4
+werkzeug==3.0.6
     # via
     #   flask
     #   functions-framework
--- a/scripts/upsert_mgrb_participants.py
+++ b/scripts/upsert_mgrb_participants.py
@@ -163,9 +163,7 @@ def main(project: str, participant_meta: str, sample_meta: str):
     participant_upserts = []
     for peid, sample_upserts in participant_sample_map.items():
         participant_upserts.append(
-            create_participant(
-                sample_upserts, peid, participant_sex_map[peid]
-            )
+            create_participant(sample_upserts, peid, participant_sex_map[peid])
         )
 
     # Upsert participants
--- a/scripts/upsert_mgrb_participants.py
+++ b/scripts/upsert_mgrb_participants.py
@@ -105,14 +105,16 @@ def read_files_from_gcs(bucket_name: str, file_path: str):
 
 
 @click.command()
+@click.option('--test', is_flag=True, help='Run in test mode.')
 @click.option('--project', help='Project name to upsert participants to.')
 @click.option('--participant-meta', help='Path to participant metadata CSV.')
 @click.option('--sample-meta', help='Path to sample metadata CSV.')
-def main(project: str, participant_meta: str, sample_meta: str):
+def main(test: bool, project: str, participant_meta: str, sample_meta: str):
     """Upsert participants to metamist"""
     # Query metamist
     data_response = get_sample_data(project)
     project_id = data_response.get('project')['id']
+    bucket_name = f'cpg-{project}-{'test' if test else 'main'}-upload'
 
     sample_eid_mapping = map_sample_eid_to_sample_data(data_response)
 
@@ -120,8 +122,8 @@ def main(project: str, participant_meta: str, sample_meta: str):
     logging.basicConfig(level=logging.INFO)
 
     # Read in csv
-    participant_data_io = read_files_from_gcs(BUCKET_NAME, participant_meta)
-    sample_data_io = read_files_from_gcs(BUCKET_NAME, sample_meta)
+    participant_data_io = read_files_from_gcs(bucket_name, participant_meta)
+    sample_data_io = read_files_from_gcs(bucket_name, sample_meta)
 
     participant_data_list = list(csv.DictReader(participant_data_io))
     sample_data_list = list(csv.DictReader(sample_data_io, delimiter='\t'))
--- a/scripts/upsert_mgrb_participants.py
+++ b/scripts/upsert_mgrb_participants.py
@@ -1,0 +1,179 @@
+import csv
+import logging
+from collections import defaultdict
+from io import StringIO
+from typing import Dict, List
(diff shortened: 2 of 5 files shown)


scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+
+BUCKET_NAME = 'cpg-mgrb-test-upload'
+
+SAMPLE_QUERY = gql(
+    """
+    query getSampleData($project: String!) {
+        project(name: $project) {
+            id
+        }
+        sample(project: {eq: $project}) {
```
vivbak, line 25: I probably would've nested sample under project here, in the spirit of GraphQL and since the two fields are inherently coupled. But I assume the decision here was to simplify parsing later?

scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+from google.cloud import storage
+
+from metamist.apis import ParticipantApi
+from metamist.graphql import gql, query
+from metamist.models import (
+    ParticipantUpsert,
+    SampleUpsert,
+)
+
+BUCKET_NAME = 'cpg-mgrb-test-upload'
```
vivbak, line 17: I wouldn't have hard-coded this. We're using a project name in a public repository, which is generally bad practice. We could've generalised this script to 'upsert_participants' and then passed this in at the CLI

vivbak: Also is the test-upload bucket the appropriate place at this point? Is this script for running in test or in main or both?

scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+    }
+    """
+)
+
+papi = ParticipantApi()
+
+
+def map_sample_eid_to_sample_data(data):
```
vivbak, line 63: Nit: Type hint

scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+    """Map sample external ID to sample data"""
+    return {sample['externalId']: sample for sample in data['sample']}
+
+
+def get_sample_data(project):
```
vivbak, line 68: Nit: Type hint

scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+        raise ValueError('Expected a dictionary response from query')
+    return response
+
+
+def create_participant(sample_data: list, peid: str, sex: int):
+    """Create participant upsert object"""
+    participant_upsert = [
+        ParticipantUpsert(
+            id=None,
+            external_ids={'': peid},
```
vivbak, line 81: Why do we pass an empty string here? I think it would be beneficial to call it something.

scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+        ),
+    ]
+    return participant_upsert
+
+
+def create_sample(project_id: int, seid: str, siid: str):
+    """Create sample upsert object"""
+    return SampleUpsert(
+        id=siid,
+        external_ids={'': seid},
```
vivbak, line 93: Same here, it feels against the spirit of the data structure to leave a key blank?

scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+
+    # Map peid to a list of sample upsert objects
+    participant_sex_map = {}
+    participant_sample_map: Dict[str, List[SampleUpsert]] = defaultdict(list)
+
+    for name_row in sample_data_list:
+        # Get the external ID
+        seid = '.'.join(name_row['CRAM'].split('.')[:2])
+
+        meta_row = metadata_map.get(name_row['RGSM (SampleID)'])
```
vivbak, line 145: If we want to generalise this script, we could pull this var out too.

scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+        # Get participant external ID
+        peid = meta_row['externalID']
+
+        # Code sex
+        sex = 2 if meta_row['isFemale'].upper() == 'TRUE' else 1
+
+        # Get the sample internal ID
+        if seid not in sample_eid_mapping:
+            print(f'Sample {seid} not found in metamist.')
+            continue
```
vivbak, line 155: Should this do more than continue? From a data management perspective, do we mind if we have samples in the bucket that are not in metamist?

scripts/upsert_mgrb_participants.py
```diff
@@ -0,0 +1,177 @@
...
+
+    participant_upserts = []
+    for peid, sample_upserts in participant_sample_map.items():
+        participant_upserts.append(
+            create_participant(sample_upserts, peid, participant_sex_map[peid])
+        )
+
+    # Upsert participants
+    for participant_upsert in participant_upserts:
+        api_response = papi.upsert_participants(project, participant_upsert)
```
vivbak, line 171: Doesn't this endpoint take in a list of participant upserts and handle in bulk?

requirements-dev.txt
```diff
@@ -110,7 +110,7 @@ docker==7.1.0
     # via testcontainers
 email-validator==2.2.0
     # via fastapi
-fastapi[all]==0.112.2
+fastapi[all]==0.115.5
```
vivbak, line 113: Wondering the rationale for the package updates here?

vivbak (changes requested):
Great Job! Not all of my comments need to be actioned and a few are nits that probably don't matter for a single use script.

requirements-dev.txt
vivbak: @dancoates FYI

scripts/upsert_mgrb_participants.py
michael-harper: My experience with this fix has been that other scripts have tried to be generalised but there was nothing that quite fit this specific issue. I don't want to say this specific case won't happen again but it seems like a fairly niche scenario that we're trying to fix here especially given the makeup of metadata files from collaborators can be vastly different

scripts/upsert_mgrb_participants.py
michael-harper: This is checking if there's samples in the metadata that aren't in metamist, which to my knowledge can happen fairly regularly given sample failures etc.

scripts/upsert_mgrb_participants.py
michael-harper: @milo-hyben ?

scripts/upsert_mgrb_participants.py
milo-hyben: We should make the bucket name as input parameter to the script.

scripts/upsert_mgrb_participants.py
milo-hyben: My understanding was, it has to be empty to becomes the main organisation external_id:
https://github.com/populationgenomics/metamist/blob/ab02bd2190107ca39db73b61b415a86960452ff7/models/base.py#L13

scripts/upsert_mgrb_participants.py
milo-hyben: I guess you are right we need to define what is the PRIMARY_EXTERNAL_ORG in this set of samples/participants.

milo-hyben: There might be an issue with this old samples, they would have PRIMARY_EXTERNAL_ORG set to empty string.

scripts/upsert_mgrb_participants.py
milo-hyben: we might make 'RGSM (SampleID)' as input parameter to the script.

scripts/upsert_mgrb_participants.py
milo-hyben: Very good point Vivian!
Lets, fix the function create_participant, so it returns only object and then we need only one call to upsert_participants with the list.

scripts/upsert_mgrb_participants.py
milo-hyben: I guess this script meant to update participants for existing samples, if there is some new sample in the file it gets ignored.

requirements-dev.txt
milo-hyben: As you can see there is a new security issue (failed check) since Friday, when it was OK: 'Found 1 known vulnerability in 1 package'
Personally I would be reluctant to merge with those red crosses :-)

vivbak (approved):
Given the time pressure I'm approving here :) But I agree with @milo-hyben in that we shouldn't be merging red crosses? Something for software to look into?

milo-hyben:
Let me fix that security check issue. Will commit a fix shortly.

PR #995, opened 2024-11-21
Upsert participants fix to allow empty external-org-name

milo-hyben:
Small fix to allow empty string as external our name for backward compatibility


michael-harper (approved):
Went through change with Milo on Zoom. Looks good and works in local test environment

PR #1000, opened 2024-12-02
Analysis files mover script

EddieLF:
This script is intended to be used for sequencing groups which have been moved across projects, or from one dataset to another. 

Assuming the records have been update in the MariaDB backend, i.e. the `analysis` and `sequencing_group` records have had their `project` field updated, this script completes the move by:

- Finding analyses of the types `cram`, `gvcf`, `web`, `sv` and moving their outputs to the new bucket
- Updating the above analysis records with the new paths
- Finding and moving a number of extra outputs which are not currently recorded in the analysis table.

It will also

- Unarchive `COLDLINE` or `ARCHIVE` blobs before copying, if the `unarchive` flag is set
- Log everything that is changed, copied, or updated

---

Test run: https://batch.hail.populationgenomics.org.au/batches/534393/jobs/1

This is only a dry run, since I don't have a repro case of an SG who's records/analyses have been moved to a new project but who's files are still in the old bucket.

### DIFF
--- a/scripts/move_analyses_across_projects.py
+++ b/scripts/move_analyses_across_projects.py
@@ -28,7 +28,7 @@ from metamist.parser.generic_metadata_parser import run_as_sync
 # Define the query to get the analysis records that need to be moved
 ANALYSES_QUERY = gql(
     """
-    query sgAnalyses($sequencingGroupIds: [String]!, $dataset: String!, $analysisTypes: [String!]) {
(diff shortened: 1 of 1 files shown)


nevoodoo (approved):
Looks good @EddieLF , I haven't looked too deep but from a general skim of the code, it looks fine! And you have `dry_run` anyway so I think it should be sweet :)

PR #1003, opened 2024-12-12
SET-250 meta table exports for sample and participant

dancoates:
This is the start of an experimental feature for metamist that allows querying a project's metadata tables on the frontend using duckdb-wasm. This would allow for complex SQL queries that are totally safe to run as they happen on the client, so there's no risk of SQL injection. 
This also enables the queries and aggregations for metamist dashboards to be written in SQL which is a much faster and more expressive way of wrangling the data for each visualisation.
It also allows for potential future features where users can write, save, and share queries and visualisations via the metamist UI without having to deploy metamist.

This PR implements metadata table export for sample and participant metadata tables. The key feature of these tables is that the meta json blob is flattened out to columns, so it can be easily queried and aggregated. Additionally, external ids are also flattened out to columns so they are more easily queried without having to join with another table.

The tables are exported as parquet files, parquet is a file format for efficiently storing tabular data. It can be quickly queried by duckdb and transformation to and from arrow is quite quick. Arrow is the in memory table format that tools like duckdb and polars use.


violetbrina:
I've had a quick first glance. Just wanting to confirm something (haven't done a fully deep dive yet). So far looks good, I just wanted to check that we are only returning data that the user has access too correct? 
I always forget what layer permission checks occur in, but just making sure they're in there.

dancoates:
> I've had a quick first glance. Just wanting to confirm something (haven't done a fully deep dive yet). So far looks good, I just wanted to check that we are only returning data that the user has access too correct? I always forget what layer permission checks occur in, but just making sure they're in there.

Yeah good question, it is this line on the routes that applies the permission check.
https://github.com/populationgenomics/metamist/pull/1003/files#diff-b663c7a535ec3597f3853c9811ce431167428756eae96a93a6254b738e740d87R265

It isn't exactly obvious that that is what is happening, I'm hoping to rejig the permissions checks some time soonish to make this stuff clearer.

db/python/tables/meta_table.py
```diff
@@ -0,0 +1,144 @@
...
+                    meta_rows_str,
+                    map_inference_threshold=-1,
+                    format='newline_delimited',
+                ).arrow(),
+            )
+            # Prefix all meta columns with `meta_` to avoid clashes with the main table
+            meta_columns = """
+                ,COLUMNS(mt.*) as 'meta_\\0'
+            """
+            meta_join = 'POSITIONAL JOIN meta_table mt'
```
violetbrina, line 106: Can't say I've ever seen a `POSITIONAL JOIN` before. Interesting, how does it work. More just curious than anything.

db/python/tables/meta_table.py
```diff
@@ -0,0 +1,144 @@
...
+        Return a flat tabular parquet file for the provied query. Optionally include
+        metadata and external_ids columns. This returns a BytesIO object with the
+        parquet file.
+        """
+
+        rows = await self.connection.fetch_all(
+            query,
+            {
+                'project': project,
+                'primary_external_org': PRIMARY_EXTERNAL_ORG,
```
violetbrina, line 59: So am I right in understanding that the `EXTERNAL_ORG_SENTINEL` is a placeholder just to make duckdb happy?

db/python/tables/meta_table.py
```diff
@@ -0,0 +1,144 @@
...
+                    external_id_rows_str,
+                    format='newline_delimited',
+                ).arrow(),
+            )
+
+            # call the primary external_id column `external_id` and prefix all other
+            # columns with `external_id_` to avoid clashes with the main table
+            external_id_columns = f"""
+                ,et.{EXTERNAL_ORG_SENTINEL} as external_id
+                ,COLUMNS(et.* EXCLUDE {EXTERNAL_ORG_SENTINEL}) as 'external_id_\\0'
```
violetbrina, line 124: And this is where it get's replaced with either `external_id` or `external_id_<column_name>`?
Not familiar with duckdb, just trying to grasp what is going on here.

db/python/tables/meta_table.py
```diff
@@ -0,0 +1,144 @@
...
+
+            FROM main_table m
+            {external_id_join}
+            {meta_join}
+        """)
+
+        buffer = BytesIO()
+        pq.write_table(combined.arrow(), buffer)
+
+        duck.close()
```
violetbrina, line 143: Just a query. The use of duckdb is a personal preference? Or does it have particular functionality to enable the creation of the parquet file? Or is the pyarrow.parquet module here the one actually doing the write to the parquet file and would take a table in a sensible format?

db/python/tables/participant.py
```diff
@@ -285,6 +285,37 @@ async def get_participants(
...
+            query=query,
+            row_getter=lambda row: {
+                'participant_id': row['id'],
+                'reported_sex': row['reported_sex'],
+                'reported_gender': row['reported_gender'],
+                'karyotype': row['karyotype'],
+            },
+            has_external_ids=True,
+            has_meta=True,
+        )
```
violetbrina, line 317: Love this. So succinct, so clean 👌

test/test_meta_table.py
```diff
@@ -0,0 +1,134 @@
...
+        result = query_parquet(
+            {'samples': samples},
+            'SELECT * FROM samples order by sample_id',
+        )
+
+        self.assertEqual(2, len(result))
+        self.assertEqual('field_1_value', result[0]['meta_field_1'])
+        self.assertEqual('field_2_value', result[1]['meta_field_2'])
+        self.assertEqual('Test01', result[0]['external_id'])
+        self.assertEqual('Test02', result[1]['external_id'])
```
violetbrina, line 134: Great tests. Super clear. Love it!

violetbrina (commented):
Nice PR. Love the new parquet file export functionality. Just some questions about that new MetaTable class but other than that looks good.

db/python/tables/meta_table.py
dancoates: It is basically like `zip` in python, or a horizontal concat in dataframes. It take two tables and joins them based on the row number. Here's the docs for it https://duckdb.org/docs/sql/query_syntax/from.html#positional-joins

db/python/tables/meta_table.py
dancoates: Yep, that's right. The default external id name is an empty string, and when duckdb reads the json rows, if we left it as an external string the column name would be an empty string, and sql columns can't be empty strings. Duckdb does kinda handle it by giving the column a name like `col1` (I can't remember exactly what it was) but the name is based on column position which isn't consistent.

db/python/tables/meta_table.py
dancoates: Yeah, duckdb allows renaming columns using the `COLUMNS` operator
https://duckdb.org/docs/sql/expressions/star.html#columns-regular-expression

These two lines give the default external id column the name `external_id` and then prefix other external id column names with `external_id_`.

db/python/tables/meta_table.py
dancoates: The pyarrow module is doing the actual writing of the parquet file. Duckdb can write parquet files but only to disk and not into a filelike BytesIO object. Because duckdb and pyarrow both operate on arrow tables the data isn't copied in memory and it is really fast to interoperate between the two tools. Arrow is a really cool specification and has implementations in a bunch of languages. It's becoming a bit of a standard for working with in-memory tabular data.

In theory this could all be done just with pyarrow, but I'm mostly using duckdb here for its json parsing abilities. The JSON we have in metamist isn't very well structured, and lots of tools like Polars will try and infer the schema of the first chunk of rows. Because our meta blobs are a bit all over the place it's very possible that row 10,000 will have some columns that no other rows have. Duckdb is really fast at parsing nested and complex json.

db/python/tables/meta_table.py
violetbrina: Fab, that's what I thought. The code is clear enough I just wanted to confirm my understanding.

db/python/tables/meta_table.py
violetbrina: Good explanation. Yeah I was just wondering the decision behind these choices more than anything. I've seen some use of pyarrow before. Duckdb is new but very cool.

violetbrina (approved):
Looks good to me!

api/utils/parquet_route.py
```diff
@@ -0,0 +1,62 @@
...
+            table_bytes = await func(*args, **kwargs)
+
+            if table_bytes is None:
+                return JSONResponse(
+                    status_code=404,
+                    content=TableError(error='Table empty or not found'),
+                )
+
+            headers = {
+                'Content-Disposition': 'attachment; filename="participant_table.parquet"'
```
jmarshall, line 48: Code here is not specific to the **participant** table, I'm thinking?

db/python/tables/participant.py
```diff
@@ -285,6 +285,37 @@ async def get_participants(
...
+        query = f"""
+            SELECT
+                p.id,
+                p.reported_sex,
+                p.reported_gender,
+                p.karyotype,
+                p.meta,
+                {mt.external_id_query('pid')}
+            FROM participant p
+            LEFT JOIN participant_external_id pid
```
jmarshall, line 300: Standard abbreviation for this table is `peid`, and similarly `seid`.

db/python/layers/participant.py
```diff
@@ -620,6 +620,10 @@ async def get_external_participant_id_to_internal_sequencing_group_id_map(
             project=project, sequencing_type=sequencing_type
         )
 
+    async def participant_table_export(self, project: int):
```
jmarshall, line 623: One more trivial comment: Other functions here (and similarly for _sample.py_ and over in _../table/*.py_) have names of the form `actionverb_object`, whereas this is `object_verb` or maybe `adjectivalphrase_noun`. Could this be `export_participant_table` or so instead, or is there something being conveyed by this difference?

jmarshall (approved):
As usual, this looks so good to me that I have only a couple of mostly superficial comments :smile:

api/utils/parquet_route.py
dancoates: Ooh yeah, good pickup, that's fixed now

db/python/layers/participant.py
dancoates: Ah thanks for picking that up, I found the naming for this feature a bit tricky generally as I'm not 100% sure how it'll be used, so I changed the names a bunch of times and must have ended up on something out of step with existing naming accidentally.

PR #1004, opened 2024-12-19
SET-342 addition of sql querying UI to frontend

dancoates:
This is the second step in the addition of a experimental feature to metamist that allows safely querying project data using SQL on the frontend. 

This PR introduces some hooks that allow running sql queries against the project database, as well as a UI for writing SQL queries.


web/src/pages/report/SqlQueryUI.tsx
```diff
@@ -0,0 +1,440 @@
...
+                        {table.status === 'error' && (
+                            <Alert severity="error">{table.errorMessage}</Alert>
+                        )}
+                    </Collapse>
+                </Fragment>
+            ))}
+        </List>
+    )
+
+    const sidebarWidth = 250
```
milo-hyben, line 357: I would probably make numeric constant as all upper case and put it on the top of the file, but not a biggie.

milo-hyben (approved):
Excellent work Dan!
I have not used duckdb before so enjoy following the code how it gets all setup.
As well your comments make reviewing so much easier.
I only had one little thing, which can be ignored :-)

web/src/pages/report/SqlQueryUI.tsx
dancoates: Ah good pickup, that was some prototype code that slipped through :)
fixed now

PR #1011, opened 2025-01-10
Switch duckdb bundling method to load from CDN

dancoates:
The duckdb wasm files are not big, but not exactly tiny either. They sit at about 34 and 38MB, this is above the 32MB limit that cloudrun has for function responses. So this PR switches to loading them from a CDN until we can set up a better system for serving static files


milo-hyben (approved):
Make sense, let's use CDN version for now.
You might want to upgrade strawberry-graphql as well to avoid that security check failure.

dancoates:
> You might want to upgrade strawberry-graphql as well to avoid that security check failure.

Yeah would be good to clear that audit failure out, I'll do it in a separate PR though

PR #1014, opened 2025-01-13
SET-238 OurDNA Dashboard v2 prototype

dancoates:
This PR adds the prototype v2 version of the OurDNA dashboard.
The reporting components are structured in such a way that it should be easy to add reports to other projects.
Reports are defined in a declarative way so that it will be possible in the future to make this sort of reporting a 1st class citizen within metamist where the reports can be created via the UI and stored in the database.

At the moment there are 3 types of report items

- A Plot, which accepts a SQL query and a Observable Plot chart definition
- A Table which accepts a SQL query and displays the results in a table
- A Metric which accepts a SQL query and displays the results as one or more large numbers

To generate data for testing the dashboard, you can run the `test/data/generate_ourdna_data.py` script. 

The dashboard link will show up on the project summary page:
image

and is split into 4 tabs

image

image

image

image


test/data/generate_ourdna_data.py
```diff
@@ -1,118 +1,242 @@
...
 
-PARTICIPANTS = [
-    ParticipantUpsert(
-        external_ids={PRIMARY_EXTERNAL_ORG: 'EX01'},
-        reported_sex=2,
-        karyotype='XX',
-        meta={'consent': True, 'field': 1},
-        samples=[
+
+ANCESTRIES = [
```
ckrishk, line 21: Do we expect this list to be configurable by event/PM team ?

ckrishk:
Looks good to me on the first look 👍

ckrishk:
Ship it 🚢

ckrishk (approved):

test/data/generate_ourdna_data.py
dancoates: This is just for generating test data for OurDNA, the values don't matter much other than trying to be roughly indicative of what the real data will look like.

web/src/pages/report/reportIndex.ts
```diff
@@ -0,0 +1,45 @@
...
+    content: React.ComponentType
+}
+
+type ReportDefinition = {
+    title: string
+    tabs: ReportTab[]
+}
+
+export const reports: Record<string, Record<string, ReportDefinition>> = {
+    ourdna: {
```
milo-hyben, line 18: Can we ask for clarification, just put a comment here, this is the place where we link project to custom reports. We've got lost a bit on this one.

nevoodoo (approved):
Reviewed with MH :)

This is actually quite impressive work, I love how it has turned out! There were a few comments that Milo left on our behalf, but otherwise looks great!

milo-hyben (approved):
We had reviewed with @nevoodoo , very impressed, especially. CTE, amazing job.
We had just a few little comments.

web/src/pages/report/components/PlotFromQuery.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const result = useProjectDbQuery(project, query)
+
+    const [measureRef, { width, height }] = useMeasure<HTMLDivElement>()
+
+    const data = result && result.status === 'success' ? result.data : undefined
+
+    useEffect(() => {
+        if (!data) return
+        const plotOptions = getPlotOptions(data)
+        console.log(width, height)
```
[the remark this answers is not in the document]
dancoates: Ah thanks, good pickup, that's a leftover debug comment. I'll add a ticket to add a no-console.log linting rule to Jira.

web/src/pages/report/data/formatQuery.ts
```diff
@@ -0,0 +1,54 @@
...
+
+// Take a list of queries and combine them into a single query using CTEs
+// This allows viewing of steps that are used to arrive at a final aggregated query
+export function formatQuery(query: UnformattedQuery) {
+    if (typeof query === 'string') return stripIndent(query)
+    const cleanQueries = query.map((qq) => {
+        return {
+            ...qq,
+            // remove any trailing semicolons from the query, in case they were added accidentally
+            query: stripIndent(qq.query.replace(/;[\w\s]*$/g, '')),
```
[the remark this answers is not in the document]
dancoates: Yeah it's a bit of an edge case, but the flow for creating the queries for reports is often to first start by writing a query in the sql editor and then copy/paste it into a report definition. If that query ends up as a CTE as part of a larger query then it would cause errors. 
This is more of a future proofing thing for when we want to allow authoring reports via the UI where it could be more likely to happen.

web/src/pages/report/reportIndex.ts
dancoates: Yeah that's right, this is sort of an interim thing while we have custom reports but don't have them stored in the database, we need somewhere to store a listing of reports. The structure here works with the code here https://github.com/populationgenomics/metamist/blob/SET-238-dashboard-v-2-prototype/web/src/pages/report/ProjectReport.tsx#L9 which matches the params in the url to load the correct report.
It is also imported in the project summary page to show a list of reports for each project, matched on the project name.

Issue #1015, opened 2025-01-14
Cohort deprecation when containing deactivated sequencing groups


EddieLF:
Sequencing groups can be deactivated or marked as `active = False` for various reasons. When this happens, any cohorts containing this sequencing group remain static. This is expected, because cohorts should be immutable sets of sequencing groups. However, there is no real way to check if a cohort is deprecated due to containing old sequencing groups.

There should be some kind of validation to see that a cohort contains only active sequencing groups. This would be useful in the production pipelines, as analyses for a cohort should not be queued if the cohort contains deactivated sequencing groups.

To validate a cohort, the ability to check the active status of each sequencing group in the cohort will be necessary. Currently, the active status of a sequencing group can be checked in GraphQL only using the `activeOnly` filter. There is no boolean field that can be read for a given sequencing group. You must start a query from the `project` or `sequencingGroups` node to use the `activeOnly` filter, sequencing groups accessed through the `cohort` node will not be able to use this filter.

We could include the `activeOnly` bool filter offered to the `project` and `sequencingGroups` query nodes ([here](https://github.com/populationgenomics/metamist/blob/dev/api/graphql/schema.py#L551)) to the `cohorts.sequencingGroups` query node ([here](https://github.com/populationgenomics/metamist/blob/dev/api/graphql/schema.py#L183)). This would allow us to query a cohort's sequencing groups with `activeOnly: eq: False` - if any sequencing groups are returned, then the cohort could be considered deprecated and invalid.  

Or, we could allow for returning the active status of a sequencing group for all queries. The [SequencingGroupInternal model](https://github.com/populationgenomics/metamist/blob/dev/models/models/sequencing_group.py#L15) has this flag, called `archived`. However this is not propagated into the [GraphQL sequencing group model fields](https://github.com/populationgenomics/metamist/blob/dev/api/graphql/schema.py#L1097). We could add this field `archived` (or it's inverse, i.e. `active == not archived`) to the GraphQL model, so that sequencing groups returned by a query can be easily validated as active or not.

dancoates:
This functionality was added in #1016

PR #1016, opened 2025-01-14
Set 376 cohort sequencing group active filters

dancoates:
relating to #1015, expose archived property and active_only filter to allow checking validity of cohorts more easily.

### DIFF
--- a/api/graphql/schema.py
+++ b/api/graphql/schema.py
@@ -180,7 +180,10 @@ class GraphQLCohort:
 
     @strawberry.field()
     async def sequencing_groups(
-        self, info: Info[GraphQLContext, 'Query'], root: 'GraphQLCohort'
+        self,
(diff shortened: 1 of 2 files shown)


dancoates:
Probably do need to add some tests for this, we don't have any tests for archived sequencing groups in cohorts at the moment

vivbak (approved):

api/graphql/schema.py
```diff
@@ -1121,6 +1130,7 @@ def from_internal(
             platform=internal.platform,
             meta=internal.meta,
             external_ids=internal.external_ids or {},
+            archived=internal.archived,
```
vivbak, line 1133: hm linter is complaining here

vivbak (commented):
Ah sorry, approved before I took a look at the linter

PR #1020, opened 2025-01-21
SET-385 Allow archiving of sequencing groups via metamist api

dancoates:
This implementation allows for simple archiving of sequencing groups.
While this is not something that needs to happen often, there are cases
where sequencing groups are replaced by other samples and are not
automatically archived. In those cases it can be necessary to manually
archive the sequencing groups and also the sample and analysis. Sample
and analysis both already allow their "active" property to be updated so
sequencing groups were the odd one out by not allowing
deactivation/archiving.


nevoodoo (approved):

amyminiter:
Thanks so much @dancoates for implementing this so quickly. 

A note only: we discussed inconsistencies between naming for boolean attributes for `analysis` (active) and `sg` ('archived'. This implementation does maintain those inconsistencies but updating that seems like a larger piece of work for another time.

amyminiter (approved):
😄

dancoates:
> A note only: we discussed inconsistencies between naming for boolean attributes for analysis (active) and sg ('archived'. This implementation does maintain those inconsistencies but updating that seems like a larger piece of work for another time.

Yeah, always tricky to manage backwards compatibility with this sort of stuff, we can't just go through and rename everything to be consistent. Potentially we could support both `archived` and `active` but as you say it is probably a larger piece of work where we consider the implications of changing the status of each type of entity and how things cascade.

PR #1027, opened 2025-02-25
Format money using `toLocaleString()` instead of complicated regex

jmarshall:
I noticed looking at billing pages that current metamist prints currency amounts with commas both before and after the decimal point (see the foobar figure on the right with 4 dp):

Screenshot 2025-02-25 at 10 04 59

IMHO this makes the numbers harder to read (and suggests that there is a separate 789 number presented after the first). This was introduced accidentally during PR #957, which changed the regex to avoid a sonarqube diagnostic.

This rewrites the code using `toLocaleString()` instead of a complicated regex. I am no JavaScript expert, but I trust this is suitably portable — we use `DateTime.toLocaleString()` et al elsewhere, though I guess that is not exactly the same thing.

With this these figures are displayed as follows:

Screenshot 2025-02-25 at 10 04 52


dancoates:
Looks like this has been available since about 2016 haha, so I'm pretty out of date

PR #1035, opened 2025-03-27
Enable create_test_subset.py to update embedded sample IDs

jmarshall:
This splits out the _create_test_subset.py_ changes from PR #504. Currently implemented for CRAM files and gVCF files. Fixes [SET-491](https://cpg-populationanalysis.atlassian.net/browse/SET-491).

[SET-491]: https://cpg-populationanalysis.atlassian.net/browse/SET-491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -844,7 +844,12 @@ def get_random_families(
     return returned_families
 
 
-def rewrite_file(old_path: str, new_path: str, new_base_path: str, sid: tuple[str, str]):
+def rewrite_file(
+    old_path: str,
+    new_path: str,
+    new_base_path: str,
+    sid: tuple[str, str],
+):
     """
     Returns the shell commands necessary to rewrite embedded IDs as appropriate
     for each type of file likely to be encountered.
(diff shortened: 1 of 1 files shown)


scripts/create_test_subset.py
```diff
@@ -945,6 +1000,12 @@ def file_exists(path: str) -> bool:
     parser.add_argument(
         '--noninteractive', action='store_true', help='Skip interactive confirmation'
     )
+    parser.add_argument(
+        '--embedded-ids',
```
vivbak, line 1004: For the CLI argument, I'm wondering if we can call this something like `modify-crams` or `update-headers` to add more clarity to the action. What do you think @jmarshall ?

vivbak: Then the logic becomes,
`if update_headers`
which seems like it makes a little more sense to me

jmarshall: Good idea. It's now not just CRAMs and theoretically these IDs could be in not just headers, but `--update-embedded-ids` and `update_embedded_ids` would read better. And the usage message should note that it's on by default.

vivbak (approved):
LGTM! Thanks John! 🙏

jmarshall:
Uh oh… Back to the drawing board somewhat. _create_test_subset.py_ requires an image with the google cloud SDK installed, while this addition requires an image with `samtools` installed. So either we'll have to create an image with both or we'll have to refactor this bigly to submit separate samtools jobs (at which point we might as well use a bcftools image and separate bcftools jobs too)…

jmarshall:
Turns out our samtools image does have google-cloud-sdk on it (which gives you a `gcloud` command) but it does not have google-cloud-storage et al on it (which would give you the `google.cloud.storage` module the _create_test_subset.py_ script has always wanted).

That's only used in `file_exists()`, which could be rewritten to shell out to `gcloud` instead.

But it doesn't matter, because the next thing the script is going to want is `from metamist.apis import …` and no sensible samtools image is ever going to have metamist installed on it…

jmarshall:
Superseded by PR #1044.

PR #1036, opened 2025-04-01
Actions security overhaul

nevoodoo:
This PR introduces [`zizmor`](https://github.com/woodruffw/zizmor?tab=readme-ov-file) to our pre-commit checks to perform static analysis on our github workflow definitions. As part of this PR, I have also tried to clean up our workflow definitions based on the first set of feedback from the static analysis.

I have also set up [OIDC auth for PyPI](https://docs.pypi.org/trusted-publishers/) which makes the use of the PYPI_API_TOKEN redundant!

The initial scan with `zizmor` showed some initial issues with some of our actions. To summarise some of the issues identified:

- Overly broad permissions given - we had permissions granted to the workflow as a whole instead of each job, which gives all jobs access to the token, even if these jobs don't need it.
- We had some template injection risks
- The `actions/checkout` action stores the GITHUB_TOKEN in the `.git/config` directory unless you explicitly choose not to via `persist-credentials: false`.

<details>

<summary>Zizmor Analysis Results</summary>

```bash
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> ./.github/workflows/deploy.yaml:37:9
   |
37 |       - uses: actions/checkout@v4
   |         ------------------------- does not set persist-credentials: false
   |
   = note: audit confidence → Low

error[excessive-permissions]: overly broad permissions
  --> ./.github/workflows/deploy.yaml:11:3
   |
11 |   id-token: write
   |   ^^^^^^^^^^^^^^^ id-token: write is overly broad at the workflow level
   |
   = note: audit confidence → High

info[use-trusted-publishing]: prefer trusted publishing for authentication
   --> ./.github/workflows/deploy.yaml:128:9
    |
128 |         uses: pypa/gh-action-pypi-publish@release/v1
    |         -------------------------------------------- info: this step
129 |         with:
130 |           user: __token__
131 |           password: ${{ secrets.PYPI_API_TOKEN }}
    |           --------------------------------------- info: uses a manually-configured credential instead of Trusted Publishing
    |
    = note: audit confidence → High

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> ./.github/workflows/deploy_schema_updater.yaml:15:9
   |
15 |       - uses: actions/checkout@v4
   |         ------------------------- does not set persist-credentials: false
   |
   = note: audit confidence → Low

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> ./.github/workflows/lint.yaml:17:9
   |
17 |       - uses: actions/checkout@v4
   |         ------------------------- does not set persist-credentials: false
   |
   = note: audit confidence → Low

warning[excessive-permissions]: overly broad permissions
  --> ./.github/workflows/lint.yaml:5:3
   |
 5 | /   lint:
 6 | |     runs-on: ubuntu-latest
...  |
65 | |       - name: pre-commit
66 | |         run: pre-commit run --all-files
   | |                                        -
   | |________________________________________|
   |                                          this job
   |                                          default permissions used due to no permissions: block
   |
   = note: audit confidence → Medium

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> ./.github/workflows/renovate.yaml:24:9
   |
24 |         - name: Checkout
   |  _________-
25 | |         uses: actions/checkout@v4
   | |_________________________________- does not set persist-credentials: false
   |
   = note: audit confidence → Low

warning[excessive-permissions]: overly broad permissions
  --> ./.github/workflows/security_checks.yaml:1:1
   |
 1 | / name: Security Checks
 2 | |
...  |
52 | |           inputs: requirements-dev.txt
53 | |           summary: true
   | |________________________- default permissions used due to no permissions: block
   |
   = note: audit confidence → Medium

warning[excessive-permissions]: overly broad permissions
  --> ./.github/workflows/security_checks.yaml:36:3
   |
36 | /   pip-audit:
37 | |     runs-on: ubuntu-latest
...  |
52 | |           inputs: requirements-dev.txt
53 | |           summary: true
   | |                        -
   | |________________________|
   |                          this job
   |                          default permissions used due to no permissions: block
   |
   = note: audit confidence → Medium

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> ./.github/workflows/test.yaml:27:9
   |
27 |       - uses: actions/checkout@v4
   |         ------------------------- does not set persist-credentials: false
   |
   = note: audit confidence → Low

warning[artipacked]: credential persistence through GitHub Actions artifacts
   --> ./.github/workflows/test.yaml:132:9
    |
132 |         - uses: actions/checkout@v4
    |  _________-
133 | |         with:
134 | |           fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
135 | |
136 | |       # Download the coverage report artifact
    | |_____________________________________________- does not set persist-credentials: false
    |
    = note: audit confidence → Low

warning[excessive-permissions]: overly broad permissions
   --> ./.github/workflows/test.yaml:1:1
    |
  1 | / name: Test
  2 | | on:
...   |
153 | |       #   env:
154 | |       #     SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
    | |____________________________________________________- default permissions used due to no permissions: block
    |
    = note: audit confidence → Medium

warning[excessive-permissions]: overly broad permissions
   --> ./.github/workflows/test.yaml:14:3
    |
 14 | /   unittests:
 15 | |     name: Run unit tests
...   |
122 | |           script: |
123 | |             core.setFailed('Web failed to build with rc = ${{ steps.runtests.outputs.web_rc }}')
    | |                                                                                                -
    | |________________________________________________________________________________________________|
    |                                                                                                  this job
    |                                                                                                  default permissions used due to no permissions: block
    |
    = note: audit confidence → Medium

warning[excessive-permissions]: overly broad permissions
   --> ./.github/workflows/test.yaml:125:3
    |
125 | /   sonarqube:
126 | |     name: SonarQube scan
...   |
153 | |       #   env:
154 | |       #     SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
    | |                                                    -
    | |____________________________________________________|
    |                                                      this job
    |                                                      default permissions used due to no permissions: block
    |
    = note: audit confidence → Medium

info[template-injection]: code injection via template expansion
   --> ./.github/workflows/test.yaml:111:9
    |
111 |         - name: Fail if unit tests are not passing
    |           ---------------------------------------- info: this step
112 |           if: ${{ steps.runtests.outputs.rc != 0}}
113 |           uses: actions/github-script@v6
114 |           with:
115 | /           script: |
116 | |             core.setFailed('Unittests failed with rc = ${{ steps.runtests.outputs.rc }}')
    | |_________________________________________________________________________________________- info: steps.runtests.outputs.rc may expand into attacker-controllable code
    |
    = note: audit confidence → Low

info[template-injection]: code injection via template expansion
   --> ./.github/workflows/test.yaml:118:9
    |
118 |         - name: Fail if web build fails
    |           ----------------------------- info: this step
119 |           if: ${{ steps.runtests.outputs.rc != 0}}
120 |           uses: actions/github-script@v6
121 |           with:
122 | /           script: |
123 | |             core.setFailed('Web failed to build with rc = ${{ steps.runtests.outputs.web_rc }}')
    | |________________________________________________________________________________________________- info: steps.runtests.outputs.web_rc may expand into attacker-controllable code
    |
    = note: audit confidence → Low

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> ./.github/workflows/trigger_schema_updater.yaml:24:9
   |
24 |       - uses: actions/checkout@v4
   |         ------------------------- does not set persist-credentials: false
   |
   = note: audit confidence → Low

error[template-injection]: code injection via template expansion
  --> ./.github/workflows/trigger_schema_updater.yaml:45:9
   |
45 |   ...   name: 'Invoke Cloud Run'
   |         ^^^^^^^^^^^^^^^^^^^^^^^^ this step
46 | / ...   run: |
47 | | ...     CLOUD_RUN_URL=$(gcloud run services describe schema-updater --region australia-southeast1 --format 'value(status.url)')/execute-liquib...
48 | | ...     TOKEN=$(gcloud auth print-identity-token --impersonate-service-account="sample-metadata-deploy@sample-metadata.iam.gserviceaccount.com...
49 | | ...       curl -X POST -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/xml" --data-binary "@db/project.xml" $CLOUD_RUN_URL
   | |___________________________________________________________________________________________________________________________________________^ github.event.inputs.environment may expand into attacker-controllable code
   |
   = note: audit confidence → High

58 findings (40 suppressed): 0 unknown, 3 informational, 0 low, 13 medium, 2 high
```
</details>

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -11,7 +11,7 @@ permissions: {}
 
 jobs:
   unittests:
-    uses: './.github/workflows/test.yaml'
+    uses: ./.github/workflows/test.yaml
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
@@ -58,7 +58,7 @@ jobs:
 
       - uses: actions/setup-java@v4
         with:
-          distribution: 'temurin' # See 'Supported distributions' for available options
+          distribution: 'temurin' # yamllint disable-line rule:quoted-strings
           java-version: '17'
 
       - name: Setup build env
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -58,7 +58,7 @@ jobs:
 
       - uses: actions/setup-java@v4
         with:
-          distribution: 'temurin' # yamllint disable-line rule:quoted-strings
+          distribution: 'temurin' # See 'Supported distributions' for available options
           java-version: '17'
 
       - name: Setup build env
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -39,8 +39,8 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: 'Authenticate to Google Cloud'
-        uses: 'google-github-actions/auth@v2'
+      - name: Authenticate to Google Cloud
+        uses: google-github-actions/auth@v2
         with:
           workload_identity_provider: 'projects/774248915715/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
           service_account: 'sample-metadata-deploy@sample-metadata.iam.gserviceaccount.com'
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -7,9 +7,7 @@ on:
       - main
       - dev
 
-permissions:
-  id-token: write
-  contents: read
+permissions: {}
 
 jobs:
   unittests:
@@ -24,6 +22,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: unittests
     environment: production
+    permissions:
+      id-token: write
+      contents: read
     env:
       DOCKER_BUILDKIT: 1
       BUILDKIT_PROGRESS: plain
@@ -35,30 +36,32 @@ jobs:
         shell: bash -eo pipefail -l {0}
     steps:
       - uses: actions/checkout@v4
-
-      - id: "google-cloud-auth"
-        name: "Authenticate to Google Cloud"
-        uses: "google-github-actions/auth@v2"
         with:
-          workload_identity_provider: "projects/774248915715/locations/global/workloadIdentityPools/github-pool/providers/github-provider"
-          service_account: "sample-metadata-deploy@sample-metadata.iam.gserviceaccount.com"
+          persist-credentials: false
 
-      - id: "google-cloud-sdk-setup"
-        name: "Set up Cloud SDK"
+      - id: 'google-cloud-auth'
+        name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@v2'
+        with:
+          workload_identity_provider: 'projects/774248915715/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
+          service_account: 'sample-metadata-deploy@sample-metadata.iam.gserviceaccount.com'
+
+      - id: 'google-cloud-sdk-setup'
+        name: 'Set up Cloud SDK'
         uses: google-github-actions/setup-gcloud@v2
 
-      - name: "gcloud docker auth"
+      - name: 'gcloud docker auth'
         run: |
           gcloud auth configure-docker australia-southeast1-docker.pkg.dev
 
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: '3.11'
 
       - uses: actions/setup-java@v4
         with:
-          distribution: "temurin" # See 'Supported distributions' for available options
-          java-version: "17"
+          distribution: 'temurin' # See 'Supported distributions' for available options
+          java-version: '17'
 
       - name: Setup build env
         run: |
@@ -80,14 +83,14 @@ jobs:
               --new-version $(cat deploy/python/version.txt)dev$(echo $(git rev-parse HEAD) | cut -c1-7)
           fi
 
-      - name: "build deployable API"
+      - name: 'build deployable API'
         run: |
           export OPENAPI_COMMAND="java -jar openapi-generator-cli.jar"
           python regenerate_api.py
           ls -lGh metamist
 
         # also copies build artifacts to api/public
-      - name: "build web front-end"
+      - name: 'build web front-end'
         run: |
           set -eo pipefail
           pushd web
@@ -96,7 +99,7 @@ jobs:
           npm run build
           popd
 
-      - name: "build image"
+      - name: 'build image'
         run: |
           docker build \
             --build-arg SM_ENVIRONMENT=$SM_ENVIRONMENT \
@@ -107,11 +110,11 @@ jobs:
       - name: Build python package
         run: python setup.py sdist
 
-      - name: "push server image"
+      - name: 'push server image'
         run: |
           docker push $SM_DOCKER
 
-      - name: "deploy to Cloud Run"
+      - name: 'deploy to Cloud Run'
         run: |
           if [[ $GITHUB_REF == 'refs/heads/main' ]]; then
             gcloud_deploy_name=sample-metadata-api
@@ -127,7 +130,5 @@ jobs:
         if: github.ref == 'refs/heads/main'
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          user: __token__
-          password: ${{ secrets.PYPI_API_TOKEN }}
           packages-dir: dist/
           skip-existing: true
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -39,18 +39,16 @@ jobs:
         with:
           persist-credentials: false
 
-      - id: 'google-cloud-auth'
-        name: 'Authenticate to Google Cloud'
+      - name: 'Authenticate to Google Cloud'
         uses: 'google-github-actions/auth@v2'
         with:
           workload_identity_provider: 'projects/774248915715/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
           service_account: 'sample-metadata-deploy@sample-metadata.iam.gserviceaccount.com'
 
-      - id: 'google-cloud-sdk-setup'
-        name: 'Set up Cloud SDK'
+      - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v2
 
-      - name: 'gcloud docker auth'
+      - name: gcloud docker auth
         run: |
           gcloud auth configure-docker australia-southeast1-docker.pkg.dev
 
@@ -83,14 +81,14 @@ jobs:
               --new-version $(cat deploy/python/version.txt)dev$(echo $(git rev-parse HEAD) | cut -c1-7)
           fi
 
-      - name: 'build deployable API'
+      - name: build deployable API
         run: |
           export OPENAPI_COMMAND="java -jar openapi-generator-cli.jar"
           python regenerate_api.py
           ls -lGh metamist
 
         # also copies build artifacts to api/public
-      - name: 'build web front-end'
+      - name: build web front-end
         run: |
           set -eo pipefail
           pushd web
@@ -99,7 +97,7 @@ jobs:
           npm run build
           popd
 
-      - name: 'build image'
+      - name: build image
         run: |
           docker build \
             --build-arg SM_ENVIRONMENT=$SM_ENVIRONMENT \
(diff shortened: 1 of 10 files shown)


.github/workflows/deploy.yaml
```diff
@@ -35,30 +36,32 @@ jobs:
         shell: bash -eo pipefail -l {0}
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
-      - id: "google-cloud-auth"
-        name: "Authenticate to Google Cloud"
-        uses: "google-github-actions/auth@v2"
+      - id: 'google-cloud-auth'
```
jmarshall, line 42: Can we just **not quote** lots of these lines? Pretty sure quotes on `id`, `name`, and `uses` are all superfluous…

jmarshall:
Interesting. I guess there's some overlap between zizmor and actionlint, but this one is more security-focussed.

.github/workflows/deploy.yaml
nevoodoo: Good point. I think we can also remove the `id` itself, which is only needed if we reference this in future steps. I've removed quotes from all workflows now.

.github/workflows/test.yaml
```diff
@@ -25,14 +27,16 @@ jobs:
...
+          persist-credentials: false
 
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
 
       - uses: actions/setup-java@v4
         with:
-          distribution: 'temurin' # See 'Supported distributions' for available options
+          distribution: 'temurin' # yamllint disable-line rule:quoted-strings
```
dancoates, line 39: ditto

dancoates (commented):
A few little things, would also be interested to see what the output of zimzor looks like, any chance you could add the output from before you fixed the problems as a comment on the PR so there's a bit of context as to why the changes were made?

.github/workflows/deploy.yaml
```diff
@@ -35,30 +36,30 @@ jobs:
...
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: '3.11'
 
       - uses: actions/setup-java@v4
         with:
-          distribution: "temurin" # See 'Supported distributions' for available options
-          java-version: "17"
+          distribution: 'temurin' # yamllint disable-line rule:quoted-strings
```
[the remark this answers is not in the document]
nevoodoo: Good pick up!

.github/workflows/lint.yaml
```diff
@@ -15,14 +17,16 @@ jobs:
...
+          persist-credentials: false
 
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
 
       - uses: actions/setup-java@v4
         with:
-          distribution: 'temurin' # See 'Supported distributions' for available options
+          distribution: 'temurin' # yamllint disable-line rule:quoted-strings
```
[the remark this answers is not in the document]
nevoodoo: Good pick up!

.github/workflows/deploy.yaml
```diff
@@ -127,7 +128,5 @@ jobs:
         if: github.ref == 'refs/heads/main'
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          user: __token__
-          password: ${{ secrets.PYPI_API_TOKEN }}
```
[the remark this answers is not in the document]
nevoodoo: I set up OIDC through PyPI, which means we do not need to use these tokens anymore 🥳

https://docs.pypi.org/trusted-publishers/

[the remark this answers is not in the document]
dancoates: oh nice!

nevoodoo:
> A few little things, would also be interested to see what the output of zimzor looks like, any chance you could add the output from before you fixed the problems as a comment on the PR so there's a bit of context as to why the changes were made?

Good point, I've add the results to the PR description now for reference

dancoates (approved):

dancoates:
> > A few little things, would also be interested to see what the output of zimzor looks like, any chance you could add the output from before you fixed the problems as a comment on the PR so there's a bit of context as to why the changes were made?
> 
> Good point, I've add the results to the PR description now for reference

Brilliant, thank you!

jmarshall:
If all the (remaining) quote mark changes are due to your IDE being overeager rather than something that zizmor has complained about, then IMHO you should undo them so that reviewers and future us can see what the actual alleged security issues were.

The ones that have been updated to remove unnecessary quotes I am ambivalent about, so could stay but as an easily distinguishable commit (or even PR).

nevoodoo:
> If all the (remaining) quote mark changes are due to your IDE being overeager rather than something that zizmor has complained about, then IMHO you should undo them so that reviewers and future us can see what the actual alleged security issues were. Otherwise they're just churn.
> 
> The ones that have been updated to remove unnecessary quotes I am ambivalent about, so could stay but as an easily distinguishable commit (or even PR).

Excellent point, I've done this and the changes look a lot easier to review. TIL something, cheers!

.github/workflows/test.yaml
```diff
@@ -111,16 +115,20 @@ jobs:
       - name: Fail if unit tests are not passing
         if: ${{ steps.runtests.outputs.rc != 0}}
         uses: actions/github-script@v6
+        env:
+          rc: ${{ steps.runtests.outputs.rc }}
         with:
           script: |
-            core.setFailed('Unittests failed with rc = ${{ steps.runtests.outputs.rc }}')
+            core.setFailed('Unittests failed with rc = ${rc}')
```
jmarshall, line 122: Have you tested this?

nevoodoo: Yes, however I think this part is almost redundant. The test steps fail correctly anyway, so I don't even see why we need this. Do you have any ideas?

jmarshall: Prior to #518, these shell snippets ran without `set -e` so the failure status from `coverage run -m pytest` did not cause the `runtests` step to fail. Instead it recorded the exit status in `$rc` and carried on to run `coverage xml` and upload the coverage report etc.

These two `Fail if unit|web tests …` steps at the end would then cause the overall workflow to fail, after having uploaded the coverage reports etc.

This changed due to https://github.com/populationgenomics/metamist/pull/518#discussion_r1267852059 which is my fault and now I understand we were overriding the default `set -e`!

IMHO this is a bug and it would be better to go back to the original behaviour and upload coverage even if the unit tests fail (in fact this is part of the point!), which could be done with something like (`\` just for clarity):

```
rc=0
coverage run -m pytest --doctest-modules --doctest-continue-on-failure \
             test/ --junitxml=test-execution.xml || rc=$?
```

At this point we could get rid of the `defaults: run: shell` entry.

jmarshall: Meanwhile the bug that this “remediation” has introduced is that this now writes a status message of `Unittests failed with rc = ${rc}` (instead of `Unittests failed with rc = 1` or whatever), because JavaScript is not Shell.

nevoodoo: Makes sense, that is why my tests "passed" because it never got to the "Fail if.." steps. I guess I also had nothing to compare previous behaviour to since this bug has been around for a while.

Looking at the docs for `actions/github-script`:

```yaml
on: push

jobs:
  echo-input:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/github-script@v6
        env:
          FIRST_NAME: Mona
          LAST_NAME: Octocat
        with:
          script: |
            const { FIRST_NAME, LAST_NAME } = process.env

            console.log(`Hello ${FIRST_NAME} ${LAST_NAME}`)
```

should work. Thoughts on using this in a new PR? I can also remove the `defaults: run: shell` and fix the bug as per your suggestion:

```
rc=0
coverage run -m pytest --doctest-modules --doctest-continue-on-failure \
             test/ --junitxml=test-execution.xml || rc=$?
```

jmarshall: My thoughts are that that is insane overkill. The tool has a point that this could be problematic for user-supplied input, but this is not user-supplied input. (A PR could alter the previous step to change the output, but if a PR affects _.github/workflows/*_ then all bets are off anyway. I assume we have “don't run actions for PRs from external people” set on our repos, so we catch that by eyeballing PRs from weirdoes before allowing actions to run.)

If you would need to apply something like this so that the pre-commit checks pass… Well, that is why we should have had a policy discussion before a significant change like adding an opinionated tool like this to our _.pre-commit-config.yaml_.

nevoodoo: I've adopted those suggestions into #1039 and the changes are very minimal.

I conducted a test run on another PR I had open with these suggestions in:
https://github.com/populationgenomics/metamist/actions/runs/14210764029/job/39817420640

PR #1038, opened 2025-04-02
Change path key in fcf outputs dict

nevoodoo:
The outputs dict in the analysis object used a `basename` key to identify the file path. This is confusing as the resulting outputs file object contains both `path` AND `basename` properties. basename normally only contains the file name and extension, so use of this to denote a file path was a poor choice from myself :/

See the thread here: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1743555245218149

### DIFF
--- a/db/python/tables/output_file.py
+++ b/db/python/tables/output_file.py
@@ -153,25 +153,23 @@ class OutputFileTable(DbBase):
             if 'main_files' in files:
                 for primary_file in files['main_files']:
                     parent_file_id = await self.create_or_update_output_file(
-                        path=primary_file['basename'],
+                        path=primary_file['path'],
                         blobs=blobs,
                     )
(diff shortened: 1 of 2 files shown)


dancoates (commented):
This looks like a reasonable change, as long as the stuff testing the tests is pulled out into a separate PR once you're done testing it.

The main thing that needs consideration is the backwards incompatibility of this change. I found one place that `basename` is being used in our code: https://github.com/populationgenomics/cpg-methods-private/blob/1954e369d9248a9557cf44084a5956d0e2986603/scripts/bioheart/update_DRAGEN_analyses.py#L73 might be good for you to have a look through too to see if I missed any

dancoates:
Also noting there's a wip implementation here:
https://github.com/populationgenomics/production-pipelines/pull/1261/files
that would need to consider this change.

PR #1043, opened 2025-04-08
Analysis moving script update output not outputs

EddieLF:
One extra fix to the analysis moving script - the `AnalysisUpdateModel` is expecting the `output` to be updated, not the `outputs` dict. Updating the `output` path automatically populates the `outputs` dict via the output_files module.


nevoodoo:
> One extra fix to the analysis moving script - the `AnalysisUpdateModel` is expecting the `output` to be updated, not the `outputs` dict. Updating the `output` path automatically populates the `outputs` dict via the output_files module.

I think using `outputs` should work, only catch here being that you need to provide it in the form of:

```
        outputs = {
            'cram': {
                'basename': 'gs://fakegcs/file2.cram',
                'secondary_files': {
                    'meta': {'basename': 'gs://fakegcs/file2.cram.meta'},
                    'ext': {'basename': 'gs://fakegcs/file2.cram.ext'},
                },
            },
        }
```

But yes this change should also work!

nevoodoo (approved):

EddieLF:
Thanks for the note, in this case doing it with just `output` is a lot simpler, since the output_files will do the work in the background to populate `outputs` correctly.

PR #1044, opened 2025-04-09
Enable create_test_subset.py to update embedded sample IDs by creating a batch

### LINKED ISSUE #1034: Sample Subset needs to update file contents
https://github.com/populationgenomics/metamist/blob/042085917a57742604534f1ac3783cc0980d1fdc/scripts/create_test_subset.py#L842-L885

The current subset generation script copies files from `-main` to `-test`, and doesn't alter the file contents.

i.e. 
- for the SG CPG1 in `main`
- we create a new identity in `test` (CPG2)
- we copy the VCF/CRAM file from main to test with a rename (`CPG1.g.vcf` -> `CPG2.g.vcf`)

The issue here is that the file _contents_ are unchanged, so the file called `CPG2.g.vcf` has the Sample ID `CPG1` in the header. This affects downstream operations, e.g. the Pedigree generated from metamist contains the new sample IDs, the files contain the old ones, we can't perform analyses which require intersecting metadata and variant data

This copy section (actioning a gsutil copy with a name change) should be replaced with a hail batch job invocation, depending on the file type.

e.g. VCF -> localise the VCF from main into a container with bcftools, run a bcftools reheader, then write the new file. This output would be directed to the test location. This matches what our pipeline already does [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/jobs/genotype.py#L371-L375).

jmarshall:
A complete reimplementation of PR #1035. We now create jobs in a new batch for reheadering SampleIDs in CRAM and gVCF file headers. Fixes SET-491. Fixes #1034.

The previous attempt in PRs #504 and #1035 needed to run `samtools` commands within the same job as the _create_test_subset.py_ script itself is running. This was doomed to failure as the script (of course) needs to have the metamist API installed, which no reasonable image containing `samtools` would have installed.

Unless this is turned off via `--no-update-embedded-ids`, invoking this will now need to have a config containing the usual hail batch entries and `images.samtools` and `images.bcftools`. But when invoked via analysis-runner, you typically have all of those anyway!

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -23,7 +23,7 @@ from typing import Any, Tuple
 from google.cloud import storage
 
 from cpg_utils.config import image_path
-from cpg_utils.hail_batch import copy_common_env, get_batch
+from cpg_utils.hail_batch import get_batch
 
 from metamist.apis import AnalysisApi, AssayApi, FamilyApi, ParticipantApi, SampleApi
 from metamist.graphql import gql, query
@@ -944,7 +944,6 @@ def copy_files_in_dict(
                 logger.info(f'Adding job to rewrite {old_path} to {new_path}')
                 batch = get_batch()
                 job = batch.new_job(f'Reheader {new_path} from {old_path}')
-                copy_common_env(job)
                 main_file_commands(batch, job, old_path, new_path, sid_replacement)
                 job.storage(file_size(old_path) + 2 * 1024**3)  # Allow 2 GiB extra
             else:
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -16,7 +16,7 @@ import logging
 import os
 import random
 import subprocess
-from argparse import ArgumentParser
+from argparse import ArgumentParser, BooleanOptionalAction
 from collections import Counter
 from typing import Any, Tuple
 
@@ -194,7 +194,8 @@ def main(
     additional_families: set[str],
     additional_samples: set[str],
     cohorts: set[str],
-    skip_ped: bool = True,
+    skip_ped: bool,
+    update_embedded_ids: bool,
 ):
     """
     Script creates a test subset for a given project.
@@ -290,6 +291,7 @@ def main(
         project,
         old_sid_to_new_sid,
         sample_to_sg_attribute_map,
+        update_embedded_ids,
     )
     logger.info('Subset generation complete!')
 
@@ -475,6 +477,7 @@ def transfer_analyses(
     project: str,
     old_sid_to_new_sid: dict[str, str],
     sample_to_sg_attribute_map: dict[tuple, dict[tuple, str]],
+    update_embedded_ids: bool,
 ):
     """
     This function will transfer the analyses from the original project to the test project.
@@ -521,6 +524,7 @@ def transfer_analyses(
                             analysis['output'],
                             project,
                             (str(sg['id']), new_sg_map[s['externalId']][0]),
+                            update_embedded_ids,
                         ),
                         status=AnalysisStatus(
                             analysis['status'].lower().replace('_', '-')
@@ -539,6 +543,7 @@ def transfer_analyses(
                             analysis['output'],
                             project,
                             (str(sg['id']), new_sg_map[s['externalId']][0]),
+                            update_embedded_ids,
                         ),
                         status=AnalysisStatus(
                             analysis['status'].lower().replace('_', '-')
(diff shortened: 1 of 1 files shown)


jmarshall:
See batches [594084](https://batch.hail.populationgenomics.org.au/batches/594084/jobs/1) and [594085](https://batch.hail.populationgenomics.org.au/batches/594085) for an example of this running on test data.

dancoates (approved):
This looks great, nice work separating it out into multiple jobs

scripts/create_test_subset.py
```diff
@@ -862,10 +937,22 @@ def copy_files_in_dict(d, dataset: str, sid_replacement: tuple[str, str] = None)
...
+
         if not file_exists(new_path):
-            cmd = ['gcloud', 'storage', 'cp', old_path, new_path]
-            logger.info(f'Copying file in metadata: {" ".join(cmd)}')
-            subprocess.run(cmd, check=True)
+            if update_embedded_ids and sid_replacement and file_reheaderable(new_path):
+                logger.info(f'Adding job to rewrite {old_path} to {new_path}')
+                batch = get_batch()
+                job = batch.new_job(f'Reheader {new_path} from {old_path}')
+                copy_common_env(job)
```
MattWellie, line 947: This isn't strictly necessary - cpg-utils copies the ENV for every job in every batch
https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/hail_batch.py#L246

scripts/create_test_subset.py
```diff
@@ -839,7 +859,62 @@ def get_random_families(
...
+    elif new_path.endswith('.tbi'):
+        job.command(rf"""
+            tabix {job.new_file.main}
+            # Mention {job.new_file.tbi} for hail as tabix has no -o option
+        """)
+        batch.write_output(job.new_file.tbi, new_path)
+
+    elif new_path.endswith('.md5'):
+        job.command(rf"md5sum {job.new_file.main} | cut -d' ' -f1 > {job.new_md5}")
+        batch.write_output(job.new_md5, new_path)
```
MattWellie, line 909: I think the flow of this method is odd - we know that when reheadering the indexes and MD5s are going to be thrown away, so they're irrelevant to the copy, and files with those extensions can be ignored.

Then we can just generate a reheadered file, a corresponding index, and an md5 file, all on the primary VCF/CRAM, exporting all 3 results in the same resource group. A single block would read much more easily to me than splitting this across methods and making decisions per additional extension/filetype.

```python
old_file = batch.read_input(old_path)

if new_path.endswith('.vcf.gz'):
    job.image(image_path('bcftools_120'))
    job.declare_resource_group(
        new_file={
             'vcf.gz': '{root}.vcf.gz', 
             'vcf.gz.tbi': '{root}.vcf.gz.tbi'
             'vcf.gz.md5': '{root}.vcf.gz.md5'
        }
    )
    job.command(rf"""
        echo {sid[0]} {sid[1]} > $BATCH_TMPDIR/samples.txt
        bcftools reheader --samples $BATCH_TMPDIR/samples.txt --write-index -o {job.new_file["vcf.gz"]} {old_file}
        md5sum {job.new_file["vcf.gz"]} | cut -d' ' -f1 > {job.new_file["vcf.gz.md5"]}
    """)
    batch.write_output(job.new_file, new_path.removesuffix('.vcf.gz'))
elif new_path.endswith('cram'):
    job.image(image_path('samtools'))
    job.declare_resource_group(
        new_file={
             'cram': '{root}.cram', 
             'cram.crai': '{root}.cram.crai'
             'cram.md5': '{root}.cram.md5'
        }
    )
    job.command(rf"""
        echo {sid[0]} {sid[1]} > $BATCH_TMPDIR/samples.txt
        mv {old_file} _{job.new_file.cram}
        samtools reheader --no-PG --in-place --command 'sed /^@RG/s/\<SM:{sid[0]}\>/SM:{sid[1]}/g' {job.new_file.cram}
        md5sum {job.new_file.cram} | cut -d' ' -f1 > {job.new_file["cram.md5"]}
    """)
    batch.write_output(job.new_file, new_path.removesuffix('.cram'))
```

MattWellie (changes requested):

scripts/create_test_subset.py
MattWellie: I'd be happy scrapping the MD5 file - that's really only relevant when receiving files from external providers, GCP provides checksums. 

My point remains regarding code flow, and we absolutely want every index and cram indexed regardless of original status.

jmarshall: The current behaviour of the script is to only copy index and md5 files if they exist. My preference is to maintain this existing behaviour. (In practice, for index files that's pretty much the same thing as always producing them.)

MattWellie: Nothing about this PR is maintenance of a previous behaviour, it's generation of new data. Choosing to generate an index for new data based on invalidated files is illogical.

In the best case scenario an index _is_ generated here, but instead of one simple call & write there are two commands in the same job, applied in different methods, which each method writing half of a declared resource group. It's convoluted.

cassimons: I appreciate the thoughtful discussion here, but I will tie break. 

I can see @MattWellie’s point about clarity and a unified code path being preferable. That said, given the scale of the script is not massive I am comfortable proceeding with @jmarshall’s implementation given they both get us to the required result.

cassimons (approved):

PR #1045, opened 2025-04-09
SET-500 GraphQL in_ filter doesnt work for meta fields

dancoates:
The query building for meta filters created a comparison using the
mariadb JSON_EXTRACT operator. This worked for quite a few of the
comparison types but not all of them due to some funkyness with the way
JSON_EXTRACT works. JSON_EXTRACT will return a JSON value, so when we
pass in a value for comparison there is some inconsistencies with how
type coercion works.

For a row with a meta value like `{"sequencing_type": "genome", "size": 24470618112, "foobar": {"baz": "qux"}}`

```sql
JSON_EXTRACT(meta, $.sequencing_type) -- returns "genome" (with quotes)
```

and comparisons work as follows:

```sql
JSON_EXTRACT(meta, $.sequencing_type) = 'genome' -- TRUE

JSON_EXTRACT(meta, $.sequencing_type) IN ('genome')             -- TRUE
JSON_EXTRACT(meta, $.sequencing_type) IN ('genome', 'exome')    -- FALSE
JSON_EXTRACT(meta, $.sequencing_type) LIKE 'gen%'               -- FALSE
JSON_EXTRACT(meta, $.sequencing_type) LIKE '"gen%'              -- TRUE
```

initially it seemed like switching from JSON_EXTRACT to JSON_VALUE would
be a good move, but upon further investigation this could cause some
problems as JSON_VALUE will return null if the specified path doesn't
resolve to a primitive value.

eg.

```sql
JSON_EXTRACT(meta, $.foobar) IS NULL    -- FALSE
JSON_VALUE(meta, $.foobar) IS NULL      -- TRUE
```

The best solution seems to be wrapping JSON_EXTRACT in JSON_UNQUOTE
which will remove the quotes that are causing problems with IN and LIKE
queries but won't break isnull queries when querying nested JSON

### DIFF
--- a/db/python/filters/generic.py
+++ b/db/python/filters/generic.py
@@ -378,8 +378,14 @@ def prepare_query_from_dict_field(
             raise ValueError(f'Filter {field_name} must be a GenericFilter')
         if '"' in key:
             raise ValueError('Meta key contains " character, which is not allowed')
+
+        # using JSON_UNQUOTE is important here as by default JSON_EXTRACT will return
+        # a quoted string which means that filters like in_ and startswith will not
+        # work as the returned value starts with a " character. Strangely `eq` queries
+        # were working fine as mariadb was casting the filter value to a JSON string
+        # before comparing.
         fconditionals, fvalues = value.to_sql(
-            f"JSON_EXTRACT({column_name}, '$.{key}')",
+            f"JSON_UNQUOTE(JSON_EXTRACT({column_name}, '$.{key}'))",
             column_name=f'{column_name}_{key}',
         )
         conditionals.append(fconditionals)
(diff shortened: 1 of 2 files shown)


jmarshall:
```sql
JSON_EXTRACT(meta, $.sequencing_type) = 'genome' -- TRUE
```

Because it miraculously unquotes it for the `=` operator?

```sql
JSON_EXTRACT(meta, $.sequencing_type) IN ('genome')             -- TRUE
JSON_EXTRACT(meta, $.sequencing_type) IN ('genome', 'exome')    -- FALSE
```

Because with one thing in the `IN`, it optimises it to `=`. But with more than one it doesn't? Incredible!

jmarshall (approved):
Well-explained one-word change, and nice test cases added.

dancoates:
> Because with one thing in the `IN`, it optimises it to `=`. But with more than one it doesn't? Incredible!

Yeah I think that's the case, very frustratingly inconsistent!

PR #1046, opened 2025-04-10
Release

jmarshall:


dancoates (approved):

jmarshall:
Hmmm… I guess we ought to add a bumpversion seeing as @dancoates's change actually affects the core metamist code.

Issue #1050, opened 2025-04-17
FamilyApi returning wrongly formatted data


MattWellie:
Identified during the create_test_subset script. Minimal repro:

```
from metamist.apis import FamilyApi
fapi = FamilyApi()

families = fapi.get_families(
      project='acute-care',
     participant_ids=[27],
)
```

In the returned object the `family_id` is a `float`, not an `int`. This same issue doesn't appear to be present in swagger or graphQL.

Relevant failed batch: https://batch.hail.populationgenomics.org.au/batches/594279/jobs/1

PR #1051, opened 2025-04-17
fix: swap some problemmatic API client queries for GraphQL

MattWellie:
Proposed as a mitigation for #1050

That issue is that subset generation is broken due to FamilyApi returning inappropriate data.

GraphQL doesn't have the same issue, so this rewrites the FamilyApi and ParticipantApi queries to use GraphQL instead. This also updates (corrects?) the participant query to use the new multi-external-ID values

Unrelated: removes AssayApi which was imported and instantiated but not used

Concerns?!:

The building of a Family TSV for ingestion looks odd - it writes 4 header elements, only writes 3 data columns, and the header ordering doesn't match what the FamilyApi expects. This might be moot as in my minimal local running the columns were all blank anyway, but if they were filled I'm not sure the data would populate correctly.

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -25,7 +25,7 @@ from google.cloud import storage
 from cpg_utils.config import image_path
 from cpg_utils.hail_batch import get_batch
 
-from metamist.apis import AnalysisApi, AssayApi, FamilyApi, ParticipantApi, SampleApi
+from metamist.apis import AnalysisApi, FamilyApi, ParticipantApi, SampleApi
 from metamist.graphql import gql, query
 from metamist.models import (
     Analysis,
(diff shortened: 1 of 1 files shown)


scripts/create_test_subset.py
```diff
@@ -165,8 +164,12 @@
     query ($project: String!) {
         project (name: $project) {
             participants {
+                externalIds
```
MattWellie, line 167: we query for participants in the target project twice - one requires all these attributes, one only requires externalIds and id. This updates the one query to return all the data to both, negating the need for similar-but-different GraphQL queries

scripts/create_test_subset.py
```diff
@@ -190,6 +193,24 @@
     """
 )
 
+# take a parameter which is an array of ints
+QUERY_FAMILY_BY_INTERNAL_IDS = gql(
```
MattWellie, line 197: replacement of the seemingly busted FamilyApi query

vivbak (approved):
This looks good to me.

vivbak (changes requested):
Sorry, as per Milo's comment can we branch off dev instead of main. Then we can release into main.

PR #1054, opened 2025-04-29
Correction & slight restructure 

MattWellie:
One major correction - the GraphQL query I added in the previous change was a query on `family by internal ID`, which was incorrect. The required query is a search for `participants by internal ID`, then obtaining the family ID for each returned participant. This is changed in the query, and in the parsing of the returned result. 

A couple of other changes: 

* A `TODO` for discussion - I'm not convinced of the `cohort_samples_n` argument. Given that cohorts are a manually generated construct, an instruction to "copy this cohort" should use the whole cohort, instead of having a mandatory accompanying int to subset from that cohort.
* A bit of extra code to remove families with no samples from the family sampling method. Given the default seed my search for 10 families returned only 7 families with samples, which made debugging quite difficult. In general it makes sense to me that a family is not eligible for selection if it only exists on paper
* additional sampling using a number of random samples is partitioned into a separate method (as with the family and cohort selections)

An iterator change worked out debugging locally - iterating over family (dict key) instead of (family ID, familiy data)

A couple of extended/corrected docstrings, and a few more logging statements

@vivbak not sure if you're still the best point of contact for this, but as the most recent developer of this script.

Given the dev-main-release cycle here, I'm raising this as a draft, and mirroring the same script [here](https://github.com/populationgenomics/production-pipelines/pull/1203) (albeit without the diff, given it's a new script)

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -198,18 +198,47 @@ QUERY_FAMILY_BY_INTERNAL_IDS = gql(
     """
     query FamilyQuery($project: String!, $internal_ids: [Int!]!) {
   project(name: $project) {
-    families(
-      id: {in_: $internal_ids}
-    ) {
-      id
-      codedPhenotype
-      externalIds
-      description
+    participants(id: {in_: $internal_ids}) {
+      families {
+        id
+        codedPhenotype
+        description
+        externalIds
+      }
     }
   }
 }
 """
 )
+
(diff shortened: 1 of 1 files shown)


vivbak:
A note that the security checks are failing, this is a known issue and AFAIK shouldn't block the merge of a PR

vivbak (approved):

PR #1060, opened 2025-05-08
Add ability to specify phenotypes when updating participants

dancoates:
previously it wasn't possible to update phenotypes, or even to set phenotypes unless using the `generic_individual_metadata_importer` method via the /{project}/individual-metadata-manifest route endpoint

this change adds the ability to set phenotypes in the usual places that you update a participant or multiple participants, making it easier to manage phenotypes. This field is always optional so this change shouldn't have any effect on existing code that updates participants

This change relates to #1056 which aimed to allow copying of phenotypic information in the create_test_subset script but unfortunately didn't work because the API didn't support that operation


MattWellie:
I like this. I'm not 100% sure it _fully_ solves the copy test subset problem. This definitely solves parts of it, so any hangers on can probably be a separate issue...

Eyeballing create_test_subset again, the initial transfer of the samples is done by [this method](https://github.com/populationgenomics/metamist/blob/dev/scripts/create_test_subset.py#L795-L835) (??). A key thing here is that a tsv file is written then uploaded, and that contains `families.codedPhenotype` (always null). If that lands [here](https://github.com/populationgenomics/metamist/blob/1592a6ab8d9bc44302bec00af5cdff7dc34f80ff/db/python/layers/family.py#L344-L364), I'm not sure how to reconcile the family level codedPhenotype with the individual-level phenotypes. Given how little data we're feed into `import_families`, it looks like going on from there to pull in the per-participant phenotypes is an internal operation.

Once the data is initially transferred, we can re-run the same transfer with [skip_ped](https://github.com/populationgenomics/metamist/blob/dev/scripts/create_test_subset.py#L327-L336), which would hit the code path which is corrected in this PR. I think.

Lots of uncertainty here, but for ease of use it would be swell to see if the `import_families` operation can also bring over phenotypes for the family members.

dancoates:
Hm, yeah I'm not super across this test subset stuff, so not sure why it acts the way it does, but it appears to run that transfer families method and then use the family ids that it returns to run import_pedigree which does create the participants, but in a very minimal way https://github.com/populationgenomics/metamist/blob/SET-585-you-cant-update-participant-phenotypes-when-updating-a-participant/db/python/layers/family.py#L298

vivbak (approved):

PR #1064, opened 2025-05-16
update strawberry to fix pydantic version incompatibility

dancoates:
A recent update to uvicorn dependency resulted in pydantic being updated to 2.11, this update caused an incompatibility with our version of strawberry which was referencing an internal property `is_new_type` from within pydantic. This requirements update upgrades to the newest version of strawberry which contains a fix for this version incompatibility.

see https://github.com/strawberry-graphql/strawberry/issues/3807

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -289,6 +289,7 @@ packaging==24.2
     #   google-cloud-bigquery
     #   gunicorn
     #   pytest
+    #   strawberry-graphql
 pathspec==0.12.1
     # via black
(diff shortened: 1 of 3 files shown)


jmarshall (approved):
Strawberry change has been there since 0.262.4, for the record. LGTM.

dancoates:
> Strawberry change has been there since 0.262.4, for the record. LGTM.

Thanks, good pickup, I think still worth updating to the newest but I've added a note to the squash commit to record that decision.

PR #1066, opened 2025-05-21
Add options to export to csv/tsv formats in Billing UI

nevoodoo:
As per the PR title, this is a change that will add export functionality to some of the Billing UI pages:
- Cost Across Invoice Months (Topics only)
- Cost By Time
- Cost By Category
- Cost by Invoice Month


web/src/pages/billing/BillingCostByCategory.tsx
```diff
@@ -198,16 +198,89 @@ const BillingCostByCategory: React.FunctionComponent = () => {
         )
     }
 
+    const exportOptions = [
+        { key: 'csv', text: 'Export to CSV', value: 'csv' },
+        { key: 'tsv', text: 'Export to TSV', value: 'tsv' },
+    ]
+
+    const exportToFile = (format: 'csv' | 'tsv') => {
```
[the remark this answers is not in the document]
milo-hyben: I kind of agree with Copilot here, can we split the exportToFile function into two parts, 1. prepare data and 2.compose file content.
I would move the 2nd part (compose file) to shared/utilities or something like that and as the input could be a function or output of the part 1.

milo-hyben (changes requested):
Thank you Yash! 
Looking almost perfect, can we refactor a bit to remove the duplicated code, some suggestions in my comment.

violetbrina (changes requested):
Nice work Yash! I'll second (or third?) Milo's comment on a slight refactor just to clean things up a bit and put the repeated logic in a utility function. Otherwise, looking good!

web/src/pages/billing/BillingCostByCategory.tsx
```diff
@@ -198,16 +198,89 @@ const BillingCostByCategory: React.FunctionComponent = () => {
...
+            const costsForSku = dates.map((date) => {
+                // Find the data row for the date
+                const rowForDate = data.find((row) => {
+                    const d =
+                        row.date instanceof Date
+                            ? row.date.toISOString().slice(0, 10)
+                            : String(row.date)
+                    return d === date
+                })
+                if (
```
dancoates, line 233: can make this a fair bit cleaner with something like:

```ts
const value = rowForDate?.values?.[sku]
return typeof value === 'number' ? value.toFixed(2) : ''
```

web/src/pages/billing/BillingCostByCategory.tsx
```diff
@@ -198,16 +198,89 @@ const BillingCostByCategory: React.FunctionComponent = () => {
...
+        const skuSet: Set<string> = new Set()
+        data.forEach((row) => {
+            const dateStr =
+                row.date instanceof Date ? row.date.toISOString().slice(0, 10) : String(row.date)
+            dateSet.add(dateStr)
+            Object.keys(row.values).forEach((sku) => {
+                skuSet.add(sku)
+            })
+        })
+        const dates = Array.from(dateSet).sort()
```
dancoates, line 218: Not sure if this is a problem, but the exported data doesn't include the forecast data.

web/src/pages/billing/BillingCostByMonth.tsx
```diff
@@ -242,16 +243,90 @@ const BillingCostByTime: React.FunctionComponent = () => {
...
+        const allMonths = months.slice().sort()
+        const headers = ['Topic', 'Cost Type', ...allMonths]
+
+        // Prepare rows: for each topic, one row for compute and one row for storage, columns are months
+        const rows: any[] = []
+        Object.keys(data)
+            .sort((a, b) => a.localeCompare(b))
+            .forEach((topic) => {
+                // Storage Cost row
+                const storageRow: any = {
```
dancoates, line 262: No need to use any here, can have a better type:

```ts
const storageRow: {[key: string]: string} = {
```

web/src/pages/billing/BillingCostByMonth.tsx
```diff
@@ -242,16 +243,90 @@ const BillingCostByTime: React.FunctionComponent = () => {
...
+                    Topic: topic,
+                    'Cost Type': CloudSpendCategory.STORAGE_COST,
+                }
+                allMonths.forEach((month) => {
+                    storageRow[month] = (
+                        data[topic]?.[month]?.[CloudSpendCategory.STORAGE_COST] ?? 0
```
dancoates, line 268: Small thing, but to be consistent with what is shown in the table, I think we should leave the cell blank if we don't have a value for it rather than setting it to zero.

web/src/pages/billing/BillingCostByMonth.tsx
```diff
@@ -242,16 +243,90 @@ const BillingCostByTime: React.FunctionComponent = () => {
...
+                    ).toFixed(2)
+                })
+                rows.push(storageRow)
+
+                // Compute Cost row
+                const computeRow: any = {
```
dancoates, line 274: similar here, don't need to use `any`

web/src/pages/billing/BillingCostByTime.tsx
```diff
@@ -335,16 +335,69 @@ const BillingCostByTime: React.FunctionComponent = () => {
...
+
+        // Build rows: for each object in `data` (each date)
+        const rowData = data.map((row) => {
+            // Each `row` is { date, values }
+            const dateStr =
+                row.date instanceof Date ? row.date.toISOString().slice(0, 10) : String(row.date)
+
+            // Get values for each group, blank if not found
+            const vals = groups.map((group) =>
+                row.values[group] !== undefined && row.values[group] !== null
```
dancoates, line 355: rather than checking for what the value shouldn't be, it is less verbose to check for what it should be: `typeof row.values[group] === 'number'`

web/src/pages/billing/BillingCostByTime.tsx
```diff
@@ -335,16 +335,69 @@ const BillingCostByTime: React.FunctionComponent = () => {
...
+
+    const exportToFile = (format: 'csv' | 'tsv') => {
+        // Create header: Date plus group/column names
+        const headerData = ['Date', ...groups]
+
+        // Build rows: for each object in `data` (each date)
+        const rowData = data.map((row) => {
+            // Each `row` is { date, values }
+            const dateStr =
+                row.date instanceof Date ? row.date.toISOString().slice(0, 10) : String(row.date)
```
dancoates, line 351: similar comment re. the date here, row.date should always be a date so no need to check for it.

dancoates (changes requested):

dancoates (changes requested):
Sorry, lots of comments! Mostly little things, let me know if any questions.

web/src/pages/billing/BillingCostByCategory.tsx
```diff
@@ -198,16 +198,89 @@ const BillingCostByCategory: React.FunctionComponent = () => {
...
+                } else {
+                    return ''
+                }
+            })
+            return [sku, ...costsForSku]
+        })
+
+        // 3. Compose file content
+        const separator = format === 'csv' ? ',' : '\t'
+        const fileData = [headerData, ...rowData].map((row) => row.join(separator)).join('\n')
```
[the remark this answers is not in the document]
nevoodoo: Yeah sorry looks like I didn't refactor this after copying the skeleton from the Insights page, good pickup thank you!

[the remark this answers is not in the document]
dancoates: all good, hopefully papaparse is pretty easy to drop in :)

web/src/pages/billing/BillingCostByCategory.tsx
```diff
@@ -198,16 +198,89 @@ const BillingCostByCategory: React.FunctionComponent = () => {
...
+                    <div style={{ textAlign: 'right' }}>
+                        <Dropdown
+                            button
+                            className="icon"
+                            floating
+                            labeled
+                            icon="download"
+                            options={exportOptions}
+                            text="Export"
+                            onChange={(_, data) => exportToFile(data.value as 'csv' | 'tsv')}
```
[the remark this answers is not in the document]
nevoodoo: Yes! I was wondering why subsequent clicks weren't working, this was it!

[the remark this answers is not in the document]
nevoodoo: Either way, it was something I was going to look at eventually, but glad you picked it up

web/src/pages/billing/BillingCostByCategory.tsx
nevoodoo: @milo-hyben @violetbrina do we want to be including forecasted data in the exported CSV?

web/src/pages/billing/BillingCostByMonth.tsx
nevoodoo: I think there are inconsistencies in the table too where some cells are blank and some have $0.00 in them:

Screenshot 2025-05-21 at 4 06 53 PM

nevoodoo: I've patched the export functionality so it is similar to the table but this inconsistency remains for now

nevoodoo:
Thanks all! Very useful feedback, PR looking in much better shape now. I actually enjoyed this and got a lot of learning out of it :) 

Just waiting on a decision for whether forecasted data needs to be in the exports, and whether or not we want this functionality on 'Cost by Invoice Month' page.

web/src/pages/billing/BillingCostByCategory.tsx
milo-hyben: Those forecasted figures would need to be revisited anyway. It is using simple model which is not accurate. Happy if those are not exported at this stage.

milo-hyben (commented):
Hey Yash, looking much better with the reuse exportTable func!

Regarding your questions:
Q1. Whether forecasted data needs to be in the exports
Answer: No, no need to export at this stage
Q2. Whether or not we want this functionality on 'Cost by Invoice Month' page
Answer: Yes, we want on this page as well.

milo-hyben (approved):
Thank you Yash! This looks much better after the refactoring!

dancoates (approved):
Nice work! this is looking really good now

violetbrina (approved):
Fantastic @nevoodoo! If you do one final update to the Flask dep it will pass the security checks as well. Otherwise that can be another PR. You have my approval.

PR #1075, opened 2025-06-25
Update the ack deadline for etl pubsub

dancoates:
When the ETL pubsub triggers the cloudrun function it will only wait `ack_deadline_seconds` for the function to complete before re-triggering the function. For large ETL payloads it can take more than 30 seconds to process the file so this value needs to be larger


violetbrina (approved):
Fastest review ever. 5min > 30sec any day.

PR #1078, opened 2025-06-27
Document First Class Files

nevoodoo:
Our documentation of this has been scarce and this PR aims to remedy that by explaining how the logic works and where to find more information to debug in case this needs revisiting in the future.

### DIFF
--- a/api/graphql/schema.py
+++ b/api/graphql/schema.py
@@ -622,14 +622,14 @@ class GraphQLProject:
     @strawberry.field()
     async def sequencing_group_history(
         self, info: Info[GraphQLContext, 'Query'], root: 'GraphQLProject'
-    ) -> list['GraphQLSequencingGroupTypeByDate']:
+    ) -> list['GraphQLSequencingGroupsByDate']:
         loader = info.context['loaders'][
             LoaderKeys.SEQUENCING_GROUPS_COUNTS_FOR_PROJECT
         ]
 
         counts = await loader.load(root.id)
 
-        return GraphQLSequencingGroupTypeByDate.from_dict(counts)
+        return GraphQLSequencingGroupsByDate.from_dict(counts)
 
 
 @strawberry.type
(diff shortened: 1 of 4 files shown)


dancoates:
This is good, but I agree with Vivian that it could be restructured a bit to be clearer. Probably best to start with info from a user perspective of how to use the feature, and then go into how it works and db details later on.

docs/output_file_model_documentation.md
```diff
@@ -0,0 +1,563 @@
+# OutputFile Model Documentation
```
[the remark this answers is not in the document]
nevoodoo: I think these should be targeted with anyone interacting with an Analysis object from Metamist. So I'd say users/devs both?

vivbak (approved):
Wowza, I'm really impressed @nevoodoo. Huge improvement. Great stuff.

docs/output_file_model_documentation.md
```diff
@@ -0,0 +1,818 @@
...
+|------------------|-----------------|----------|
+| ✅ File exists and accessible | Full OutputFile object with `valid: true` | Normal operation |
+| ❌ File not found or inaccessible | Simple string (no OutputFile created) | Planned outputs, missing files, permission issues |
+
+> [!NOTE]
+>
+> **All OutputFiles have `valid: true`**: Because the code returns `None` for any file that fails validation (see `create_or_update_output_file()`), all OutputFile records in the database have `valid: true`. Files that can't be validated simply don't get OutputFile records.
+>
+> [!CAUTION]
+> **Matrix Table (.mt) files** are currently not supported. These Hail-specific files are treated as directories in GCS and cannot be validated using the standard file validation process.
```
dancoates: Would be good to be clear here what "not supported" means. Is the file marked as invalid?

docs/output_file_model_documentation.md
```diff
@@ -0,0 +1,818 @@
...
+
+### Parent-Child Relationships
+
+Many bioinformatics file formats require companion files (index files, checksum files, etc.). OutputFile models these as parent-child relationships.
+
+```mermaid
```
dancoates: Tiny thing, but this isn't readable in dark mode, I think if you set the text colour for the nodes it would fix it?

docs/output_file_model_documentation.md
```diff
@@ -0,0 +1,818 @@
...
+This section gets you started with the most common use cases for OutputFiles in Metamist.
+
+### Upload a Single File
+
+The simplest way to add an output file to an analysis is to include it in the `outputs` field when creating or updating an analysis.
+
+**REST API Example:**
+
+```bash
+POST /analysis/{project}/
```
dancoates: Should we be showing the graphql examples by default to encourage use of that rather than the REST api?

docs/output_file_model_documentation.md
```diff
@@ -0,0 +1,818 @@
...
+```
+
+**How it works:**
+
+- **Primary files** have `parent_id = null`
+- **Secondary files** reference their primary file via `parent_id`
+- In API responses, secondary files are nested under `secondary_files` in their parent's object
+
+> [!NOTE]
+> **`parent_id` in API Responses**: The `parent_id` field only appears in **primary files** (where it's set to `null`). Secondary files do **not** include the `parent_id` field in API responses, even though the relationship is tracked in the database via the `parent_id` foreign key.
```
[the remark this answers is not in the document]
nevoodoo: Yep, good point! Since the relationship is implicit in the JSON structure, we really don't need this field in the API response.

PR #1081, opened 2025-07-18
SET-655 add external ids to ourdna reports

dancoates:
Both the AGD (Australian Genome Database) and Participant Portal Ids are useful for CPG teams to trace samples through processing site data or participant portal data.


milo-hyben (approved):
Looking good!

PR #1083, opened 2025-07-22
Add metamist settings to graphql schema

dancoates:
These settings can be useful to know in the UI or in other usages of the metamist graphql client.
For example if you want to determine whether a certain ID is a Sample ID or a Sequencing Group ID.


milo-hyben (approved):
Looking good!

PR #1088, opened 2025-08-05
Bug: allow upsert of samples with assayless sequencing groups

amyminiter:
I am working on the ingestion of a large number of `analysis` files that have been generated from long read data. These files have been produced on external pipelines for which the CPG does not have assay data (`bam` or `fastq`). Long read sequencing was performed for a small subset of samples (n=25) from an existing CPG cohort. We have existing `samples` and `participant` metadata previously ingested into metamist for this subset.

To ingest these data, the following steps are required:
* Create a new sequencing group for each `sample` with `type`, `technology` and `platform` to reflect the long read data being ingested
* Create a new `cohort` associated with all newly created sequencing groups to accommodate two cohort-level `analysis` objects
* For each sample-level `analysis` file provided, create a new `analysis` object associated with their corresponding `sample's` new `sequencing group`

## Current functionality

Currently, the `create_sample` endpoint is configured to accept an empty or missing list of assays. Yet in the `sequencing group` `layer` module, the following error is returned when attempting to upsert a sample that includes a sequencing group with no assays:
```
\"/Users/amymin/Documents/Code/metamist-local/api/routes/sample.py\", line 178, in update_sample\n    await st.upsert_sample(sample.to_internal())\n  File \"/Users/amymin/Documents/Code/metamist-local/db/python/layers/sample.py\", line 276, in upsert_sample\n    await sglayer.upsert_sequencing_groups(sample.sequencing_groups)\n  File \"/Users/amymin/Documents/Code/metamist-local/db/python/layers/sequencing_group.py\", line 350, in upsert_sequencing_groups\n    assay_ids = [a.id for a in sg.assays]\n                ^^^^^^^^^^^^^^^^^^^^^^^^^\nTypeError: 'NoneType' object is not iterable\n"}
```
The `sequencing group` `tables` module also does not accommodate an empty list of `assays` when inserting linked `sequencing groups` and `assays` into metamist.

## Updates in this PR
Very minor! Small checks are performed in the `layers` and `tables` modules for an empty list of `assays`. If `None`, the method to upserting the linking table `assays_sequencing_groups` is not called.

### DIFF
--- a/db/python/layers/sequencing_group.py
+++ b/db/python/layers/sequencing_group.py
@@ -347,7 +347,7 @@ class SequencingGroupLayer(BaseLayer):
         # but we're inside a transaction, so it's not actually committing anything
         # so should be quick to "write" in serial
         for sg in to_insert:
-            assay_ids = [a.id for a in sg.assays]
+            assay_ids = [a.id for a in sg.assays] if sg.assays else None
             sg.id = await self.seqgt.create_sequencing_group(
                 sample_id=sg.sample_id,
                 type_=sg.type,
--- a/db/python/layers/sequencing_group.py
+++ b/db/python/layers/sequencing_group.py
@@ -347,7 +347,7 @@ class SequencingGroupLayer(BaseLayer):
         # but we're inside a transaction, so it's not actually committing anything
         # so should be quick to "write" in serial
         for sg in to_insert:
-            assay_ids = [a.id for a in sg.assays] if sg.assays else None
+            assay_ids = [a.id for a in sg.assays] if sg.assays else []
             sg.id = await self.seqgt.create_sequencing_group(
                 sample_id=sg.sample_id,
                 type_=sg.type,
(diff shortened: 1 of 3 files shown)


db/python/layers/sequencing_group.py
```diff
@@ -347,7 +347,7 @@ async def upsert_sequencing_groups(
         # but we're inside a transaction, so it's not actually committing anything
         # so should be quick to "write" in serial
         for sg in to_insert:
-            assay_ids = [a.id for a in sg.assays]
+            assay_ids = [a.id for a in sg.assays] if sg.assays else None
```
dancoates, line 350: The table `create_sequencing_group` method is expecting assay_ids to be of type `list[str]`, so we shouldn't pass None here. The check in the table method will work with an empty list too, so you should be able to replace `None` with `[]` here.

amyminiter:
@dancoates I've modified the existing test for upserting objects. This tests with three dummy participants and in the current form, all have a non-Empty list of sequencing groups to upsert. I removed the assays from the final participant, Athena, and added a query with a comical number of joins to test len of Athena's upserted sequencing groups. There might be a better way!

Could we go over in Software Hotdesk? I don't have MariaDB configured to run in Docker so will be easy to get this off the ground if you could test the test for me.

dancoates (approved):

PR #1090, opened 2025-08-06
SET-547 Custom Project Groupings

nevoodoo:
This change adds the option to select multiple projects/topics to group by. The key themes we wanted to address:
- [x] multiple project and topic selections when querying data
- [x] saving selections in the URL params
- [x] quick load times and feedback where possible

The following components have been added/changed for this:

New components/models:
- **MultiFieldSelector**: A MUI dropdown that allows picking multiple options
- **BillingRunningCostQueryModel**: Python model for improved querying with filters (did not really end up using this as we opted for client side filtering here but I had already worked on the model and left this in case we need it for future use cases. We have a BillingTotalCostQueryModel so it also makes sense to have this one.)

Modifications:
- **BillingInvoiceMonthCost**: added server side filtering
- **BillingCostByMonth**: changed Column Visibility button text to use `Topics` instead of `Fields`. Added server side filtering
- **BillingCostByTime**: added server side filtering and added breakdown views
- **BillingCostByCategory**: added server side filtering for multiple projects
- **ColumnVisibility**: make button text customizable by passing props
Additionally, I've added a query model for filtering through the `Running Costs` data, for better typing on the UI. 
- **HorizontalStackedBarChart**: the chart originally resized to use the whole screen but this is not visually appealing when you have fewer projects selected. You end up with oversized bars so this has been updated to be more responsive.


milo-hyben (approved):
Looking good Yash!
We just need to do a lot of testing once we move to dev, each report should be tested as  there have been a lot of changes.

web/src/pages/billing/BillingCostByCategory.tsx
```diff
@@ -93,24 +176,183 @@ const BillingCostByCategory: React.FunctionComponent = () => {
...
+            selectedPeriod,
+            start,
+            end,
+            recs.value === BillingColumn.GcpProject ? selectedProjects : [],
+            recs.value === BillingColumn.Topic ? selectedTopics : []
+        )
+        // Show loading immediately to prevent flash of old data
+        if (Boolean(start) && Boolean(end)) {
+            setIsLoading(true)
+            const query = buildFilteredQueryModel()
```
dancoates, line 199: These changes to use `buildFilteredQueryModel()` are relying on the updated state being available immediately, which won't be the case. You'll notice that when testing this in the UI whenever you select something it will use the previous value. You'll need to pass the updated value along to the call to buildFilteredQueryModel.
See this bit of the docs for more info: https://react.dev/reference/react/useState#ive-updated-the-state-but-logging-gives-me-the-old-value

web/src/pages/billing/BillingCostByMonth.tsx
```diff
@@ -173,7 +187,22 @@ const BillingCostByTime: React.FunctionComponent = () => {
         return CloudSpendCategory.COMPUTE_COST
     }
 
-    const getData = (query: BillingTotalCostQueryModel) => {
+    // Pre-fetch topics data on component mount
+    React.useEffect(() => {
+        const preFetchTopics = async () => {
```
dancoates, line 192: Similar comment in this to the category page component re using promise chaining rather than defining a function

web/src/pages/billing/BillingCostByMonth.tsx
```diff
@@ -426,6 +484,7 @@ const BillingCostByTime: React.FunctionComponent = () => {
                                 minWidth: '115px',
                                 height: '36px',
                             }}
+                            buttonText="Topics"
```
dancoates, line 487: Can we remove the column visibility select as it is kind of doublling up with the topic selection

dancoates (changes requested):

dancoates (changes requested):
Nice work on this PR, there's a lot of really good stuff in here. It's a pretty big one so quite a few comments and a few bugs that need fixing. Let me know if any questions.

web/src/pages/billing/components/BillingCostByTimeTable.tsx
```diff
@@ -23,7 +44,19 @@ interface IBillingCostByTimeTableProps {
     setVisibleColumns: (columns: Set<string>) => void
     expandCompute?: boolean
     setExpandCompute?: (expand: boolean) => void
-    exportToFile: (format: 'csv' | 'tsv') => void
+    exportToFile: (format: 'csv' | 'tsv', exportData?: ExportData) => void
+    groupBy?: BillingColumn
+    selectedProjects?: string[]
+    breakdownData?: { [date: string]: { [field: string]: { [category: string]: number } } }
+    openRows?: string[]
+    handleToggle?: (date: string) => void
```
dancoates, line 52: handleToggle isn't used in this component, should be removed as a prop

web/src/pages/billing/BillingInvoiceMonthCost.tsx
```diff
@@ -477,7 +702,7 @@ const BillingCurrentCost = () => {
                     <Grid>
                         <Grid.Column width={12}>
                             <HorizontalStackedBarChart
-                                data={costRecords}
+                                data={chartData}
```
dancoates, line 705: I can't comment on the actual line cos it isn't changed but it'd be good to have a loader when in table view when things are loading. Replacing:

```tsx
{!showAsChart ? (
    <Table celled compact sortable>
```

with 

```tsx
{!showAsChart && isLoading ? <LoadingDucks /> : null}

{!showAsChart && !isLoading ? (
    <Table celled compact sortable>

```

seems to work

web/src/pages/billing/BillingCostByTime.tsx
```diff
@@ -51,182 +72,480 @@ const BillingCostByTime: React.FunctionComponent = () => {
...
+                end,
+                expandCompute,
+                visibleColumns,
+                selectedProjects,
+                selectedTopics
+            )
         }
     }
 
+    const onProjectsSelect = (
```
dancoates, line 224: I'm not sure exactly where the bug is, but filtering by project/topic doesn't seem to be working at the moment, it updates the URL and if you then refresh the page the data updates - but new data isn't fetched when you select topics/projects

dancoates (changes requested):
This is really close! just one bug that needs fixing and a couple of little things otherwise

dancoates (approved):

PR #1092, opened 2025-08-12
Set 554 R1 Create a new billing page that ignores topics and drills down per sg

dancoates:


milo-hyben (approved):
Looking good Dan!
We just need to watch the file BillingCostBySample.tsx, if more functionality needed in the future (e.g. forecasting) we might consider splitting / refactoring. Will leave as is for now.

PR #1093, opened 2025-08-13
POPGEN-877 Sequencing group upsert returning None values

amyminiter:
`db/python/tables/sequencing_group.py` was returning the id of newly created sequencing groups, but this value was not being returned from any of its calling functions. As a result, a sequencing_group id of None was being returned via db/python/layers/sequencing_group.py upsert_sequencing_groups(). This also extends to any upserted metadata as only the 

This PR fixes that bug by ensuring that returned values are captured in routes and layers modules for sample and sequencing group objects.

### DIFF
--- a/db/python/layers/sequencing_group.py
+++ b/db/python/layers/sequencing_group.py
@@ -357,11 +357,6 @@ class SequencingGroupLayer(BaseLayer):
                 assay_ids=assay_ids,
                 open_transaction=False,
             )
-            # Now save this new id to sequencing_groups; we want these values returned
-            # sample_id maps sgs from to_insert with those in sequencing_groups
-            for sg_original in sequencing_groups:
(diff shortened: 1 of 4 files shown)


db/python/layers/sequencing_group.py
```diff
@@ -357,6 +357,11 @@ async def upsert_sequencing_groups(
                 assay_ids=assay_ids,
                 open_transaction=False,
             )
+            # Now save this new id to sequencing_groups; we want these values returned
+            # sample_id maps sgs from to_insert with those in sequencing_groups
+            for sg_original in sequencing_groups:
```
amyminiter, line 362: I _really_ tried to improve the naming of variables in this loop. But extremely difficult without modifying variable names in the surrounding function. Extremely open to suggestions.

PR #1097, opened 2025-08-20
Audit refactor

EddieLF:
# Metamist Audit refactor

The audit module is intended to 

1. Scan the dataset's upload bucket for files
2. Identify if the files can be deleted by analysing records in Metamist
3. Identify if any files should be ingested
4. Write reports about its findings
5. Delete the files which have been identified
6. Write a report about the files deleted and record this in an analysis record

## Purpose

This PR completely refactors and re-implements the Metamist audit module. 

The previous version of this module was poorly structured and difficult to understand and maintain. The refactor should focus on improving its architecture, readability, and testability. I used Claude Opus 4.1 in VSCode to help with improving the architecture during this refactor.

## Proposed Changes

- Implement "Separation of Concerns" design principles to ensure each component of the module is handling only the responsibilities in it's scope
- Decouple components of the module to promote "Dependency Injection"
- Use dataclasses and frozen configurations for everything where possible
- Implement business logic through pure functions containing no I/O or external dependencies
- Implement better file matching mechanisms through flexible and composable strategies that compare with rigour, e.g. checksum comparisons 
- Abstract the data access methods into clean and well defined classes

I recommend reviewing the new README to understand how these goals have been met by the changes in this refactor.

## TODO Checklist

- [x] Improve the `audit_metadata.tsv` which is written to record the input config - probably save to a json, or write to the analysis.meta
- [x] Save the reports outside of the main-upload bucket
- [x] Rename report "files_to_ingest" -> "files_to_review"
- [x] Add new entrypoint`"annotate_audit_report` which filters and marks files for deletion and adds justification
- [x] Remove typehints from Typing
- [x] Write analysis records containing the audit result deletion
- [x] Create new / update existing unit tests

## Future scope

- Implement some method (e.g. cron job, scheduled cloud run job) to run this module in an automated fashion at regular intervals
- Implement some ways to better track and display audit results (e.g. Slack posts, or display in the Metamist webUI)
- Add in a Slack connector

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
-# pylint: disable=too-many-instance-attributes,too-many-lines,too-many-locals
+# pylint: disable=too-many-instance-attributes,too-many-locals
+# pylint: disable=C0302
 
 """ Example Invocation
 


metamist/audit/data_access/__init__.py
MattWellie, line 1: In general I prefer `__init__` being empty, I don't think this content offers any benefits. This imports would be still be `from metamist.audit.data_access import MetamistDataAccess` with or without it

The `__all__` attribute is only useful for `from x import *` importing, which should be avoided

metamist/audit/adapters/storage_client.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def delete_blobs(
+        self,
+        paths: List[str],
+    ):
+        """
+        Delete blobs from Google Cloud Storage.
+        """
+        for path in paths:
+            self.delete_blob(path)
```
MattWellie, line 218: IDK how much real time would be saved here, but if audits are per-bucket, this multiple delete could be more efficient by creating a single instance of `self.client.get_bucket(bucket_name)` (in self.delete_blob) and rolling multiple deletes on the same instance. Currently this creates a new client.bucket with each file.

PR #1105, opened 2025-09-05
Clean up metamist readme and installation instructions

dancoates:
The metamist installation instructions had got pretty out of control, with bits and pieces added over the years that provided too many options and made them hard to follow for someone new to the project. This update cleans up and streamlines the documentation so that it should be easier to follow. 

- Make it clear that local installation is not required to work with metamist python client and production data
- Remove windows installation instructions
- Remove unnecessary depdendency installation instructions, wget, java etc. These are installed automatically by homebrew
- Move web client instructions to the bottom so that anyone who only needs the API set up doesn't have to read or follow them.
- Only document recommended installation options in detail. The previous version was very confusing as it wasn't clear what path to follow. Anyone who wishes to install in an alternative way will be able to work out the steps from the base instructions.
- Switch usages of mysql cli to mariadb cli


docs/installation.md
```diff
@@ -1,196 +1,79 @@
...
-Add this to your `.zshrc` to auto-load on next shell session.
-
-If you installed `nvm`, you will need to add lazy load since `nvm` has high load penalties.
+### Database setup
 
-- For Oh-My-Zsh users, you can just add the `nvm` plugin to your `.zshrc` via these [instructions](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/nvm/README.md)
+#### Running MariaDB
 
-- If you do NOT have Oh-My-Zsh, you can use this [plugin](https://github.com/undg/zsh-nvm-lazy-load):
+Metamist uses a MariaDB 11 database. Docker is probably the easiest way to run the metamist MariaDB database locally.
```
vivbak, line 46:
```suggestion
Metamist uses a MariaDB 11 database. Docker is the easiest way to run the Metamist MariaDB database locally.
```
Nit, instructional rather than narrative.

docs/installation.md
```diff
@@ -1,196 +1,79 @@
...
-```shell
-git clone https://github.com/undg/zsh-nvm-lazy-load $ZSH/custom/plugins/zsh-nvm
-
-#Add this to your plugins variable in the `.zshrc` file and then source the file.
-plugins=(... zsh-nvm-lazy-load)
-```
+We have found that [OrbStack](https://orbstack.dev/) is faster and easier to use than [Docker Desktop](https://docs.docker.com/desktop/) but either should work fine.
 
-Once set up, install the OpenAPI Generator:
+It is also possible to install MariaDB stanalone using [homebrew](brew.sh) or another package manager.
```
vivbak, line 50:
```suggestion
You can also install MariaDB standalone using [homebrew](brew.sh) or another package manager.
```
Nit, typo, tighten.

docs/installation.md
```diff
@@ -1,196 +1,79 @@
...
-```shell
-# Install npm via nvm
-nvm install --lts
-npm install @openapitools/openapi-generator-cli -g
+```bash
+docker run --name mariadb-p3306 -e MYSQL_ALLOW_EMPTY_PASSWORD=1 -p 3306:3306 -d docker.io/library/mariadb:11.7.2
 ```
 
-Add this to your `.zshrc`:
+If you need to run on a different port as `3306` is already taken you can specify that in the port mapping.
```
vivbak, line 58:
```suggestion
If port 3306 is already in use, you can specify a different port in the mapping. For example:
```
nit, tighten

vivbak (approved):
LGTM <3

PR #1106, opened 2025-09-09
SET-547-R3 Custom Project Grouping Fixes

nevoodoo:
Fixing a few UI bugs here on the recent Custom Project Groupings feature:

### CostByTime
- Dark Mode in the table component
- Group by Topic data not displaying when in breakdown view
- Add % to the Donut chart slices
- Fix Column Visibility Dropdown not working

### BillingInvoiceMonthCost
- Column positioning issue when using the column visibility dropdown on a previous month
- Rounding off logic was causing aggregations to be off by $0.01 as the total and the individual values were rounded off separately
- Make consistent aggregation text on the project view to use "All GCP Projects" instead of "Total (X Projects)" since all other views use "All XYZ" format
- Changing the Group By was setting off a debounced query instead of an immediate query to the API

### CostByMonth
- Improve error message when selecting incompatible start and end dates on CostByMonth
- Use consistent ordering of Compute Type in the exported CSV/TSV files i.e Compute Cost and _then_ Storage Cost

### DIFF
--- a/db/python/tables/bq/billing_base.py
+++ b/db/python/tables/bq/billing_base.py
@@ -391,7 +391,6 @@ class BillingBaseTable(BqDbBase):
         total_monthly_category: dict,
         total_daily_category: dict,
         results: list[BillingCostBudgetRecord],
-        filters: dict | None = None,
     ) -> list[BillingCostBudgetRecord]:
         """
         Add total row: compute + storage to the results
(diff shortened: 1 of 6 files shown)


dancoates (approved):
Looks great! I've tested this against the reports from testing on dev, and the only one that doesn't seem to be addressed is for the cost by time page csv export date formatting being different from production. I think the change to have it consistent with what is in the table is a reasonable one so all good to leave as is.

PR #1113, opened 2025-09-17
SET-438 report error message when you don t have access to a project is confusing

hrrsheen:
When viewing a custom report or dashboard, if the user doesn’t have access, the page won’t load properly and the error message that is displayed is unhelpful.


dancoates (approved):
This looks good, nice work updating the viewer context to send through the loading and error state too.

PR #1119, opened 2025-09-22
SET-553 R1 display average per sample storage costs each month

milo-hyben:
This PR adds new category cost (Avg Sample Storage Cost) to Billing Cost By Month report.
It is using simple logic:
[Avg Sample Storage Cost] = [Cloud Storage] / [No of samples]
where [No of samples] is calculated using audit log information and it is accumulated over the months.
 
Both Backend and Frontend have been updated to show the new row.
First row in the report (All Topics) does not include this category as it is not relevant, e.g. some topics do not have samples.

### DIFF
--- a/db/python/tables/sample.py
+++ b/db/python/tables/sample.py
@@ -2,6 +2,8 @@ import asyncio
 from datetime import date
 from typing import Any, Iterable
 
+from dateutil.relativedelta import relativedelta
+
 from db.python.filters import GenericFilter
 from db.python.filters.sample import SampleFilter
 from db.python.tables.base import DbBase
@@ -247,12 +249,16 @@ class SampleTable(DbBase):
                 'participant_id': row['participant_id'],
                 'type': row['type'],
                 'active': parse_sql_bool(row['active']),
-                'sample_root_id': sample_id_format(row['sample_root_id'])
-                if row['sample_root_id']
-                else None,
-                'sample_parent_id': sample_id_format(row['sample_parent_id'])
-                if row['sample_parent_id']
-                else None,
+                'sample_root_id': (
+                    sample_id_format(row['sample_root_id'])
+                    if row['sample_root_id']
+                    else None
+                ),
+                'sample_parent_id': (
+                    sample_id_format(row['sample_parent_id'])
+                    if row['sample_parent_id']
+                    else None
+                ),
             },
             has_external_ids=True,
             has_meta=True,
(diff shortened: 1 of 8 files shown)


db/python/tables/sample.py
```diff
@@ -687,3 +693,55 @@ async def get_samples_with_missing_participants_by_internal_id(
...
+        values = {'project_ids': project_ids} if project_ids else {}
+
+        _query = f"""
+        WITH t AS(
+            SELECT project, id, min(row_start) as sample_first_date
+            FROM sample FOR SYSTEM_TIME ALL
+            {where_str}
+            GROUP BY project, id
+        )
+        SELECT project, CAST(EXTRACT(YEAR FROM sample_first_date) AS INTEGER) AS year, CAST(EXTRACT(MONTH FROM sample_first_date) AS INTEGER) AS month, COUNT(*) AS count
```
dancoates, line 715: would be good to add some newlines here for readability

db/python/tables/sample.py
```diff
@@ -687,3 +693,55 @@ async def get_samples_with_missing_participants_by_internal_id(
...
+
+        rows = await self.connection.fetch_all(_query, values)
+        result: dict[int, dict[date, int]] = {}
+        accumulated_sample_count: dict[int, int] = {}
+
+        # results are sorted by project, year, month
+        # accumulate projects previous months into the processed month
+        for r in rows:
+            proj = r['project']
+            if proj not in result:
```
dancoates, line 729: tiny thing, but you could use a defaultdict for `result` instead of needing to do this check

db/python/tables/sample.py
```diff
@@ -687,3 +693,55 @@ async def get_samples_with_missing_participants_by_internal_id(
...
+                accumulated_sample_count.get(proj, 0) + r['count']
+            )
+            result[proj][date.fromisoformat(f"{r['year']}-{r['month']:02d}-01")] = (
+                accumulated_sample_count[proj]
+            )
+
+        # append all the months up to the current month
+        for proj, months in result.items():
+            current_month = date.today().replace(day=1)
+            while current_month not in months:
```
dancoates, line 743: I think this could go back in time infinitely if the months list is empty to start with?

dancoates (commented):
This looks good, just a few very minor comments and one thing that might be a potential bug if I'm reading it right.

db/python/tables/sample.py
milo-hyben: Good point, had addressed it.

dancoates (approved):

PR #1120, opened 2025-09-23
SET-456 import families endpoint throws 500 error on empty sheet

hrrsheen:


api/routes/family.py
```diff
@@ -190,11 +190,21 @@ async def import_families(
 
     family_layer = FamilyLayer(connection)
     reader = csv.reader(codecs.iterdecode(file.file, 'utf-8-sig'), delimiter=delimiter)
+    rows = [r for r in reader if not r[0].startswith('#')]
+
+    if len(rows) == 0:
+        if has_header:
+            return {
```
dancoates, line 197: Rather than returning tuples here, I think it would be more correct to raise a HTTPException. There's some code here: https://github.com/populationgenomics/metamist/blob/dev/api/server.py#L149 that will take that exception and format the response in a consistent way.

dancoates (changes requested):

api/routes/family.py
```diff
@@ -190,11 +190,21 @@ async def import_families(
...
+                'success': False,
+                'message': 'A header was expected but file is empty.',
+            }, 400
+        return {}, 204
```
dancoates, line 201: Also, not sure if 204 is ok to use here, MDN docs say:

> Note that the response must not include any content or the [Content-Length](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Length) header (browsers may reject responses that include content).

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204

So sounds like this could cause errors if this was used in a browser?

dancoates: I think a 200 would probably be fine

api/routes/family.py
hrrsheen: Made the change.

dancoates (approved):
Looks good, the tests seem to cover the different cases well

api/routes/family.py
```diff
@@ -190,12 +190,26 @@ async def import_families(
...
-    rows = [r for r in reader if not r[0].startswith('#')]
-    if len(rows[0]) == 1:
+    # Header with no data
+    if headers and not rows:
+        return {
+            'success': True,
+            'warnings': ['Submitted file contained a header with no data'],
+        }
+
+    if len(headers) == 1:
```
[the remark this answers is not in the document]
hrrsheen: Very true! Fixed.

test/test_family.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @run_as_sync
+    async def test_import_families_header_no_content(self):
+        """
+        Test importing families from a file with a header but no data.
+        """
+        with TemporaryFile(mode='wb+', prefix='test', suffix='.tsv') as f:
+            f.write(b'Some\ttest\theader\twithout\tdata\n')
+            f.seek(0)
+            emptyTestFile = UploadFile(f)
```
jmarshall, line 56: Boring comment I know, but this one is not an `emptyTestFile` so variable could be better named.

hrrsheen: Fair point. I've updated the variable name.

PR #1121, opened 2025-09-23
SET-468 update analysis endpoint unnecessarily requires status parameter to be passed

hrrsheen:


dancoates (dismissed):
Looks good, nice simple change

PR #1122, opened 2025-09-24
SET-597 creating an empty cohort should raise an error

hrrsheen:
At the moment when you create a cohort, it is possible for the criteria or template that you provide to result in a empty cohort with no sequencing group ids. We should raise an error in these cases rather than allowing an empty cohort, as empty cohorts are not useful.


test/test_cohort.py
```diff
@@ -72,74 +72,19 @@ async def test_create_cohort_missing_args(self):
...
-        )
-
-        _ = await self.cohortl.create_cohort_from_criteria(
-            project_to_write=self.project_id,
-            description='Cohort from template',
-            cohort_name='Cohort from empty template',
-            dry_run=False,
-            template_id=tid,
+        self.assertIn(
+            'Cohort creation criteria resulted in no sequencing groups being selected. Please check the criteria and try again',
```
jmarshall, line 86: Suggest checking for a subset of the message (e.g. `'criteria resulted in no sequencing groups'`), so if the message is later tweaked with luck it'll only need tweaking in one place.

jmarshall (approved):
One fairly trivial comment, otherwise LGTM.

test/test_cohort.py
hrrsheen: Good suggestion, I've made the change.

PR #1123, opened 2025-09-25
Replace some usages of deprecated functions

jmarshall:
We require `strawberry-graphql[fastapi]==0.268.1`, so it's safe to use [`graphql_ide`, introduced in 0.213.0](https://strawberry.rocks/docs/breaking-changes/0.213.0).

Follow PR #888 in using pydantic's `model_dump()`.

### DIFF
--- a/api/graphql/schema.py
+++ b/api/graphql/schema.py
@@ -1617,5 +1617,5 @@ schema = strawberry.Schema(
     query=Query, mutation=Mutation, extensions=[QueryDepthLimiter(max_depth=10)]
 )
 MetamistGraphQLRouter: GraphQLRouter = GraphQLRouter(
-    schema, graphiql=True, context_getter=get_context
+    schema, graphql_ide='graphiql', context_getter=get_context
 )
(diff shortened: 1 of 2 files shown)


jmarshall:
There is currently one remaining warning from the test suite:

```
/…/lib/python3.11/site-packages/google/cloud/storage/blob.py:29:
DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
    import cgi
```

We currently pin `google-cloud-storage==1.43.0`. Getting rid of this warning would require updating that to [2.8.0](https://github.com/googleapis/python-storage/releases/tag/v2.8.0). (Revisiting that pinning is probably more feasible than it appears: 1.43.0 was the second-to-last 1.x release and pre-dates 2.8.0 by less than six months; the [2.0 major bump](https://github.com/googleapis/python-storage/releases/tag/v2.0.0) appears to be due to Python 2 removal rather than API changes.)

ckrishk (approved):

PR #1158, opened 2025-11-27
Consolidate deploy workflow into test

nevoodoo:
## Description

Separates test and deploy into independent workflow files with different behavior per branch:

dev: Deploy runs concurrently with tests (faster feedback loop)
main: Deploy is gated on tests passing via workflow_call (safe prod deploys)

## Changes

`test.yaml`: Now excludes main from its push trigger (`branches-ignore: [main]`), since tests on main run via `deploy-prod.yaml`'s `workflow_call`. 

`deploy-dev.yaml` (new): Triggers on push to dev, deploys immediately without waiting for tests.

`deploy-prod.yaml` (new): Triggers on push to main, calls `test.yaml` via `workflow_call` and only deploys if tests pass. 

`deploy.yaml` (deleted): Replaced by the two branch-specific workflows above.

I was originally trying to use `workflow_run` but that always reads the workflow file from the repo's default branch (dev), meaning changes to the prod deploy workflow would take effect via dev rather than main. Using `workflow_call` avoids this and the workflow file is always read from the branch being pushed to.

### DIFF
--- a/.github/workflows/deploy-dev.yaml
+++ b/.github/workflows/deploy-dev.yaml
@@ -1,0 +1,124 @@
+name: Deploy Dev
+on:
+  push:
+    branches:
+      - dev
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  deploy:
+    name: Deploy Dev
+    runs-on: ubuntu-latest
+    environment: production
+    permissions:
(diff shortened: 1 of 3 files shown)


hrrsheen (approved):
LGTM

PR #1193, opened 2026-02-05
Task/[SET-901] pg migrate family and family participants

rameshka:
This PR migrates `family` and `family_participant` entities to **PostgreSQL**. It also updates related test classes that the `Family` layer is tested.


db/python/tables/family.py
```diff
@@ -143,20 +148,14 @@ async def search(
...
-                'project_ids': project_ids,
-                'search_pattern': escape_like_term(query) + '%',
-                'limit': limit,
-            },
-        )
+        search_pattern = (escape_like_term(query) + '%',)
+        rows = await (
+            await self.connection.pg_connection.execute(
+                t'SELECT project, family_id, external_id FROM '
+                t'family_external_id WHERE project = ANY({project_ids}) AND external_id ILIKE {search_pattern} LIMIT {limit}'
```
rameshka: Used `ILIKE` here, as PostgreSQL `LIKE` is case-sensitive. Also, previous MariaDB `LIKE`  depends on the collation (mariadb:11.7.2 default collation `utf8mb4_uca1400_ai_ci`).

db/python/tables/family.py
```diff
@@ -190,72 +189,73 @@ async def update_family(
...
+                    )
+
+                if to_update:
+                    await curr.execute(t'SELECT project FROM family WHERE id = {id_}')
+                    project = await curr.fetchone()
+
+                    _update_query = """
+                            INSERT INTO family_external_id (project, family_id, name, external_id, audit_log_id)
+                            VALUES (%(project)s, %(id)s, %(name)s, %(external_id)s, %(audit_log_id)s)
+                            ON CONFLICT (family_id, name)
```
rameshka: `ON CONFLICT` defined on the primary key

db/python/layers/participant.py
```diff
@@ -623,7 +623,9 @@ async def upsert_participant(
     ) -> ParticipantUpsertInternal:
         """Create a single participant"""
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
+            self.connection.pg_connection.transaction
+            if open_transaction
```
dancoates: It'd be good to change this `open_transaction` functionality, having it be optional via an arg is a bit of an accident waiting to happen. With psycopg we can check if there is a connection open and only open on if it isn't already open. 

see here for an example: https://github.com/populationgenomics/metamist/pull/1207/changes#diff-a1706cd75c100ac4a93e1ee7e844a6db374f1159f23492ae4ad33718729e68dfR297

db/python/tables/family.py
```diff
@@ -54,17 +61,16 @@ async def query(
...
     ) -> tuple[set[ProjectId], list[FamilyInternal]]:
         """Get all families for some project"""
-        _query = """
-            SELECT f.id, JSON_OBJECTAGG(feid.name, feid.external_id) AS external_ids,
-                   f.description, f.coded_phenotype, f.meta, f.project
-            FROM family f
-            INNER JOIN family_external_id feid ON f.id = feid.family_id
-        """
+
+        _query = [
```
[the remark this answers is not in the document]
dancoates: I see the list is added to below, but could still start with a list that is a single multi line template string rather than three strings

db/python/tables/family.py
```diff
@@ -190,72 +189,73 @@ async def update_family(
...
-                    """,
-                    {'id': id_, 'names': to_delete, 'audit_log_id': audit_log_id},
-                )
-
-            if to_update:
-                project = await self.connection.fetch_val(
-                    'SELECT project FROM family WHERE id = :id',
-                    {'id': id_},
-                )
+            meta_param = to_db_json(meta)
```
dancoates: Use psycopg's `Jsonb` adapter instead of to_db_json:
https://www.psycopg.org/psycopg3/docs/basic/adapt.html#json-adaptation

db/python/tables/family.py
```diff
@@ -190,72 +189,73 @@ async def update_family(
...
+                if external_ids is None:
+                    external_ids = {}
+
+                to_delete = [k.lower() for k, v in external_ids.items() if v is None]
+                to_update = {
+                    k.lower(): v for k, v in external_ids.items() if v is not None
+                }
+
+                if to_delete:
+                    # psycopg does no support executing multiple parameterized statements in a single execution
```
dancoates: No need for this comment, to the reader it implies that this isn't desirable but it's actually better practice to run separate statements separately.

db/python/tables/family.py
```diff
@@ -272,38 +272,33 @@ async def create_family(
...
-                    'description': description,
-                    'coded_phenotype': coded_phenotype,
-                    'meta': to_db_json(meta or {}),
-                    'audit_log_id': audit_log_id,
-                },
-            )
+        project_param = project or self.project_id
+        conn = self.connection.pg_connection
+        async with conn.transaction():
+            async with conn.cursor(row_factory=scalar_row) as curr:
```
dancoates: combine into a single async with

db/python/tables/family.py
```diff
@@ -324,63 +319,46 @@ async def insert_or_update_multiple_families(
...
-                        'id': existing_id,
-                        'description': descr,
-                        'coded_phenotype': cph,
-                        'meta': to_db_json(mt or {}),
-                        'audit_log_id': audit_log_id,
-                    },
+        # each query executes independently
+        project_param = project or self.project_id
+
+        async with self.connection.pg_connection.cursor(row_factory=scalar_row) as curr:
```
dancoates: This is a problem in the old code, but there's multiple inserts here so should be happening in a transaction

db/python/tables/family_participant.py
```diff
@@ -93,19 +93,24 @@ async def create_rows(
...
-            placeholder_keys = ', '.join(f':{k}' for k in d_keys)
+            placeholder_keys = ', '.join(f'%({k})s' for k in d_keys)
             update_keys = ', '.join(
-                f'{k}=:{k}' for k in d_keys if k not in ignore_keys_during_update
+                f'{k}=EXCLUDED.{k}'
+                for k in d_keys
+                if k not in ignore_keys_during_update
             )
+
             _query = f"""
```
dancoates: The query building here and above might need a bit more of a refactor, you can't (or at least shouldn't) use f-strings to template a query string to pass to psycopg, at the moment the type checking is failing because it only allows literal strings or psycopg's composed or template strings for the query param.

dancoates (commented):
This is looking great, there's mostly just a few style/formatting things and a couple of improvements that we can make to existing code.

The main style thing is that it's better to use multi line template strings rather than having multiple strings that later get joined together, it makes it much easier to edit queries in the future. I've commented on this in a few places but it'd be good to update all cases. 

The other thing I noticed was the use of `to_db_json` - we don't need to use this any more as psycopg has built in adaption for json, again I only commented on a few instances of this but would be good to update them all.

Could you also please remove the `from_db` method from the family model, it isn't used any more.

db/python/tables/family_participant.py
rameshka: Sure @dancoates. 
Having looked at the logic. It does not look like any dynamic query building is necessary here, as claimed by the comment (on the code). The list of columns that get updated remains constant and does not get filtered out based on the input.

db/python/layers/participant.py
```diff
@@ -679,13 +679,15 @@ async def upsert_participants(
     ):
         """Batch upsert a list of participants with sequences"""
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
```
rameshka, line 682: left the transaction handling in the 'participant' layer in this state, as @violetbrina is already working on this entity.

dancoates (approved):
Nice @rameshka this is looking good

rameshka:
> Nice @rameshka this is looking good

Thanks for reviewing @dancoates.

PR #1202, opened 2026-02-10
Migrate participant and participant phenotype queries

violetbrina:
Mirgrate the participant and participant phenotype queries to use postgres connection

In the process of this migration an issue was uncoverd in the test `test_graphql_upsert_participant_with_phenotypes`. The project passed through graphql as a string was not being used in the upsert. The appropriate data access checks were made, but then the value was not being utilised in the upsert_participants method itself.

This PR includes a fix where in `upsert_participants` in `api/graphql/mutations/participant.py` the project id is fetched when the project access is checked and that value is passed down all the way through to the table where `create_participant` is called.

All the tests should be passing excluding `test_query_by_sample` since the Sample tables have yet to be migrated.

### DIFF
--- a/api/graphql/mutations/participant.py
+++ b/api/graphql/mutations/participant.py
@@ -108,16 +108,21 @@ class ParticipantMutations:
         from api.graphql.schema import GraphQLParticipant  # noqa: PLC0415
 
         async with info.context['get_connection']() as connection:
-            connection.check_access_to_projects_for_names(
+            projects = connection.get_and_check_access_to_projects_for_names(
                 [project], FullWriteAccessRoles
             )
+            projects = list(projects)
+
+            assert len(projects) >= 1, 'Project not found or user does not have access'
+            project_id = projects[0].id
 
             pt = ParticipantLayer(connection)
             results = await pt.upsert_participants(
                 [
                     ParticipantUpsert.from_dict(strawberry.asdict(p)).to_internal()
                     for p in participants
-                ]
+                ],
+                project=project_id,
             )
             updated_participants = await pt.get_participants_by_ids(
                 [p.id for p in results]  # type: ignore [arg-type]
--- a/api/graphql/mutations/participant.py
+++ b/api/graphql/mutations/participant.py
@@ -113,7 +113,8 @@ class ParticipantMutations:
             )
             projects = list(projects)
 
-            assert len(projects) >= 1, 'Project not found or user does not have access'
+            if len(projects) < 1:
+                raise ValueError('Project not found or user does not have access')
             project_id = projects[0].id
 
             pt = ParticipantLayer(connection)
--- a/db/python/layers/participant.py
+++ b/db/python/layers/participant.py
@@ -629,7 +629,7 @@ class ParticipantLayer(BaseLayer):
         in_transaction = conn.info.transaction_status == TransactionStatus.INTRANS
         with_function = conn.transaction if not in_transaction else NoOpAenter
 
-        async with with_function():
+        async with self.connection.transaction():
             if participant.id:
                 project_ids = await self.pttable.get_project_ids_for_participant_ids(
                     [participant.id]
@@ -688,7 +688,7 @@ class ParticipantLayer(BaseLayer):
         async with with_function():
             # Create or update participants
             for p in participants:
-                await self.upsert_participant(p, project, open_transaction=False)
+                await self.upsert_participant(p)
 
         # Format and return response
         return participants
--- a/db/python/layers/participant.py
+++ b/db/python/layers/participant.py
@@ -623,7 +623,6 @@ class ParticipantLayer(BaseLayer):
         self,
         participant: ParticipantUpsertInternal,
         project: ProjectId | None = None,
-        open_transaction=True,
     ) -> ParticipantUpsertInternal:
         """Create a single participant"""
         conn = self.connection.pg_connection
@@ -680,7 +679,6 @@ class ParticipantLayer(BaseLayer):
         self,
         participants: list[ParticipantUpsertInternal],
         project: ProjectId | None = None,
-        open_transaction=True,
     ):
         """Batch upsert a list of participants with sequences"""
         conn = self.connection.pg_connection
--- a/db/python/layers/participant.py
+++ b/db/python/layers/participant.py
@@ -3,6 +3,8 @@ from collections import defaultdict
 from enum import Enum
 from typing import Any
 
+from psycopg.pq import TransactionStatus
+
 from db.python.connect import Connection
 from db.python.filters import GenericFilter
 from db.python.layers.base import BaseLayer
@@ -624,11 +626,9 @@ class ParticipantLayer(BaseLayer):
         open_transaction=True,
     ) -> ParticipantUpsertInternal:
         """Create a single participant"""
-        with_function = (
-            self.connection.pg_connection.transaction
-            if open_transaction
-            else NoOpAenter
-        )
+        conn = self.connection.pg_connection
+        in_transaction = conn.info.transaction_status == TransactionStatus.INTRANS
+        with_function = conn.transaction if not in_transaction else NoOpAenter
 
         async with with_function():
             if participant.id:
@@ -683,11 +683,9 @@ class ParticipantLayer(BaseLayer):
         open_transaction=True,
     ):
         """Batch upsert a list of participants with sequences"""
-        with_function = (
-            self.connection.pg_connection.transaction
-            if open_transaction
-            else NoOpAenter
-        )
+        conn = self.connection.pg_connection
+        in_transaction = conn.info.transaction_status == TransactionStatus.INTRANS
+        with_function = conn.transaction if not in_transaction else NoOpAenter
 
         async with with_function():
             # Create or update participants
--- a/db/python/layers/participant.py
+++ b/db/python/layers/participant.py
@@ -681,11 +681,7 @@ class ParticipantLayer(BaseLayer):
         project: ProjectId | None = None,
     ):
         """Batch upsert a list of participants with sequences"""
-        conn = self.connection.pg_connection
-        in_transaction = conn.info.transaction_status == TransactionStatus.INTRANS
-        with_function = conn.transaction if not in_transaction else NoOpAenter
-
-        async with with_function():
+        async with self.connection.transaction():
             # Create or update participants
             for p in participants:
                 await self.upsert_participant(p, project=project)
--- a/db/python/layers/participant.py
+++ b/db/python/layers/participant.py
@@ -630,10 +630,6 @@ class ParticipantLayer(BaseLayer):
             else NoOpAenter
         )
 
-        project = project or self.connection.project_id
-        if not project:
-            raise ValueError('Project must be specified for upserting participant')
-
         async with with_function():
             if participant.id:
                 project_ids = await self.pttable.get_project_ids_for_participant_ids(
@@ -683,6 +679,7 @@ class ParticipantLayer(BaseLayer):
     async def upsert_participants(
         self,
         participants: list[ParticipantUpsertInternal],
+        project: ProjectId | None = None,
         open_transaction=True,
     ):
         """Batch upsert a list of participants with sequences"""
@@ -695,7 +692,7 @@ class ParticipantLayer(BaseLayer):
         async with with_function():
             # Create or update participants
             for p in participants:
-                await self.upsert_participant(p, open_transaction=False)
+                await self.upsert_participant(p, project, open_transaction=False)
 
         # Format and return response
         return participants
--- a/db/python/layers/participant.py
+++ b/db/python/layers/participant.py
@@ -357,7 +357,8 @@ class ParticipantLayer(BaseLayer):
             # if there are no participants to add, skip the next step
             return '0 participants updated'
 
-        async with self.connection.connection.transaction():
+        conn = self.connection.pg_connection
+        async with conn.transaction():
             sample_ids_to_update = {
                 external_sample_map_with_no_pid[external_id]: pid
                 for external_id, pid in unlinked_participants.items()
@@ -407,7 +408,8 @@ class ParticipantLayer(BaseLayer):
         # currently only does the seqr metadata template
 
         # filter to non-comment rows
-        async with self.connection.connection.transaction():
+        conn = self.connection.pg_connection
+        async with conn.transaction():
             ppttable = ParticipantPhenotypeTable(self.connection)
 
             self._validate_individual_metadata_headers(headers)
@@ -596,7 +598,7 @@ class ParticipantLayer(BaseLayer):
         return id_map
 
     async def get_external_participant_id_to_internal_sequencing_group_id_map(
-        self, project: int, sequencing_type: str = None
+        self, project: int, sequencing_type: str | None = None
     ) -> list[tuple[str, int]]:
         """
         Get a map of {external_participant_id} -> {internal_sequencing_group_id}
@@ -618,13 +620,19 @@ class ParticipantLayer(BaseLayer):
     async def upsert_participant(
         self,
         participant: ParticipantUpsertInternal,
-        project: ProjectId = None,
+        project: ProjectId | None = None,
         open_transaction=True,
     ) -> ParticipantUpsertInternal:
         """Create a single participant"""
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
+            self.connection.pg_connection.transaction
+            if open_transaction
+            else NoOpAenter
         )
+
+        project = project or self.connection.project_id
+        if not project:
+            raise ValueError('Project must be specified for upserting participant')
 
         async with with_function():
             if participant.id:
@@ -679,7 +687,9 @@ class ParticipantLayer(BaseLayer):
     ):
         """Batch upsert a list of participants with sequences"""
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
+            self.connection.pg_connection.transaction
+            if open_transaction
+            else NoOpAenter
         )
 
         async with with_function():
@@ -995,7 +1005,7 @@ class ParticipantLayer(BaseLayer):
         fp_row = await self.get_family_participant_data(
             family_id=old_family_id, participant_id=participant_id
         )
-        async with self.connection.connection.transaction():
+        async with self.connection.pg_connection.transaction():
             await self.remove_participant_from_family(
                 family_id=old_family_id, participant_id=participant_id
             )
--- a/db/python/layers/participant.py
+++ b/db/python/layers/participant.py
@@ -616,17 +616,11 @@ class ParticipantLayer(BaseLayer):
     # region UPSERTS / UPDATES
 
     async def upsert_participant(
-        self,
-        participant: ParticipantUpsertInternal,
-        project: ProjectId = None,
-        open_transaction=True,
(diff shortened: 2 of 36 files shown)


db/python/tables/participant_phenotype.py
```diff
@@ -55,19 +67,21 @@ async def get_key_value_rows_for_participant_ids(
...
 
         _query = """
-SELECT participant_id, description, value
-FROM participant_phenotypes
-WHERE participant_id in :participant_ids AND value IS NOT NULL
-"""
-        rows = await self.connection.fetch_all(
-            _query, {'participant_ids': participant_ids}
-        )
+            SELECT participant_id, description, value
```
rameshka: minor_suggestion: It looks like we can use the template string format here

db/python/tables/participant.py
```diff
@@ -1,220 +1,218 @@
...
-            'p.project',
-            'p.audit_log_id',
-        ]
-    )
+    def __init__(self, connection: Connection):
+        self.connection = connection
+
+    keys = [
+        'p.id',
+        sql.SQL('JSON_OBJECTAGG(peid.name, peid.external_id) AS external_ids'),
```
rameshka: Is postgreSQL `jsonb_object_agg ` here?

db/python/tables/participant.py
```diff
@@ -388,34 +395,33 @@ async def update_participants(
...
+        )
+
+        # Filter out empty updates and join the rest with commas
+        updates = [u for u in updates if u]
+        updaters_str = sql.SQL(', ').join(updaters)
 
-        updaters_str = ', '.join(updaters)
-        _query = f'UPDATE participant SET {updaters_str} WHERE id = :pid'
-        await self.connection.execute_many(_query, list_values)
+        _query = t'UPDATE participant SET {updaters_str} WHERE id = %s'
```
rameshka: TODO

db/python/tables/participant.py
```diff
@@ -442,69 +448,66 @@ async def update_participant(
...
             if to_delete:
                 # Set audit_log_id to this transaction before deleting the rows
-                _audit_update_query = """
-                UPDATE participant_external_id
-                SET audit_log_id = :audit_log_id
-                WHERE participant_id = :pid AND name IN :names
+                _audit_update_query = t"""
+                    UPDATE participant_external_id
+                    SET audit_log_id = {audit_log_id:l}
+                    WHERE participant_id = {participant_id:l} AND name IN {to_delete:l}
```
rameshka: I tested out a few queries with 'IN' with a list of `str` or `int`, and it gave me `psycopg.errors.SyntaxError`. name=ANY({to_delete:l}) works. Also, [from the documentation](https://www.psycopg.org/psycopg3/docs/basic/adapt.html#lists-adaptation).

db/python/tables/participant.py
```diff
@@ -442,69 +448,66 @@ async def update_participant(
...
-                    },
-                )
+                await conn.execute(_audit_update_query)
 
-                _delete_query = """
-                DELETE FROM participant_external_id
-                WHERE participant_id = :pid AND name IN :names
+                _delete_query = t"""
+                    DELETE FROM participant_external_id
+                    WHERE participant_id = {participant_id:l} AND name IN {to_delete:l}
```
rameshka: I think we can use the default parameter format with `s` (or not specifying will default to `s`). From my understanding, `l` is used in special cases if the SQL query does not support standard server-side parameter binding

db/python/tables/participant.py
```diff
@@ -1,220 +1,218 @@
...
-        SELECT project
-        FROM participant
-        WHERE id in :participant_ids
-        GROUP BY project"""
-        rows = await self.connection.fetch_all(
-            _query, {'participant_ids': participant_ids}
-        )
+            SELECT project
+            FROM participant
+            WHERE id = ANY(%(participant_ids)s)
```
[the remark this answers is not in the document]
violetbrina: Good point. Changed!

db/python/tables/participant.py
```diff
@@ -388,34 +395,33 @@ async def update_participants(
...
+        updaters_str = sql.SQL(', ').join(updaters)
 
-        updaters_str = ', '.join(updaters)
-        _query = f'UPDATE participant SET {updaters_str} WHERE id = :pid'
-        await self.connection.execute_many(_query, list_values)
+        _query = t'UPDATE participant SET {updaters_str} WHERE id = %s'
+
+        conn = self.connection.pg_connection
+        cur = conn.cursor()
+        await cur.executemany(_query, [(pid,) for pid in participant_ids])
```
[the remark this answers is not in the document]
violetbrina: It's a requirement for executemany (otherwise every query is identical). But I've cleaned up the format as per the new guide in slack.

db/python/tables/participant.py
violetbrina: Ah gotcha. I thought that :l :q and :i were an exclusive set of options and was just trying to be explicit. But happy enough to remove them.

api/graphql/mutations/participant.py
```diff
@@ -108,16 +108,21 @@ async def upsert_participants(
         from api.graphql.schema import GraphQLParticipant  # noqa: PLC0415
 
         async with info.context['get_connection']() as connection:
-            connection.check_access_to_projects_for_names(
+            projects = connection.get_and_check_access_to_projects_for_names(
                 [project], FullWriteAccessRoles
             )
+            projects = list(projects)
+
+            assert len(projects) >= 1, 'Project not found or user does not have access'
```
dancoates: Can we raise a NotFoundError here rather than using an assert so that the error code is returned correctly

db/python/layers/participant.py
```diff
@@ -618,12 +620,14 @@ async def export_participant_table(self, project: int):
...
         participant: ParticipantUpsertInternal,
-        project: ProjectId = None,
+        project: ProjectId | None = None,
         open_transaction=True,
     ) -> ParticipantUpsertInternal:
         """Create a single participant"""
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
+            self.connection.pg_connection.transaction
+            if open_transaction
```
dancoates: Rather than continuing to pass around this `open_transaction` param which is easy to forget to do, psycopg has the ability to check whether you are in a transaction and only open a tranaction if one isn't already open. 

example here:

https://github.com/populationgenomics/metamist/pull/1207/changes#diff-a1706cd75c100ac4a93e1ee7e844a6db374f1159f23492ae4ad33718729e68dfR297

db/python/tables/participant.py
```diff
@@ -1,220 +1,218 @@
...
-            'p.project',
-            'p.audit_log_id',
-        ]
-    )
+    def __init__(self, connection: Connection):
+        self.connection = connection
+
+    keys = [
+        'p.id',
+        sql.SQL('json_object_agg(peid.name, peid.external_id) AS external_ids'),
```
dancoates: It doesn't make much technical difference, but can we use jsonb_object_agg here as we are using jsonb everywhere else.

db/python/tables/participant.py
```diff
@@ -1,220 +1,218 @@
...
-            query_lines.append('ORDER BY pp.id')
+                query_template += t"""
+                    INNER JOIN family_external_id feid ON feid.family_id = f.id
+                """
 
-        if limit:
-            query_lines.append('LIMIT :limit')
-            values['limit'] = limit
+        # WHERE, ORDER BY, LIMIT, OFFSET
+        query_template += t' WHERE ' + sql.SQL(' AND ').join(wheres) if wheres else t''
```
dancoates: This will fail if wheres is falsey as it will create an empty where statement. You can avoid this by doing `if wheres else t'TRUE'` so that it will resolve to `WHERE TRUE`

db/python/tables/participant.py
```diff
@@ -1,220 +1,218 @@
...
-                needs_family_eid = True
-                feid_wheres, feid_values = filter_.family.to_sql(
-                    {'external_id': 'feid.external_id'}, only=['external_id']
+        query_template += (
+            t' INNER JOIN participant_external_id peid ON pp.id = peid.participant_id'
+        )
+
+        # Check filter for the sample table and sample_external_id table
+        if filter_.sample or filter_.sequencing_group or filter_.assay:
+            if not filter_.sample:
```
dancoates: I'm not sure this is correct - having a sg or assay filter require joining with the sample table, but shouldn't require a sample filter if I'm reading the previous code correctly?

db/python/tables/participant.py
```diff
@@ -1,220 +1,218 @@
...
+                ON p.id = {participant_eid_table_alias:i}.participant_id
+            """
+            if participant_eid_table_alias
+            else t''
+        )
 
-        if skip:
-            query_lines.append('OFFSET :offset')
-            values['offset'] = skip
+        def format_key(key: str | sql.SQL) -> sql.Composable:
```
dancoates: Type for key should be `str | Composable`

db/python/tables/participant.py
```diff
@@ -323,55 +332,53 @@ async def create_participant(
...
+        conn = self.connection.pg_connection
 
-        _query = """
-INSERT INTO participant
-    (reported_sex, reported_gender, karyotype, meta, audit_log_id, project)
-VALUES
-    (:reported_sex, :reported_gender, :karyotype, :meta, :audit_log_id, :project)
-RETURNING id
-        """
+        meta_value = to_db_json(meta or {})
```
dancoates: Use psycopg's json adaption here: https://www.psycopg.org/psycopg3/docs/basic/adapt.html#json-adaptation

db/python/tables/participant.py
```diff
@@ -323,55 +332,53 @@ async def create_participant(
...
+
+        cur = await conn.execute(_query)
+        row = await cur.fetchone()
+        if not row:
+            raise ValueError('Failed to create participant')
+        new_id = row['id']
+
+        cur = conn.cursor()
+        async with cur.copy(
+            'COPY participant_external_id (project, participant_id, name, external_id, audit_log_id) FROM STDIN'
```
dancoates: Not sure about using COPY here, feels like a premature optimisation with a less-nice syntax for something that will be writing at most a few rows?

db/python/tables/participant.py
```diff
@@ -388,34 +395,33 @@ async def update_participants(
...
-            values['karyotype'] = karyotypes
-        if metas:
-            updaters.append('meta = JSON_MERGE_PATCH(COALESCE(meta, "{}"), :meta')
-            values['meta'] = metas
-
-        keys = list(values.keys())
-        list_values = [
-            {k: item[idx] for k, item in values.items()}
-            for idx in range(len(values[keys[0]]))
-        ]
```
dancoates: This update_participants method isn't used, best to just delete it.

db/python/tables/participant.py
```diff
@@ -429,9 +435,9 @@ async def update_participant(
         """
         Update participant
         """
-        audit_log_id = await self.audit_log_id()
-        updaters = ['audit_log_id = :audit_log_id']
-        fields = {'pid': participant_id, 'audit_log_id': audit_log_id}
+        conn = self.connection.pg_connection
```
dancoates: There's multiple mutations in this method, they should be happening in a transaction

db/python/tables/participant.py
```diff
@@ -442,69 +448,64 @@ async def update_participant(
...
+                    )
+
+        updates: list[Template | None] = [t'audit_log_id = {audit_log_id}']
+
+        updates.append(t'reported_sex = {reported_sex}' if reported_sex else None)
+        updates.append(
+            t'reported_gender = {reported_gender}' if reported_gender else None
+        )
+        updates.append(t'karyotype = {karyotype}' if karyotype else None)
+        meta_literal = to_db_json(meta)
```
dancoates: Use psycopg's Json adaption https://www.psycopg.org/psycopg3/docs/basic/adapt.html#json-adaptation

db/python/tables/participant.py
```diff
@@ -577,15 +572,18 @@ async def get_participants_by_families(
         self, family_ids: list[int]
     ) -> tuple[set[ProjectId], dict[int, list[ParticipantInternal]]]:
         """Get list of participants keyed by families, duplicates results"""
-        _query = f"""
-            SELECT fp.family_id, {self.keys_str}
+        _query = t"""
+            SELECT fp.family_id, {self.keys_str:i}
```
dancoates: I don't think this will work, `:i` expects a single column, not a list

db/python/tables/participant.py
```diff
@@ -627,13 +627,16 @@ async def get_external_ids_by_participant(
         if not participant_ids:
             return {}
 
-        _query = """
-        SELECT participant_id AS id, JSON_ARRAYAGG(external_id) AS external_ids_list
-        FROM participant_external_id
-        WHERE participant_id IN :pids
-        GROUP BY participant_id
+        _query = t"""
+            SELECT participant_id AS id, JSON_ARRAYAGG(external_id) AS external_ids_list
```
dancoates: you should be able to use straight `array_agg` here and it will be returned as a list so you don't need to parse the json

db/python/tables/participant.py
```diff
@@ -646,41 +649,43 @@ async def get_external_participant_id_to_internal_sequencing_group_id_map(
...
-        SELECT project, participant_id AS id, external_id
-        FROM participant_external_id
-        WHERE project in :project_ids AND external_id LIKE :search_pattern
-        LIMIT :limit
+
+        search_literal = escape_like_term(query) + '%'
+        _query = t"""
+            SELECT project, participant_id AS id, external_id
+            FROM participant_external_id
+            WHERE project in {project_ids} AND external_id LIKE {search_literal}
```
dancoates: use ilike, the collation in the mariadb database meant that like comparisons were case insensitive, but for postgres they will be case sensitive unless we use `ILIKE`

db/setup.md
```diff
@@ -0,0 +1,15 @@
...
+> docker compose -f docker-compose.yaml build
+
+## Run
+
+Run detached (-d)
+> docker compose run -d postgres
+
+## Build and Run
+
+> docker compose up -d
```
dancoates: Tiny thing, but seeing as most of the time this is all you need to run, should this be at the top? 

Building and running separately is a bit more of an edge case.

test/test_participant.py
```diff
@@ -83,106 +93,125 @@ async def test_query_by_exids(self):
...
+        assert resp['data']['project'] is not None
 
-        self.assertEqual(1, len(resp['project']['participants']))
+        assert len(resp['data']['project']['participants']) == 1
 
-    @run_as_sync
-    async def test_query_by_sample(self):
+    @pytest.mark.asyncio
+    @pytest.mark.project_roles(['reader', 'writer'])
+    async def test_query_by_sample(self, connection_with_project: Connection):
```
dancoates, line 148: Mark this one as skipped while it is failing so that the test suite passes.

dancoates (commented):
Looking good! Just a bunch of small comments

db/python/tables/participant.py
dancoates: Also self.keys_str is not defined on the class

api/graphql/mutations/participant.py
violetbrina: Sure thing!

db/python/layers/participant.py
violetbrina: Ah excellent, noted! I'll switch to this new convention now.

db/python/tables/participant.py
violetbrina: Sure thing!

db/python/tables/participant.py
violetbrina: Ah excellent point. I think I got a bit overzealous in the refactor. You're totally right, the sample query needs a join, but the filter isn't required. I'll fix it now.

db/python/tables/participant.py
violetbrina: Yeah that's what I initially had but then got a bunch of typing errors. I'll switch it back but I do wish the type checker knew that a sql.SQL type was also a type of sql.Composable.

db/python/tables/participant.py
violetbrina: Ah good to know. It's definitely been a minute since using postgres, I'm so used to living in MariaDB land.

db/python/tables/participant.py
violetbrina: Definitely missed that one. I think I switched the self.keys to a string and back to an array again but missed this one on changing it back. Fixed now!

db/python/tables/participant.py
violetbrina: Excellent point. Thanks!

db/python/tables/participant.py
violetbrina: Sure, I'm happy to re-write it as an insert. Don't mind either way.

db/python/layers/participant.py
```diff
@@ -669,19 +678,17 @@ async def upsert_participant(
...
     ):
         """Batch upsert a list of participants with sequences"""
-        with_function = (
-            self.connection.pg_connection.transaction
-            if open_transaction
-            else NoOpAenter
-        )
+        conn = self.connection.pg_connection
+        in_transaction = conn.info.transaction_status == TransactionStatus.INTRANS
+        with_function = conn.transaction if not in_transaction else NoOpAenter
```
dancoates: This can be moved to using the transaction helper on the connection

db/python/tables/participant.py
```diff
@@ -280,6 +288,7 @@ async def get_participants(
         )
         return particicpants
 
+    # TODO: cannot convert until MetaTable is converted
```
dancoates: Can remove this todo now :)

db/python/tables/participant.py
```diff
@@ -317,99 +326,66 @@ async def create_participant(
...
-                'pid': new_id,
-                'name': name.lower(),
-                'external_id': eid,
-                'audit_log_id': audit_log_id,
-            }
-            for name, eid in external_ids.items()
-            if eid is not None
-        ]
-        await self.connection.execute_many(_eid_query, _eid_values)
+        conn = self.connection.pg_connection
```
dancoates: Can use the transaction helper from the connection here

db/python/tables/participant.py
```diff
@@ -317,99 +326,66 @@ async def create_participant(
...
-            {k: item[idx] for k, item in values.items()}
-            for idx in range(len(values[keys[0]]))
-        ]
+            cur = await conn.execute(_query)
+            row = await cur.fetchone()
+            if not row:
+                raise ValueError('Failed to create participant')
+            new_id = row['id']
+
+            _query = """
```
dancoates: tiny thing, but can we keep this names `_eid_query` to avoid confusion with the above `_query`?

db/python/tables/participant.py
```diff
@@ -423,91 +399,107 @@ async def update_participant(
...
-
-        if reported_sex:
-            updaters.append('reported_sex = :reported_sex')
-            fields['reported_sex'] = reported_sex
-
-        if reported_gender:
-            updaters.append('reported_gender = :reported_gender')
-            fields['reported_gender'] = reported_gender
+        conn = self.connection.pg_connection
+        in_transaction = conn.info.transaction_status == TransactionStatus.INTRANS
```
dancoates: Can use the transaction helper here

dancoates (approved):

dancoates (approved):
Looks great! Just a few small comments but otherwise good to go

PR #1204, opened 2026-02-10
[PG] Migrate sequencing group queries

hrrsheen:
This migrates sequencing group table methods to be compatible with postgres, as well as converts the corresponding test code to use pytest.
Additionally, unused sequencing group table methods have been removed.

Not implemented:
- Table functions that are not covered by sequencing group tests have not had tests written. This includes the following:
    - `get_sequencing_groups_by_analysis_ids`
    - `get_samples_create_date_from_sgs`
    - `get_all_sequencing_group_ids_by_sample_ids_by_type`
    - `get_type_numbers_for_project`
    - `recreate_sequencing_group_with_new_assays`
- Sequencing group tests functions that relate to Analyses have been skipped
- Sequencing group tests that rely on the use of samples across multiple projects have been skipped
    - This is only until migration of samples is complete


api/graphql/mutations/sequencing_group.py
```diff
@@ -120,5 +120,5 @@ async def archive_sequencing_groups(
             ]
             await slayer.archive_sequencing_groups(raw_ids)
 
-            updated_sgs = await slayer.get_sequencing_groups_by_ids(raw_ids)
+            updated_sgs = await slayer.get_sequencing_groups_by_ids(raw_ids, active_only=False)
```
violetbrina: Any reason for this behaviour change? It looks like active_only=True was the previous behaviour. If this is supposed to be the true default maybe change it in the method itself? 

If not, and this is the assumption when updating sgs then I would add a comment as to why.

db/python/tables/sequencing_group.py
```diff
@@ -83,63 +82,63 @@ def construct_query(
...
             }
-            awheres, avalues = filter_.assay.to_sql(a_overrides)
-            _query.extend(
-                [
-                    'INNER JOIN sequencing_group_assay sga ON sg.id = sga.sequencing_group_id',
-                    'INNER JOIN assay a ON sga.assay_id = a.id',
-                ]
+            assay_where_template = filter_.assay.to_sql(a_overrides)
+            _query.append(
+                t"""\
```
violetbrina: This is truely the tinniest of notes. As in, don't even fix it, it's fine. But you don't need the `\` for the multiline template string.

db/python/tables/sequencing_group.py
```diff
@@ -83,63 +82,63 @@ def construct_query(
...
-                        FROM
-                            analysis
-                    ) AS anlysis_query ON analysis_sequencing_group.analysis_id = anlysis_query.id
-                GROUP BY
-                    sequencing_group_id
-                HAVING
-                    {cram_wheres}
-            ) AS sg_filequery ON sg.id = sg_filequery.sequencing_group_id
-            """
+                f"""\
```
violetbrina: Template string instead of format string?

db/python/tables/sequencing_group.py
```diff
@@ -149,102 +148,76 @@ def construct_query(
...
             limit=limit,
             skip=skip,
         )
 
-        rows = await self.connection.fetch_all(query, query_values)
-        sgs = [SequencingGroupInternal.from_db(**dict(r)) for r in rows]
+        cur = await self.connection.pg_connection.execute(query)
+        rows = await cur.fetchall()
+
+        sgs = [SequencingGroupInternal(**r) for r in rows]
```
violetbrina: I think there is another way to do this. I wasn't aware until I saw it in the `family.py` table code. Don't feel any need to change this, just offering an alternative that I know I wasn't aware of if you're interested:
```python
        async with self.connection.pg_connection.cursor(
            row_factory=class_row(FamilyInternal)
        ) as cur:
            await cur.execute(t'{joined_query:q} WHERE {where_params:q} GROUP BY f.id')
            family_internal_list = await cur.fetchall()
```

db/python/tables/sequencing_group.py
```diff
@@ -405,157 +347,145 @@ async def create_sequencing_group(
...
+            'sample_id': sample_id,
+            'type': type_.lower(),
+            'technology': technology.lower(),
+            'platform': platform.lower(),
+            'meta': to_db_json(meta or {}),
+        }
+        # check if any values are None and raise an exception if so
+        bad_keys = [k for k, v in values.items() if v is None]
+        if bad_keys:
+            raise ValueError(f'Must provide values for {", ".join(bad_keys)}')
```
violetbrina, line 345: Is there a reason you're both asserting that the values are not None and also raising the value error? I'd just remove the asserts and leave the bad_keys check and value error message.

db/python/tables/sequencing_group.py
```diff
@@ -588,7 +523,7 @@ async def get_sequencing_group_counts_by_month(
         )
         for row in rows:
             project = row['project']
-            month_created: date = row['sg_date'].replace(day=1)
+            month_created: date = row['sg_date'].date().replace(day=1)
             sg_type = row['type']
             sg_tech = row['technology']
             num_sg = row['num_sg']
```
violetbrina, line 509: As an alternative to the `.date()` method here you could just add `sg_first_date::date` to the query.

violetbrina (changes requested):
A couple of very small comments, mostly nothing to change just one or two easy fixes.

db/python/tables/sequencing_group.py
hrrsheen: Ah yes, I was doing this because otherwise it inserted an extra newline when I was debugging the template string with `sql.as_string`. 
But in the end I just got an LLM to apply the correct newlines and whitespace to the raw printout, so it's not needed anymore.

api/graphql/mutations/sequencing_group.py
hrrsheen: By default `get_sequencing_groups_by_ids` will only return non-archived sequencing groups. And since this mutation archives sequencing groups, `updated_sgs` would always be empty.

hrrsheen: Actually this brings up a point about how `active_only` works. Right now, setting it to `True` will only return non-archived entries and setting it to `False` will only return archived entries.
I think the phrasing feels a little counter-intuitive to me as I'd expect `active_only=False` to return both archived and non-archived entries.

db/python/tables/sequencing_group.py
```diff
@@ -83,63 +79,63 @@ def construct_query(
...
-                HAVING
-                    {cram_wheres}
-            ) AS sg_filequery ON sg.id = sg_filequery.sequencing_group_id
-            """
+                f"""\
+                INNER JOIN (
+                    SELECT
+                        sequencing_group_id,
+                        FIND_IN_SET('cram', GROUP_CONCAT(LOWER(anlysis_query.type))) > 0 AS has_cram,
+                        FIND_IN_SET('gvcf', GROUP_CONCAT(LOWER(anlysis_query.type))) > 0 AS has_gvcf
```
dancoates: Postgres doesn't have FIND_IN_SET or GROUP_CONCAT, but you should be able to use `bool_or` to do this more efficiently anyway. At the moment the group concat has to evaluate all analyses which could be quite slow, bool_or will bail as soon as it finds a matching value.

```
select 
	asg.sequencing_group_id,
	bool_or(a.type = 'cram') as has_cram,
	bool_or(a.type = 'gvcf') as has_gvcf
from analysis_sequencing_group asg
join analysis a on a.id = asg.analysis_id
group by 1
```

db/python/tables/sequencing_group.py
```diff
@@ -83,63 +79,63 @@ def construct_query(
...
+                    )
+                    GROUP BY id
+                    HAVING {created_on_condition:q}
+                ) AS sg_timequery ON sg.id = sg_timequery.id"""
             )
 
         if filter_.has_cram is not None or filter_.has_gvcf is not None:
-            cram_wheres, cram_values = filter_.to_sql(
+            cram_where_template = filter_.to_sql(
                 sql_overrides, only=['has_cram', 'has_gvcf']
```
dancoates, line 115: I'm a bit confused as to how this was working, sql_overrides doesn't have anything for `has_cram` or `has_gvcf` but those won't work in the `having` clause?

db/python/tables/sequencing_group.py
```diff
@@ -149,102 +145,76 @@ def construct_query(
...
-            'sg.sample_id',
-            'sg.type',
-            'sg.technology',
-            'sg.platform',
-            'sg.meta',
-            'sg.archived',
+        select_keys = [
+            sql.Identifier('sg', 'id'),
+            sql.Identifier('s', 'project'),
+            # sql.SQL('jsonb_object_agg("sgexid"."name", "sgexid"."external_id") as external_ids'),
```
dancoates: Is this intended to be commented out? If it is something we want to come back to later then can you please add a TODO comment so we don't miss it?

api/graphql/mutations/sequencing_group.py
dancoates: Could you expand a bit on why `slayer.get_sequencing_groups_by_ids` is changed to return only non archived sequencing groups by default? I'm not quite following that change.

db/python/filters/generic.py
```diff
@@ -120,11 +120,17 @@ def to_sql(
         if not isinstance(column, str):
             raise ValueError(f'Column {column!r} must be a string')
 
-        column_query = (
-            column_expression
-            if column_expression
-            else t'{sql.Identifier(*column.split(".")):i}'
-        )
+        if column:
+            keywords = [segment.split('.') for segment in column.split(' ')]
```
dancoates: This isn't quite as robust as I'd like. It implies that any expression would work, but if had field overrides like:

```python
field_mapping = {'test_string': "test_string || '.123'"}
```

which is valid postgres query expression, then this would fail. 

A more robust approach would be to update the to_sql `column` arg and the `field_overrides` below to accept either a string or a template string. That way in simple cases you can express overrides like:

```python
field_mapping = {
    'test_string': 'tgf.test_string',
}
```

and for more complex ones, you would specify it as a template string.


```python
field_mapping = {
    'test_string': t'NOT test_bool',
}
```

or with the silly example I came up with above:

```python
field_mapping = {'test_string': t"test_string || '.123'"}
```

The implementation here would look like.


```python
column_query = (
    column
    if isinstance(column, Template)
    else t'{sql.Identifier(*column.split(".")):i}'
)
```


If you could delete the column_expression arg at the same time that'd be great, I don't think we need it anymore.

db/python/layers/sequencing_group.py
```diff
@@ -134,124 +128,25 @@ async def get_all_sequencing_group_ids_by_sample_ids_by_type(
...
-
-        return pids
-
     async def get_type_numbers_for_project(self, project: ProjectId) -> dict[str, int]:
         """Get sequencing type numbers (of groups) for a project"""
         return await self.seqgt.get_type_numbers_for_project(project)
 
     # region CREATE / MUTATE
 
-    async def create_sequencing_group_from_assays(
```
dancoates, line 165: Nice improvements in this file :) always nice to see deleted code and variable naming fixes!

db/python/tables/sequencing_group.py
```diff
@@ -358,37 +297,37 @@ async def get_sequencing_groups_by_analysis_ids(
...
             'sg.type',
             'sg.technology',
             'sg.platform',
             'sg.meta',
             'sg.archived',
         ]
-        _query = f"""
-        SELECT {', '.join(keys)}, asg.analysis_id
+        _query = t"""\
+        SELECT {sql.SQL(', ').join(keys):q}, asg.analysis_id
```
dancoates: Small thing, but these keys aren't used anywhere else so we could just write them in the query

db/python/tables/sequencing_group.py
```diff
@@ -405,157 +344,138 @@ async def create_sequencing_group(
...
+        async with self.connection.transaction():
             if existing_sg_ids:
                 await self.archive_sequencing_groups([s['id'] for s in existing_sg_ids])
 
-            id_of_seq_group = await self.connection.fetch_val(
-                _query,
-                {**values, 'audit_log_id': await self.audit_log_id()},
-            )
+            cur = await conn.execute(_sq_insert_query)
+            new_sg_id = await cur.fetchone()
```
dancoates, line 388: Will need a check here to ensure the row exists, fetchone won't error out if nothing is retured. In other places we raise an InternalError for this situation because if the row wasn't created than something has gone wrong.

dancoates (commented):
Nice stuff! There's a lot of work in this one. Just a few things that need updating, happy to chat if anything isn't clear.

api/graphql/mutations/sequencing_group.py
hrrsheen: My mistake, it looks like I accidentally removed the existing default `active_only=None` filter parameter during development. I was trying to fix an issue where the sequencing group queries would only return non-archived sequencing groups by default but I don't think this change actually addressed the root cause.

db/python/tables/sequencing_group.py
hrrsheen: A very cool suggestion!

db/python/tables/sequencing_group.py
hrrsheen: Ah yes this was a testing blindspot because I was waiting for the analyses queries to be migrated. I forgot to leave a `TODO` comment.

hrrsheen: I've made the migration for this section of the query now.

db/python/tables/sequencing_group.py
hrrsheen: Mainly because most of the values have `.lower()` applied to them, but you're right there's probably a better way to do this.

db/python/filters/generic.py
hrrsheen: I've made the change to how column overrides are interpreted.

db/python/tables/sequencing_group.py
hrrsheen: My mistake, that was leftover from development. I've uncommented it and the query is working in the tests.

db/python/tables/sequencing_group.py
hrrsheen: I've migrated this portion of the query now.

db/python/tables/sequencing_group.py
hrrsheen: Good call!

dancoates (approved):
Nice work, this is looking great! Just need to fix the lint/format errors and is good to merge

violetbrina (approved):
LGTM

jmarshall:
@hrrsheen: This PR removed `get_participant_ids_sequencing_group_ids_for_sequencing_type()` but it appears that is still used in _db/python/layers/seqr.py_. Possibly it shouldn't be used there, or possibly we need to reinstate it?

PR #1210, opened 2026-02-16
Migrate sample queries (SET-906)

violetbrina:
Migrate the sample queries to postgres

Also adds in a fix to the `to_sql` function in the generic filter to return `None` if the filters are empty.

### DIFF
--- a/db/python/filters/generic.py
+++ b/db/python/filters/generic.py
@@ -6,6 +6,7 @@ from typing import Any, TypeVar
 
 from psycopg import sql
 
+from db.python import filters
 from db.python.utils import escape_like_term
 from models.base import SMBase
 
@@ -292,6 +293,7 @@ class GenericFilterModel:
                         f'Filter {field.name} must be a GenericFilter or dict[str, GenericFilter]'
                     )
 
+        filters = [f for f in filters if f is not None]
         if len(filters) == 0:
             return None
 
--- a/db/python/filters/generic.py
+++ b/db/python/filters/generic.py
@@ -172,6 +172,9 @@ class GenericFilter[T](SMBase):
             else:
                 filters.append(t'{column_query:q} IS NOT NULL')
 
+        if len(filters) == 0:
+            return t''
+
         return sql.SQL(' AND ').join(filters)
 
     def transform(self, func: Callable[[T], X]) -> GenericFilter[X]:
@@ -252,7 +255,7 @@ class GenericFilterModel:
         field_overrides: dict[str, str] | None = None,
         only: list[str] | None = None,
         exclude: list[str] | None = None,
-    ) -> Template:
+    ) -> Template | None:
         """Convert the model to SQL, and avoid SQL injection"""
 
         _foverrides = field_overrides or {}
@@ -288,6 +291,9 @@ class GenericFilterModel:
                     raise ValueError(
                         f'Filter {field.name} must be a GenericFilter or dict[str, GenericFilter]'
                     )
+
+        if len(filters) == 0:
+            return None
 
         return sql.SQL(' AND ').join(filters)
 
--- a/db/python/filters/generic.py
+++ b/db/python/filters/generic.py
@@ -6,7 +6,6 @@ from typing import Any, TypeVar
 
 from psycopg import sql
 
-from db.python import filters
 from db.python.utils import escape_like_term
 from models.base import SMBase
 
@@ -103,7 +102,7 @@ class GenericFilter[T](SMBase):
 
     def to_sql(
         self, column: str, column_expression: Template | None = None
-    ) -> Template:
+    ) -> Template | None:
         """
         Convert to SQL, and avoid SQL injection
 
@@ -116,7 +115,7 @@ class GenericFilter[T](SMBase):
         Returns:
             Template
         """
-        filters: list[Template] = []
+        filters: list[Template | None] = []
 
         if not isinstance(column, str):
             raise ValueError(f'Column {column!r} must be a string')
@@ -173,10 +172,11 @@ class GenericFilter[T](SMBase):
             else:
                 filters.append(t'{column_query:q} IS NOT NULL')
 
-        if len(filters) == 0:
-            return t''
+        filters_rm_none: list[Template] = [f for f in filters if f is not None]
+        if len(filters_rm_none) == 0:
+            return None
 
-        return sql.SQL(' AND ').join(filters)
+        return sql.SQL(' AND ').join(filters_rm_none)
 
     def transform(self, func: Callable[[T], X]) -> GenericFilter[X]:
         """
@@ -271,7 +271,7 @@ class GenericFilterModel:
             )
 
         fields = dataclasses.fields(self)
-        filters: list[Template] = []
+        filters: list[Template | None] = []
         for field in fields:
             if only and field.name not in only:
                 continue
@@ -293,11 +293,11 @@ class GenericFilterModel:
                         f'Filter {field.name} must be a GenericFilter or dict[str, GenericFilter]'
                     )
 
-        filters = [f for f in filters if f is not None]
-        if len(filters) == 0:
+        filters_rm_none: list[Template] = [f for f in filters if f is not None]
+        if len(filters_rm_none) == 0:
             return None
 
-        return sql.SQL(' AND ').join(filters)
+        return sql.SQL(' AND ').join(filters_rm_none)
 
 
 def prepare_query_from_dict_field(
--- a/db/python/layers/sample.py
+++ b/db/python/layers/sample.py
@@ -7,7 +7,7 @@ from db.python.layers.assay import AssayLayer
 from db.python.layers.base import BaseLayer, Connection
 from db.python.layers.sequencing_group import SequencingGroupLayer
 from db.python.tables.sample import SampleFilter, SampleTable
-from db.python.utils import NoOpAenter, NotFoundError
+from db.python.utils import NotFoundError
 from models.models.project import (
     FullWriteAccessRoles,
     ProjectId,
@@ -285,7 +285,6 @@ class SampleLayer(BaseLayer):
     async def upsert_samples(
         self,
         samples: list[SampleUpsertInternal],
-        open_transaction: bool = True,
         project: ProjectId = None,
     ) -> list[SampleUpsertInternal]:
         """Batch upsert a list of samples with sequences"""
--- a/db/python/layers/sample.py
+++ b/db/python/layers/sample.py
@@ -306,7 +306,6 @@ class SampleLayer(BaseLayer):
                     project=project,
                     process_sequencing_groups=False,
                     process_assays=False,
-                    open_transaction=False,
                 )
 
             # Upsert all sequencing_groups (in turn relevant assays)
--- a/db/python/layers/sample.py
+++ b/db/python/layers/sample.py
@@ -291,12 +291,6 @@ class SampleLayer(BaseLayer):
         """Batch upsert a list of samples with sequences"""
         seqglayer: SequencingGroupLayer = SequencingGroupLayer(self.connection)
 
-        with_function = (
-            self.connection.pg_connection.transaction
-            if open_transaction
-            else NoOpAenter
-        )
-
         sids = [s.id for s in samples if s.id]
         if sids:
             pjcts = await self.st.get_project_ids_for_sample_ids(sids)
@@ -304,7 +298,7 @@ class SampleLayer(BaseLayer):
                 pjcts, allowed_roles=FullWriteAccessRoles
             )
 
-        async with with_function():
+        async with self.connection.transaction():
             # Create or update samples
             for sample in samples:
                 await self.upsert_sample(
--- a/db/python/layers/sample.py
+++ b/db/python/layers/sample.py
@@ -94,7 +94,7 @@ class SampleLayer(BaseLayer):
     async def get_sample_id_map_by_external_ids(
         self,
         external_ids: list[str],
-        project: ProjectId = None,
+        project: ProjectId,
         allow_missing=False,
     ) -> dict[str, int]:
         """Get map of samples {(any) external_id: internal_id}"""
@@ -225,7 +225,9 @@ class SampleLayer(BaseLayer):
     ) -> SampleUpsertInternal:
         """Upsert a sample"""
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
+            self.connection.pg_connection.transaction
+            if open_transaction
+            else NoOpAenter
         )
         if sample.id:
             pjcts = await self.st.get_project_ids_for_sample_ids([sample.id])
@@ -296,7 +298,9 @@ class SampleLayer(BaseLayer):
         seqglayer: SequencingGroupLayer = SequencingGroupLayer(self.connection)
 
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
+            self.connection.pg_connection.transaction
+            if open_transaction
+            else NoOpAenter
         )
 
         sids = [s.id for s in samples if s.id]
--- a/db/python/layers/sample.py
+++ b/db/python/layers/sample.py
@@ -221,14 +221,8 @@ class SampleLayer(BaseLayer):
         project: ProjectId | None = None,
         process_sequencing_groups: bool = True,
         process_assays: bool = True,
-        open_transaction: bool = True,
     ) -> SampleUpsertInternal:
         """Upsert a sample"""
-        with_function = (
-            self.connection.pg_connection.transaction
-            if open_transaction
-            else NoOpAenter
-        )
         if sample.id:
             pjcts = await self.st.get_project_ids_for_sample_ids([sample.id])
             self.connection.check_access_to_projects_for_ids(
@@ -242,7 +236,7 @@ class SampleLayer(BaseLayer):
             )
 
         # safely ignore nested samples here
-        async with with_function():
+        async with self.connection.transaction():
             for r in self.unwrap_nested_samples([sample]):
                 s = r.sample
                 if not s.id:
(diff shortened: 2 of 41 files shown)


db/python/layers/sample.py
```diff
@@ -225,7 +225,9 @@ async def upsert_sample(
     ) -> SampleUpsertInternal:
         """Upsert a sample"""
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
+            self.connection.pg_connection.transaction
```
hrrsheen: Are we wanting to move away from using `open_transaction` now that we can just directly check on the connection whether a transaction is in progress? Like [here](https://github.com/populationgenomics/metamist/pull/1207/changes#diff-f84cf14c6808d7043aef2bd75a52d5bc9b8d727771bb0f5e31aa032c76b5143aR136) for example

db/python/tables/sample.py
```diff
@@ -289,47 +300,46 @@ async def insert_sample(
...
-        cs_keys = ', '.join(keys)
-        cs_id_keys = ', '.join(f':{k}' for k in keys)
-        _query = f"""\
-        INSERT INTO sample
-            ({cs_keys})
-        VALUES ({cs_id_keys}) RETURNING id;
+        meta_value = to_db_json(meta or {})
+        project_value = project or self.project_id
+
+        _query = t"""
```
hrrsheen, line 304: Simple, to the point.
(referring to the query, which the preview doesn't show)

test/test_sample.py
```diff
@@ -235,19 +254,19 @@ def test_nested_sample_unwrapping_basic(self):
...
         unwrapped = SampleLayer.unwrap_nested_samples([sample])
 
-        self.assertEqual(4, len(unwrapped))
+        assert len(unwrapped) == 4
 
         first_row = unwrapped[0]
-        self.assertTupleEqual(
-            (None, None, 1),
-            (first_row.root, first_row.parent, first_row.sample.id),
+        assert (first_row.root, first_row.parent, first_row.sample.id) == (
```
hrrsheen: The original code seemed to like using these assertTupleEqual methods, but I wonder if it would be better to just be verbose and have it as separate asserts.

violetbrina: Yeah not a bad thought. If nothing else it will look a little less messy.

db/python/filters/generic.py
```diff
@@ -172,6 +172,9 @@ def to_sql(
             else:
                 filters.append(t'{column_query:q} IS NOT NULL')
 
+        if len(filters) == 0:
```
dancoates: This is interesting cos it means the types in psycopg aren't quite right, types say that `sql.SQL().join()` with a list of templates should return a `Template`. But that isn't right because if the list is empty it returns a `Composed`. This is a good fix for that inaccuracy.

db/python/filters/generic.py
```diff
@@ -289,6 +292,9 @@ def to_sql(
                         f'Filter {field.name} must be a GenericFilter or dict[str, GenericFilter]'
                     )
 
+        if len(filters) == 0:
```
dancoates: do we need to filter out empty template strings before doing this check and joining? If multiple filters have empty results then they'll be empty strings joined with ` AND ` which would be a syntax error.

db/python/layers/sample.py
```diff
@@ -296,7 +292,9 @@ async def upsert_samples(
         seqglayer: SequencingGroupLayer = SequencingGroupLayer(self.connection)
 
         with_function = (
-            self.connection.connection.transaction if open_transaction else NoOpAenter
+            self.connection.pg_connection.transaction
```
dancoates: Can use the transaction helper here too

db/python/tables/sample.py
```diff
@@ -23,121 +25,126 @@ class SampleTable(DbBase):
...
-                'project': 'ss.project',
-                'id': 'ss.id',
-                'meta': 'ss.meta',
-                'sample_root_id': 'ss.sample_root_id',
-                'sample_parent_id': 'ss.sample_parent_id',
-            },
-            exclude=['sequencing_group', 'assay', 'external_id'],
-        )
-        wheres = [_wheres]
+        wheres: list[Template] = []
```
dancoates: type should be `list[Template | None]` given that to_sql can now return None

db/python/tables/sample.py
```diff
@@ -23,121 +25,126 @@ class SampleTable(DbBase):
...
-            awheres, avalues = filter_.assay.to_sql(
-                {
-                    'id': 'a.id',
-                    'meta': 'a.meta',
-                    'type': 'a.type',
-                }
+        if filter_.sequencing_group or filter_.assay:
+            if not filter_.sequencing_group:
+                raise ValueError(
+                    'Assay filter provided without sequencing group filter, which is not supported'
```
dancoates: This doesn't appear to be a restriction in the previous code, so I think we should continue supporting assay filters without sg filter

db/python/tables/sample.py
```diff
@@ -166,10 +174,10 @@ async def query(
             's.sample_root_id',
             's.sample_parent_id',
         ]
-        _query, values = self.construct_query(
-            filter_, keys, sample_eid_table_alias='sexid'
-        )
-        rows = await self.connection.fetch_all(_query, values)
+
+        _query = self.construct_query(filter_, keys, sample_eid_table_alias='sexid')
+        cur = await self.connection.pg_connection.execute(_query)
```
dancoates: Can we use a row cursor here for a cleaner implementation of converting row to SampleInternal

db/python/tables/sample.py
```diff
@@ -216,17 +224,18 @@ async def get_sample_id_to_project_map(
...
-            _query += ' AND active IS TRUE'
+        _query = t'SELECT id, project FROM sample WHERE project = ANY({project_ids})'
+        _query += t' AND active IS TRUE' if active_only else t''
 
-        rows = await self.connection.fetch_all(_query, {'project_ids': project_ids})
+        cur = await self.connection.pg_connection.execute(_query)
+        rows = await cur.fetchall()
         return {row['id']: row['project'] for row in rows}
 
+    # TODO: Confirm this works once the MetaTable is also ported
```
dancoates: Can remove this comment as meta table is ported now

db/python/tables/sample.py
```diff
@@ -288,47 +297,46 @@ async def insert_sample(
...
-        ]
-
-        keys = [k for k, _ in kv_pairs]
-        cs_keys = ', '.join(keys)
-        cs_id_keys = ', '.join(f':{k}' for k in keys)
-        _query = f"""\
-        INSERT INTO sample
-            ({cs_keys})
-        VALUES ({cs_id_keys}) RETURNING id;
+        meta_value = to_db_json(meta or {})
```
dancoates: Can we use psycopg's Jsonb json adaption here

db/python/tables/sample.py
```diff
@@ -288,47 +297,46 @@ async def insert_sample(
...
 
-        id_of_new_sample = await self.connection.fetch_val(
-            _query,
-            dict(kv_pairs),
-        )
+        try:
+            cur = await self.connection.pg_connection.execute(_query)
+            row = await cur.fetchone()
+            assert row is not None and 'id' in row
+        except DatabaseError as e:
```
dancoates: I might be misunderstanding, but this try/except doesn't seem to be doing anything here?

db/python/tables/sample.py
```diff
@@ -288,47 +297,46 @@ async def insert_sample(
...
+                {audit_log_id}, {sample_parent_id}, {sample_root_id}, {project_value})
+            RETURNING id;
         """
 
-        id_of_new_sample = await self.connection.fetch_val(
-            _query,
-            dict(kv_pairs),
-        )
+        try:
+            cur = await self.connection.pg_connection.execute(_query)
```
dancoates, line 313: There's multiple mutation queries happening in one method here, can we wrap them in a transaction please

db/python/tables/sample.py
```diff
@@ -363,76 +369,77 @@ async def update_sample(
...
-                _query = 'SELECT project FROM sample WHERE id = :id'
-                row = await self.connection.fetch_one(_query, {'id': id_})
+                _query = t'SELECT project FROM sample WHERE id = {id_}'
+
+                try:
+                    cur = await self.connection.pg_connection.execute(_query)
+                    row = await cur.fetchone()
+                    assert row is not None and 'project' in row
+                except DatabaseError as e:
+                    raise DatabaseError(
```
dancoates: Is there a particular reason to wrap this query and change the error message? We don't do this on any other queries and this one is very unlikely to fail so not sure it's worth the extra verbosity?

db/python/tables/sample.py
```diff
@@ -363,76 +369,77 @@ async def update_sample(
...
-        if sample_root_id is not None:
-            values['sample_root_id'] = sample_root_id
-            fields.append('sample_root_id = :sample_root_id')
+                async with self.connection.pg_connection.cursor() as cur:
+                    await cur.executemany(_update_query, _eid_values)
+
+        # Type, Active, Sample Parent ID, Sample Root ID
+        fields.append(t'type = {type_}') if type_ else None
+        fields.append(
+            t'meta = JSON_MERGE_PATCH(COALESCE(meta, "{{}}"), {meta})'
```
dancoates: I don't think the double quotes here will work, and `meta` will need to be passed through the `Jsonb` adaption method

db/python/tables/sample.py
```diff
@@ -474,57 +481,51 @@ def dict_merge(meta1, meta2):
...
+        # Query to update the kept sample with the merged meta and audit log id
+        queries.append(t"""
             UPDATE sample
-            SET audit_log_id = :audit_log_id,
-                meta = :meta
-            WHERE id = :id
-        """
-        _query_seqs = """
+            SET audit_log_id = {audit_log_id},
+                meta = {meta}
```
dancoates: meta will need to be passed through `Jsonb` adaption

db/python/tables/sample.py
```diff
@@ -559,21 +564,18 @@ async def search(
...
-        """.strip()
-        rows = await self.connection.fetch_all(
-            _query,
-            {
-                'project_ids': project_ids,
-                'search_pattern': escape_like_term(query) + '%',
-                'limit': limit,
-            },
-        )
+            WHERE s.project IN ({project_ids}) AND seid.external_id LIKE {search_pattern}
```
dancoates: Can we switch this to ILIKE so it matches the mariadb functionality

dancoates: Also `IN` won't work, switch to ANY

db/python/tables/sample.py
```diff
@@ -648,13 +656,15 @@ async def get_samples_create_date(self, sample_ids: list[int]) -> dict[int, date
...
-        _query = """
-        SELECT id, min(row_start)
-        FROM sample FOR SYSTEM_TIME ALL
-        WHERE id in :sids
-        GROUP BY id"""
-        rows = await self.connection.fetch_all(_query, {'sids': sample_ids})
-        return {r[0]: r[1].date() for r in rows}
+        _query = t"""
+            SELECT id, min(row_start) as date_created
+            FROM sample FOR SYSTEM_TIME ALL
```
dancoates: Will need to query the history table for this

db/python/tables/sample.py
```diff
@@ -663,21 +673,22 @@ async def get_history_of_sample(self, id_: int):
         # see into one aggregate record. For now, leave the query as is, with external_ids unavailable.
         keys = [
             'id',
-            """'{"(not available)": "(not available)"}' AS external_ids""",
+            sql.SQL("""'{"(not available)": "(not available)"}' AS external_ids"""),
```
dancoates: I think this probably won't work, would at least need casting to JSONB but might also need to use `jsonb_build_object`?

db/python/tables/sample.py
```diff
@@ -663,21 +673,22 @@ async def get_history_of_sample(self, id_: int):
...
         ]
-        keys_str = ', '.join(keys)
-        _query = f'SELECT {keys_str} FROM sample FOR SYSTEM_TIME ALL WHERE id = :id'
+        keys_str = sql.SQL(', ').join(SampleTable.format_keys(keys))
+        _query = t'SELECT {keys_str:q} FROM sample FOR SYSTEM_TIME ALL WHERE id = {id_}'
 
-        rows = await self.connection.fetch_all(_query, {'id': id_})
+        cur = await self.connection.pg_connection.execute(_query)
+        rows = await cur.fetchall()
         samples = [SampleInternal.from_db(dict(d)) for d in rows]
```
dancoates: Would be nice to use a a row cursor here to convert to SampleInternal

db/python/tables/sample.py
```diff
@@ -704,14 +715,13 @@ async def get_monthly_samples_count_per_project(
...
 
-        where_str = 'WHERE project in :project_ids' if project_ids else ''
-        values = {'project_ids': project_ids} if project_ids else {}
+        where_str = t'WHERE project = ANY({project_ids})' if project_ids else ''
 
-        _query = f"""
+        _query = t"""
         WITH t AS(
             SELECT project, id, min(row_start) as sample_first_date
             FROM sample FOR SYSTEM_TIME ALL
```
dancoates: Will need to query history table for this

db/python/tables/sample.py
```diff
@@ -663,21 +673,22 @@ async def get_history_of_sample(self, id_: int):
...
             'project',
             'author',
             'sample_root_id',
             'sample_parent_id',
             # 'audit_log_id',  # TODO SampleInternal does not allow an audit_log_id field
         ]
-        keys_str = ', '.join(keys)
-        _query = f'SELECT {keys_str} FROM sample FOR SYSTEM_TIME ALL WHERE id = :id'
+        keys_str = sql.SQL(', ').join(SampleTable.format_keys(keys))
+        _query = t'SELECT {keys_str:q} FROM sample FOR SYSTEM_TIME ALL WHERE id = {id_}'
```
dancoates: Need to query history table here

dancoates (commented):
This is looking pretty close, though there were quite a few spots with syntax that won't work with postgres, given that these weren't picked up by the tests it'd indicate test coverage is pretty low - we might need to add more tests for these, if not in this PR then in a more general effort to bump coverage after we've got some more PRs merged.

db/python/filters/generic.py
violetbrina: Fascinating, I didn't realise that the method indicated it should return a `Template` type. But this will work for now.

db/python/filters/generic.py
violetbrina: Ah good thought, I'll add that in as well.

db/python/tables/sample.py
violetbrina: Oh now that's nifty! Didn't know you could do that. Absolutely.

db/python/tables/sample.py
violetbrina: I feel like I'm getting deja vu. Either I already fixed this or I'm thinking of the participant PR 😆

db/python/tables/sample.py
violetbrina: Ah I think I was using this with a different message for debugging and just forgot to remove it.

db/python/tables/sample.py
violetbrina: Fixed

db/python/tables/sample.py
violetbrina: Oh of course! Thank you, I totally missed this.

db/python/tables/sample.py
violetbrina: Done!

db/python/tables/sample.py
violetbrina: I'll convert it use the json_build_object and see how that goes.

violetbrina: I've changed it just be an empty json, let me know if we're ok with that change. The 'not available' key and value seemed a bit weird anyway...

dancoates: Yeah I think that's reasonable - I suspect this isn't really used so changing very old weird things hopefully won't break anything.

db/python/filters/generic.py
```diff
@@ -6,6 +6,7 @@
 
 from psycopg import sql
 
+from db.python import filters
```
dancoates: Accidental extra import here I think

db/python/filters/generic.py
```diff
@@ -289,6 +293,10 @@ def to_sql(
                         f'Filter {field.name} must be a GenericFilter or dict[str, GenericFilter]'
                     )
 
+        filters = [f for f in filters if f is not None]
```
dancoates: `f` won't be `None` it'll be an empty template string

db/python/tables/sample.py
```diff
@@ -663,22 +680,41 @@ async def get_history_of_sample(self, id_: int):
...
-        keys_str = ', '.join(keys)
-        _query = f'SELECT {keys_str} FROM sample FOR SYSTEM_TIME ALL WHERE id = :id'
 
-        rows = await self.connection.fetch_all(_query, {'id': id_})
-        samples = [SampleInternal.from_db(dict(d)) for d in rows]
+        keys_str = sql.SQL(', ').join(SampleTable.format_keys(keys))
+        _query = t"""
+            SELECT {keys_str:q}
+            FROM (
+                SELECT *
```
dancoates, line 700: I know this is to avoid the verbosity of listing the columns twice, but `SELECT *` is pretty risky - if we were to add a new column to the sample table this would break straight way as all our models have `ConfigDict(extra='forbid')` set so extra keys will raise an error

dancoates (approved):

dancoates (approved):
Looks good, just a few small things left to fix, along with the linting and then should be good to go

db/python/tables/sample.py
dancoates: derp, ignore me - I missed the outer select

db/python/filters/generic.py
violetbrina: Ah it can be, since `filter_.to_sql` can return None. Just had to update the return type.

dancoates: Hm, I can't see anywhere in to_sql that returns None?

dancoates: Oh gotcha, sorry was looking before the most recent commit

PR #1211, opened 2026-02-16
Add transaction helper to remove boilerplate for nested transactions

dancoates:
This helper on the connection class allows us to conditionally enter a transaction if not already in one. Psycopg supports nested transactions with postgres savepoints but partial rollbacks is not a feature that we actually need - so in most cases it is fine to just have a single parent transaction. There's a bit of code involved in checking the current transaction status so this wraps that in a method on the connection model to save having to write it each time.

### DIFF
--- a/db/python/connect.py
+++ b/db/python/connect.py
@@ -7,8 +7,10 @@ import json
 import logging
 import os
 from collections.abc import Iterable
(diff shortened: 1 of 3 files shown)


rameshka (approved):
Thanks for this work @dancoates . It is looking good.

PR #1212, opened 2026-02-17
[SET-898] Migrate Audit log queries

rameshka:
This PR migrate Audit log table-related SQL queries in the `audit_log.py` and `base.py` to PostgreSQL.

### DIFF
--- a/api/graphql/mutations/participant.py
+++ b/api/graphql/mutations/participant.py
@@ -108,16 +108,22 @@ class ParticipantMutations:
         from api.graphql.schema import GraphQLParticipant  # noqa: PLC0415
 
         async with info.context['get_connection']() as connection:
-            connection.check_access_to_projects_for_names(
+            projects = connection.get_and_check_access_to_projects_for_names(
                 [project], FullWriteAccessRoles
             )
+            projects = list(projects)
+
+            if len(projects) < 1:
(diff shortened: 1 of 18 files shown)


violetbrina (commented):
Nice! Are there any existing tests that encompass these changes to the audit log queries? If not directly at least indirectly? It would be good just to confirm this change works.

violetbrina:
> Nice! Are there any existing tests that encompass these changes to the audit log queries? If not directly at least indirectly? It would be good just to confirm this change works.

Or even just a manual test that also would probably be fine for now if we make a note to add test coverage later :)

rameshka:
> > Nice! Are there any existing tests that encompass these changes to the audit log queries? If not directly at least indirectly? It would be good just to confirm this change works.
> 
> Or even just a manual test that also would probably be fine for now if we make a note to add test coverage later :)

@violetbrina, thank you for taking the time to review the PR.
There aren’t any explicit tests covering these queries at the moment. However, I manually tested the database layer functions to test the updated queries.

And, yes, it will be good to have these functions covered.
(I also noticed that some functions in `audit_log.py` are also unused.)

dancoates (approved):
This looks good, agree that we need to ensure these are covered by tests, but I suspect the coverage will come from other related tests. We can check coverage once some other PRs are merged and make sure that these are covered.

PR #1213, opened 2026-02-17
SET-902 Migrate meta table queries

dancoates:
This probably isn't as performant as it could be as it is converting the jsonb columns to text, only to then parse them again in duckdb. We could probably get them as dicts and then create an arrow table from that, but would need to manually define the schema for the table which would be a bit tricky. There's a rough plan to change how these meta tables work in the future so don't want to do much of a refactor for now as long as performance isn't too bad when we test in production.

All tests are skipped for now, but should work in theory. I've manually tested the endpoints and they seem to work fine.

### DIFF
--- a/db/python/tables/meta_table.py
+++ b/db/python/tables/meta_table.py
@@ -1,5 +1,6 @@
 from collections.abc import Callable
 from io import BytesIO, StringIO
+from string.templatelib import Template
 from typing import Any
 
 # Unfortunately some of these libs have partially or completely missing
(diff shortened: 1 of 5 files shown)


db/python/tables/meta_table.py
```diff
@@ -99,7 +94,7 @@ async def entity_meta_table(
                     meta_rows_str,
                     map_inference_threshold=-1,
                     format='newline_delimited',
-                ).arrow(),
+                ).fetch_arrow_table(),
```
violetbrina, line 97: Is `fetch_arrow_table` a syntax change or a genuinely different method? Just curious.

violetbrina (approved):
LGTM

hrrsheen (approved):
LGTM

db/python/tables/meta_table.py
dancoates: It was a breaking change in duckdb that we hit due to upgrading for python 3.14 - `.arrow()` changed to return am arrow `RecordBatchReader` rather than an arrow table. I think it was changed to allow for lazy evaluation but that meant that the results were never actually flushed so the request would hang indefinitely.

PR #1214, opened 2026-02-17
[SET-899] Migrate cohort queries 

rameshka:
This PR migrates the `cohort` table to PostgreSQL. Along with the query changes,  `test_cohort.py` and
`test_cohort_status.py` files are also updated. However, most of the testcases in these classes are skipped for now as they are dependent on `Sample` and `Sequencing_Group` entities. 
Therefore, each query level update was also verified manually.

Additionally, `test_mutations.py` also covers parts of the cohort layer. Since this test class depends on several other entities, it would be more appropriate to update it once those related migrations are completed.

### DIFF
--- a/db/python/tables/cohort.py
+++ b/db/python/tables/cohort.py
@@ -153,7 +153,7 @@ class CohortTable(DbBase):
         Get a cohort template by ID
         """
         _query = t"""
-        SELECT id as id, name, description, criteria, project FROM cohort_template WHERE id = {template_id}
+        SELECT id, name, description, criteria, project FROM cohort_template WHERE id = {template_id}
         """
         template = await (
             await self.connection.pg_connection.execute(
--- a/db/python/tables/cohort.py
+++ b/db/python/tables/cohort.py
@@ -134,8 +134,8 @@ class CohortTable(DbBase):
 
         async with self.connection.pg_connection.cursor(
             row_factory=class_row(CohortTemplateInternal)
-        ) as cursor:
-            cohort_templates = await (await cursor.execute(_query)).fetchall()
+        ) as cur:
+            cohort_templates = await (await cur.execute(_query)).fetchall()
 
         projects = {c.project for c in cohort_templates}
         return projects, cohort_templates
--- a/db/python/tables/cohort.py
+++ b/db/python/tables/cohort.py
@@ -1,10 +1,13 @@
 import dataclasses
 import datetime
-from typing import Any
+from string.templatelib import Template
+
+from psycopg import sql
+from psycopg.rows import scalar_row
 
 from db.python.filters import GenericFilter, GenericFilterModel
 from db.python.tables.base import DbBase
-from db.python.utils import NotFoundError, to_db_json
+from db.python.utils import NotFoundError
 from models.base import parse_sql_bool
 from models.enums.cohort import CohortStatus, CohortUpdateStatus
 from models.models.cohort import (
@@ -49,19 +52,6 @@ class CohortTable(DbBase):
     Capture Cohort table operations and queries
     """
 
-    table_name = 'cohort'
-    cohort_get_keys = [
-        'c.id as c_id',
-        'c.name as c_name',
-        'c.template_id as c_template_id',
-        'c.description as c_description',
-        'c.author as c_author',
-        'c.project as c_project',
-        'c.status as c_status',
-    ]
-
-    template_keys = ['id', 'name', 'description', 'criteria', 'project']
-
     async def query(
         self, filter_: CohortFilter
     ) -> tuple[list[CohortInternal], set[ProjectId]]:
@@ -70,7 +60,7 @@ class CohortTable(DbBase):
         filter_status = filter_.status
         filter_.status = None  # reset filter and use to filter on the rows fetched
 
-        wheres, values = filter_.to_sql(
+        where_params: Template = filter_.to_sql(
             field_overrides={
                 'id': 'c.id',
                 'name': 'c.name',
@@ -79,13 +69,18 @@ class CohortTable(DbBase):
                 'project': 'c.project',
             }
         )
-        if not wheres:
+        if not list(where_params):
             raise ValueError(f'Invalid filter: {filter_}')
 
-        cohort_get_keys_str = ','.join(self.cohort_get_keys)
-
-        _query = f"""
-        SELECT {cohort_get_keys_str},
+        _query = t"""
+        SELECT
+        c.id as c_id,
+        c.name as c_name,
+        c.template_id as c_template_id,
+        c.description as c_description,
+        c.author as c_author,
+        c.project as c_project,
+        c.status as c_status,
         exists (
             select 1
             from cohort_sequencing_group csg
@@ -97,15 +92,16 @@ class CohortTable(DbBase):
             and (sg.archived or not s.active)
         ) as is_invalid
         FROM cohort c
-        WHERE {wheres}
+        WHERE {where_params:q}
         """
 
-        rows = await self.connection.fetch_all(_query, values)
+        rows = await (
+            await self.connection.pg_connection.connection.execute(_query)
+        ).fetchall()
         cohorts_list = []
-        for row in rows:
-            row_dict = dict(row)
-            is_active = row_dict['c_status'] == CohortStatus.active.value
-            is_invalid = parse_sql_bool(row_dict['is_invalid'])
+        for cohort_row in rows:
+            is_active = cohort_row['c_status'] == CohortStatus.active.value
+            is_invalid = parse_sql_bool(cohort_row['is_invalid'])
 
             if is_active:
                 cohort_status = (
@@ -115,7 +111,7 @@ class CohortTable(DbBase):
                 cohort_status = CohortStatus.archived
 
             if _custom_matches_filter(cohort_status, filter_status):
-                cohorts_list.append(CohortInternal.from_db(row_dict, cohort_status))
+                cohorts_list.append(CohortInternal.from_db(cohort_row, cohort_status))
 
         projects = {c.project for c in cohorts_list}
         return cohorts_list, projects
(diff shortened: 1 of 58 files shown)


db/python/tables/cohort.py
```diff
@@ -248,29 +225,23 @@ async def create_cohort(
...
-                sequencing_group_ids=sequencing_group_ids,
-            )
+        return NewCohortInternal(
+            dry_run=False,
+            cohort_id=cohort_id,
+            sequencing_group_ids=sequencing_group_ids,
+        )
 
-    async def get_cohort_by_id(self, cohort_id: int) -> CohortInternal:
+    async def get_template_by_cohort_by_id(self, cohort_id: int) -> int:
```
rameshka, line 229: This function was only used to retrieve the `template_id` given the `cohort_id`. Therefore, simplified the function accordingly.

db/python/tables/cohort.py
```diff
@@ -124,73 +115,65 @@ async def get_cohort_sequencing_group_ids(self, cohort_id: int) -> list[int]:
...
         self, filter_: CohortTemplateFilter
     ) -> tuple[set[ProjectId], list[CohortTemplateInternal]]:
         """Query CohortTemplates"""
-        wheres, values = filter_.to_sql(field_overrides={})
-        if not wheres:
+
+        wheres_params: Template = filter_.to_sql(field_overrides={})
+
+        if not list(wheres_params):
             raise ValueError(f'Invalid filter: {filter_}')
```
violetbrina: Any reason you make it a list here in the if statement?

db/python/tables/cohort.py
```diff
@@ -124,73 +115,65 @@ async def get_cohort_sequencing_group_ids(self, cohort_id: int) -> list[int]:
...
-        FROM cohort_template
-        WHERE {wheres}
-        """
+        _query = t'SELECT id, name, description, criteria, project FROM cohort_template WHERE {wheres_params:q}'
+
+        rows = await (await self.connection.pg_connection.execute(_query)).fetchall()
+        cohort_templates = []
+        for row in rows:
+            row['criteria'] = row.get('criteria') or {}
+            cohort_templates.append(CohortTemplateInternal(**row))
```
violetbrina: Cast the `criteria` field to `::jsonb` and try using the `row_factory`

violetbrina (changes requested):
A couple of very small comments from me. Looking good.

db/python/tables/cohort.py
rameshka: @violetbrina I had this logic before the generic filter `to_sql` function returned none. Now, as that is already merged, I will update this logic.

db/python/tables/cohort.py
rameshka: thanks @violetbrina . 
I was overthinking here. I was trying to handle a case if a entry was inserted with `'null'::jsonb` (from Jsonb(None)) as the value of the criteria column. (When fetching such an entry, it will be cast to `None` but `CohortTemplateInternal.criteria` can not be `None`). 
But then realised, it can go wrong in many other ways if the database table doesn't have proper values in the `criteria` column. So updated it to use `row_factory` with `class_row`.

db/python/tables/cohort.py
rameshka: Also updated `family_participants.py`

db/python/tables/cohort.py
```diff
@@ -70,7 +61,7 @@ async def query(
         filter_status = filter_.status
         filter_.status = None  # reset filter and use to filter on the rows fetched
 
-        wheres, values = filter_.to_sql(
+        where_params: Template = filter_.to_sql(
```
dancoates: type should be `Template | None`

db/python/tables/cohort.py
```diff
@@ -97,15 +86,16 @@ async def query(
             and (sg.archived or not s.active)
         ) as is_invalid
         FROM cohort c
-        WHERE {wheres}
+        WHERE {where_params:q}
         """
 
-        rows = await self.connection.fetch_all(_query, values)
+        rows = await (
+            await self.connection.pg_connection.connection.execute(_query)
```
dancoates: One too many `connection`'s here

db/python/tables/cohort.py
```diff
@@ -97,15 +86,16 @@ async def query(
...
+        ).fetchall()
         cohorts_list = []
-        for row in rows:
-            row_dict = dict(row)
-            is_active = row_dict['c_status'] == CohortStatus.active.value
-            is_invalid = parse_sql_bool(row_dict['is_invalid'])
+        for cohort_row in rows:
+            is_active = cohort_row['status'] == CohortStatus.active.value
+            is_invalid = parse_sql_bool(cohort_row.pop('is_invalid'))
```
dancoates: Shouldn't need to parse_sql_bool here, it should be converted to a python bool automatically

db/python/tables/cohort.py
```diff
@@ -124,73 +115,64 @@ async def get_cohort_sequencing_group_ids(self, cohort_id: int) -> list[int]:
...
         return [row['sequencing_group_id'] for row in rows]
 
     async def query_cohort_templates(
         self, filter_: CohortTemplateFilter
     ) -> tuple[set[ProjectId], list[CohortTemplateInternal]]:
         """Query CohortTemplates"""
-        wheres, values = filter_.to_sql(field_overrides={})
-        if not wheres:
+
+        wheres_params: Template = filter_.to_sql(field_overrides={})
```
dancoates: Type should be `Template | None`

dancoates: Or you can just leave the type annotation off, it'll be inferred by the return type of to_sql

db/python/tables/cohort.py
```diff
@@ -124,73 +115,64 @@ async def get_cohort_sequencing_group_ids(self, cohort_id: int) -> list[int]:
...
+
+        if not list(wheres_params):
```
dancoates: This seems to be still converting to a list? Should it be `if wheres_params is None:`?

db/python/tables/cohort.py
dancoates: Or can leave off as the type can be inferred

test/test_cohort.py
```diff
@@ -21,21 +23,22 @@
 )
 from models.utils.cohort_id_format import cohort_id_format
 from models.utils.sequencing_group_id_format import sequencing_group_id_format
-from test.testbase import DbIsolatedTest, run_as_sync
+from test.conftest import GraphQLQueryFunction
 
 
-class TestCohortBasic(DbIsolatedTest):
+@pytest.mark.skip(reason='Skipped until SG table migrated to PostgreSQL')
```
dancoates: These tests should be able to be unskipped now that samples and sgs are merged

dancoates (commented):
This is looking good! just a few small things and would be good to unskip the tests before we merge.

db/python/tables/cohort.py
rameshka: Thanks for catching that.

db/python/tables/cohort.py
rameshka: thanks. Have missed it.

rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1214/commits/8cae6640a95b2587bc5373038d3ec4c0201480ca

db/python/tables/cohort.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1214/commits/8cae6640a95b2587bc5373038d3ec4c0201480ca

db/python/tables/cohort.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1214/commits/8cae6640a95b2587bc5373038d3ec4c0201480ca

db/python/tables/cohort.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1214/commits/8cae6640a95b2587bc5373038d3ec4c0201480ca

db/python/tables/cohort.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1214/commits/8cae6640a95b2587bc5373038d3ec4c0201480ca

db/python/tables/cohort.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1214/changes/183b4c71e9d63fab9863661d1dd07b0be6456a71

db/python/tables/cohort.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1214/changes/183b4c71e9d63fab9863661d1dd07b0be6456a71 and https://github.com/populationgenomics/metamist/pull/1214/changes/6f80d28afa5c798e8562dd4e19e51a6613778a41

test/test_cohort.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1214/changes/021bf95d078aa2083f2c3ea8153fa2c3e437c5ca and https://github.com/populationgenomics/metamist/pull/1214/changes/af7ae8cdc776550c9eaf125ca7cd1afdcfa040b6

rameshka: @dancoates, while unskipping those test cases, I found an issue in the CohortTable.query logic.

If the CohortFilter contains only a status filter, the previous implementation would fetch all cohorts and then filter them (in the application) by the specified status (since we have both derived cohort status and cohort statuses stored in the database). CohortFilter is empty; the previous `filter_.to_sql` returned true, which allowed the query to retrieve all cohorts.

With the updated `filter_.to_sql` behaviour, it now returns None when the CohortFilter is empty. To preserve the same functionality, I updated the query logic.

dancoates (approved):
Looks good!

rameshka:
Thanks @violetbrina and @dancoates for reviewing.

violetbrina (approved):
LGTM :)

PR #1216, opened 2026-02-18
SET-900 Migrate comment queries to postgres

dancoates:
Most tests skipped at the moment but should in theory work when re-enabled as more things are migrated.

### DIFF
--- a/db/python/tables/comment.py
+++ b/db/python/tables/comment.py
@@ -1,4 +1,7 @@
 from itertools import groupby
+from string.templatelib import Template
+
+from psycopg import sql
 
 from db.python.tables.base import DbBase
 from db.python.utils import InternalError, NotFoundError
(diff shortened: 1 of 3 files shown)


db/python/tables/comment.py
```diff
@@ -447,55 +476,48 @@ async def add_comment_to_entity(
...
-            INSERT INTO comment (content, status, audit_log_id)
-            VALUES (:content, 'active', :audit_log_id) RETURNING id;
-        """
+        audit_log_id = await self.connection.audit_log_id()
 
-        join_insert = f"""
-            INSERT INTO {join_table} (comment_id, {join_column}, audit_log_id)
-            VALUES (:comment_id, :entity_id, :audit_log_id);
-        """
+        audit_log_id = await self.connection.audit_log_id()
```
violetbrina: Probably don't need two?
```suggestion
        audit_log_id = await self.connection.audit_log_id()
```

violetbrina (approved):
One very small change. Otherwise LGTM.

db/python/tables/comment.py
dancoates: Oops! Thank you - good pick up. the diff on the suggested change looks like it is deleting extra code - probably just a github thing but I've fixed manually just in case.

PR #1217, opened 2026-02-18
temporarily disable mypy type checking

dancoates:
this'll enable us to enforce that other  linting checks pass while working on the migration


rameshka (approved):

violetbrina (approved):
Down with mypy! (for now haha...)

PR #1218, opened 2026-02-18
SET-896 Migrate analysis queries

violetbrina:
Migrate the analysis queries to use postgres, and migrate all the tests to pytest.

Discovered and fixed two minor bugs in `participant.py` and `sequencing_group.py` also.

### DIFF
--- a/db/python/tables/analysis.py
+++ b/db/python/tables/analysis.py
@@ -1,11 +1,10 @@
 import dataclasses
 import datetime
 from collections import defaultdict
-from string.templatelib import Template
 from typing import Any
 
 from psycopg import sql
-from psycopg.pq import Jsonb
+from psycopg.types.json import Jsonb
 
 from db.python.filters import (
     GenericFilter,
--- a/db/python/tables/analysis.py
+++ b/db/python/tables/analysis.py
@@ -1,7 +1,11 @@
 import dataclasses
 import datetime
 from collections import defaultdict
+from string.templatelib import Template
 from typing import Any
+
+from psycopg import sql
+from psycopg.pq import Jsonb
 
 from db.python.filters import (
     GenericFilter,
@@ -46,10 +50,11 @@ class AnalysisTable(DbBase):
         self, analysis_ids: list[int]
     ) -> set[ProjectId]:
         """Get project IDs for sampleIds (mostly for checking auth)"""
-        _query = (
-            'SELECT project FROM analysis WHERE id in :analysis_ids GROUP BY project'
-        )
-        rows = await self.connection.fetch_all(_query, {'analysis_ids': analysis_ids})
+        _query = t"""
+            SELECT project FROM analysis WHERE id = ANY({analysis_ids}) GROUP BY project
+        """
+        acur = await self.connection.pg_connection.execute(_query)
+        rows = await acur.fetchall()
         return set(r['project'] for r in rows)
 
     async def create_analysis(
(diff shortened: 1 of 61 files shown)


db/python/tables/analysis.py
```diff
@@ -65,36 +69,33 @@ async def create_analysis(
...
-            _query = f"""\
-INSERT INTO analysis
-    ({cs_keys})
-VALUES ({cs_id_keys}) RETURNING id;"""
-
-            id_of_new_analysis = await self.connection.fetch_val(
-                _query,
-                dict(kv_pairs),
-            )
+            filtered_pairs = {k: v for k, v in kv_pairs.items() if v is not None}
```
hrrsheen: `project` is the only non-nullable column (that we set manually, there's also `id` and `sys_period`) on the analysis table, so I don't think we actually need to filter for non-`None` values here.

db/python/tables/analysis.py
```diff
@@ -355,24 +354,27 @@ async def get_latest_complete_analysis_for_type(
...
+            )
+        """
+        acur = await self.connection.pg_connection.execute(_query)
+        rows = await acur.fetchall()
+
         if len(rows) == 0:
             raise NotFoundError(f"Couldn't find any analysis with type {analysis_type}")
+
         latest_analysis = rows[0]
         latest_analysis_data = dict(latest_analysis)
```
hrrsheen: `rows` should already be a dict, so there's no need to cast it here.

db/python/tables/analysis.py
```diff
@@ -65,36 +69,33 @@ async def create_analysis(
...
-                _query,
-                dict(kv_pairs),
-            )
+            filtered_pairs = {k: v for k, v in kv_pairs.items() if v is not None}
+            cs_keys = sql.SQL(', ').join(filtered_pairs.keys())
+            cs_values = sql.SQL(', ').join(t'{k}' for k in filtered_pairs)
+
+            _query = t"""
+                INSERT INTO analysis ({cs_keys})
+                VALUES ({cs_values})
```
hrrsheen: `cs_keys` and `cs_values` are queries, so you'll need the `:q` flag in the template.

db/python/tables/analysis.py
```diff
@@ -65,36 +69,33 @@ async def create_analysis(
...
-    ({cs_keys})
-VALUES ({cs_id_keys}) RETURNING id;"""
-
-            id_of_new_analysis = await self.connection.fetch_val(
-                _query,
-                dict(kv_pairs),
-            )
+            filtered_pairs = {k: v for k, v in kv_pairs.items() if v is not None}
+            cs_keys = sql.SQL(', ').join(filtered_pairs.keys())
+            cs_values = sql.SQL(', ').join(t'{k}' for k in filtered_pairs)
```
hrrsheen: This will populate `cs_values` with the keys of `filtered_pairs`.

db/python/tables/analysis.py
```diff
@@ -65,36 +69,33 @@ async def create_analysis(
...
-INSERT INTO analysis
-    ({cs_keys})
-VALUES ({cs_id_keys}) RETURNING id;"""
-
-            id_of_new_analysis = await self.connection.fetch_val(
-                _query,
-                dict(kv_pairs),
-            )
+            filtered_pairs = {k: v for k, v in kv_pairs.items() if v is not None}
+            cs_keys = sql.SQL(', ').join(filtered_pairs.keys())
```
hrrsheen: This will result in the keys being joined using single quotes, so cs_keys will be interpreted as a sequence of string values:
```
'type', 'status', 'meta', 'audit_log_id', 'project', 'active', 'timestamp_completed'
```

You'll want to make sure they're interpreted as identifiers, something like:
```python
cs_keys = sql.SQL(', ').join(sql.Identifier(k) for k in filtered_pairs.keys())
```

db/python/tables/analysis.py
```diff
@@ -136,72 +138,65 @@ async def add_cohorts_to_analysis(self, analysis_id: int, cohort_ids: list[int])
...
         if meta is not None and len(meta) > 0:
-            fields['meta'] = to_db_json(meta)
-            setters.append('meta = JSON_MERGE_PATCH(COALESCE(meta, "{}"), :meta)')
+            meta = Jsonb(meta)
+            setters.append(t'meta = JSON_MERGE_PATCH(COALESCE(meta, "{{}}"), {meta})')
 
-        fields_str = ', '.join(setters)
-        _query = f'UPDATE analysis SET {fields_str} WHERE id = :analysis_id'
+        fields_str = sql.SQL(', ').join(setters)
+        _query = t'UPDATE analysis SET {fields_str} WHERE id = {analysis_id}'
```
hrrsheen: `fields_str` here needs a `:q`.

hrrsheen (changes requested):
For the most part I just have a few comments about errors I received when I ran the `test_analysis_output_files.py` tests against this branch. Otherwise looks good!

hrrsheen (approved):
LGTM!

PR #1219, opened 2026-02-19
Task/[SET-987] Migrate Analysis Runner queries

rameshka:
This PR migrate analysis_runner queries.

### DIFF
--- a/db/python/tables/analysis_runner.py
+++ b/db/python/tables/analysis_runner.py
@@ -36,9 +36,7 @@ class AnalysisRunnerTable(DbBase):
         """
 
         where_params: Template = filter_.to_sql()
-
-        if where_params is None:
-            raise ValueError(f'Invalid filter: {filter_}')
+        wheres_query = t'WHERE {where_params:q}' if where_params is not None else t''
 
         _query = t"""
         SELECT
@@ -46,7 +44,7 @@ class AnalysisRunnerTable(DbBase):
             description, driver_image, config_path, cwd, environment,
             hail_version, batch_url, submitting_user, COALESCE(meta, {'{}'}::jsonb) as meta, output_path, audit_log_id
         FROM analysis_runner
-        WHERE {where_params:q}
+        {wheres_query:q}
         """
         async with self.connection.pg_connection.cursor(
             row_factory=class_row(AnalysisRunnerInternal)
--- a/db/python/tables/analysis_runner.py
+++ b/db/python/tables/analysis_runner.py
@@ -1,9 +1,11 @@
 import dataclasses
 import datetime
+from string.templatelib import Template
+
+from psycopg.types.json import Jsonb
 
 from db.python.filters import GenericFilter, GenericFilterModel
 from db.python.tables.base import DbBase
-from db.python.utils import to_db_json
 from models.models.analysis_runner import AnalysisRunnerInternal
 from models.models.project import ProjectId
 
@@ -25,8 +27,6 @@ class AnalysisRunnerTable(DbBase):
     Capture Analysis table operations and queries
     """
 
-    table_name = 'analysis_runner'
-
     async def query(
         self, filter_: AnalysisRunnerFilter
     ) -> list[AnalysisRunnerInternal]:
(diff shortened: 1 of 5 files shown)


db/python/tables/analysis_runner.py
```diff
@@ -25,68 +27,59 @@ class AnalysisRunnerTable(DbBase):
...
-FROM analysis_runner
-WHERE {where_str}
+        _query = t"""
+        SELECT
+            project, ar_guid, timestamp, access_level, repository, commit, script,
+            description, driver_image, config_path, cwd, environment,
+            hail_version, batch_url, submitting_user, meta, output_path, audit_log_id
+        FROM analysis_runner
+        WHERE {where_params:q}
     """
```
violetbrina: I guess this is a matter of preference but I'd indent this section or at least the ending """

db/python/tables/analysis_runner.py
```diff
@@ -25,68 +27,59 @@ class AnalysisRunnerTable(DbBase):
...
-        rows = await self.connection.fetch_all(_query, values)
-        return [AnalysisRunnerInternal.from_db(**dict(row)) for row in rows]
+        rows = await (await self.connection.pg_connection.execute(_query)).fetchall()
+
+        analysis_runner_internal_list = []
+        for row in rows:
+            row['meta'] = row.get('meta') or {}
+            analysis_runner_internal_list.append(AnalysisRunnerInternal(**row))
+
+        return analysis_runner_internal_list
```
violetbrina: If you cast the meta using `::jsonb` then you can use a row_factory. Example:

```python
async with self.connection.pg_connection.cursor(
    row_factory=class_row(FamilyInternal)
) as cur:
    await cur.execute(t'{joined_query:q} WHERE {where_params:q} GROUP BY f.id')
    family_internal_list = await cur.fetchall()
```

violetbrina (changes requested):
Two very small comments, but otherwise it looks good!

db/python/tables/analysis_runner.py
rameshka: Thanks @violetbrina, for pointing this out. I was trying to handle rows with `None` meta values after retrieving the rows. But it can be handled elegantly as you suggested. I have updated the query.

db/python/tables/analysis_runner.py
```diff
@@ -25,68 +28,57 @@ class AnalysisRunnerTable(DbBase):
...
-    project, ar_guid, timestamp, access_level, repository, commit, script,
-    description, driver_image, config_path, cwd, environment,
-    hail_version, batch_url, submitting_user, meta, output_path, audit_log_id
-FROM analysis_runner
-WHERE {where_str}
-    """
-        rows = await self.connection.fetch_all(_query, values)
-        return [AnalysisRunnerInternal.from_db(**dict(row)) for row in rows]
+        if where_params is None:
+            raise ValueError(f'Invalid filter: {filter_}')
```
dancoates: Not sure if this is correct? In general empty filters should be fine and not require raising an error

dancoates (commented):
Just one thing that I think needs fixing, but otherwise looks good

db/python/tables/analysis_runner.py
rameshka: hmm. I introduced this check to be consistent with the other queries. But as the original query did not check for empty filters, it proves your point @dancoates. I will remove this.

db/python/tables/analysis_runner.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1219/changes/8e8222dd039bd0b2972422accd5f843d91846543

db/python/tables/analysis_runner.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1219/changes/f76ac5162a3d881866caf8043d10fb43d3d6aa0d

dancoates (approved):

violetbrina (approved):
LGTM :)

PR #1220, opened 2026-02-19
(SET-986) Migrate project queries to template strings

violetbrina:
These postgres migration queries were done prior to the update to Python 3.14. This uplifts them to use the Template string feature we are now relying on in 3.14.


dancoates (approved):
Looks good!

rameshka (approved):
LGTM !!.

PR #1221, opened 2026-02-19
[SET-988] Migrate Enum table queries

rameshka:
This PR migrates the Enum table queries to PostgreSQL and adds a new test class to verify those queries.

### DIFF
--- a/db/python/enum_tables/enums.py
+++ b/db/python/enum_tables/enums.py
@@ -65,10 +65,7 @@ class EnumTable(DbBase):
         _query = t"""
             INSERT INTO {self._get_table_name():i} (id, name, audit_log_id)
             VALUES ({name_param}, {name_param}, {audit_log_id})
-            ON CONFLICT (id)
-            DO UPDATE SET
-            name = EXCLUDED.name,
-            audit_log_id = EXCLUDED.audit_log_id
+            ON CONFLICT (id) DO NOTHING
         """
 
         await self.connection.pg_connection.execute(_query)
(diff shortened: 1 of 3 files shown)


db/python/enum_tables/enums.py
```diff
@@ -58,15 +59,19 @@ async def insert(self, value: str):
...
-            VALUES (:name, :name, :audit_log_id)
-            ON DUPLICATE KEY UPDATE name = :name, audit_log_id = :audit_log_id
+        name_param = value.lower()
+        audit_log_id = await self.audit_log_id()
+
+        _query = t"""
+            INSERT INTO {self._get_table_name():i} (id, name, audit_log_id)
+            VALUES ({name_param}, {name_param}, {audit_log_id})
+            ON CONFLICT (id)
+            DO UPDATE SET
```
rameshka: Through Metamist, we cannot update the `name` of an enum entry (`id` == `name` for all entries inserted from the application), and an attempt to insert the same enum again will only update the audit entry.  As this does not change any values, we can have a `DO NOTHING` here. 

or else 

If there's a possibility of having enum entries with `id` != `name` (e.g., manual injection or another way), we can leave the above query as is (this will allow setting' id' and' name' to be equal).

dancoates: Yeah I think a DO NOTHING would make sense here - there's no point updating the audit log if nothing changes

dancoates (approved):
Looks good, I agree with your point on switching to `DO NOTHING`. Can you also please add the new enum tests to the pytest.toml so that the tests run on CI.

rameshka:
> Looks good, I agree with your point on switching to `DO NOTHING`. Can you also please add the new enum tests to the pytest.toml so that the tests run on CI.

Thanks for reviewing @dancoates .
Oh yeah, I have forgotten to add it. will do.

PR #1223, opened 2026-02-23
[PG] Migrate output file queries

hrrsheen:
Migrates the analysis output file queries to be compatible with postgres.
Migrates the corresponding tests to pytest.


db/python/tables/output_file.py
```diff
@@ -127,22 +123,14 @@ async def add_output_file_to_analysis(
...
-            {
-                'analysis_id': analysis_id,
-                'file_id': file_id,
-                'json_structure': json_structure,
-                'output': output,
-            },
-        )
+            VALUES
+                ({analysis_id}, {file_id}, {json_structure}, {output})
+            ON CONFLICT DO NOTHING"""
```
violetbrina, line 127: If someone tries to upload the same file meta twice, will this behaviour mean it silently does nothing? We might want to raise an error of some kind or if not, possibly a message to say that the output file already exists and was not updated.

test/test_analysis_output_files.py
```diff
@@ -384,22 +407,30 @@ async def test_project_deletion(self):
...
-        project = self.project_id_map[self.project_id]
-        self.assertTrue(await pttable.delete_project_data(project))
+        proj_permission_table = ProjectPermissionsTable(connection_with_project)
+        project = connection_with_project.project
+        assert await proj_permission_table.delete_project_data(project)
 
-        self.assertEqual(await self.row_count('analysis_outputs'), 0)
-        self.assertEqual(await self.row_count('output_file'), 0)
+        assert (await row_count('analysis_outputs')) == 0
+        assert (await row_count('output_file')) == 0
```
violetbrina, line 437: Maybe add another test after this for trying to insert the same output file meta twice?

violetbrina (changes requested):
Looking good. Just had one question and one suggestion for adding a test.

db/python/tables/output_file.py
hrrsheen: This was just replicating the behaviour from before the migration, but that's a good point.

db/python/tables/output_file.py
```diff
@@ -127,22 +123,14 @@ async def add_output_file_to_analysis(
         # The IGNORE is to avoid duplicate entries if the same file is added multiple times
         # and we used this over ON DUPLICATE because there are reported deadlocks with that
         # syntax in high concurrency situations?
```
dancoates: Can we remove this comment as it isn't relevant to postgres

db/python/tables/output_file.py
dancoates: I think this is ok - it's not so much that it does nothing as that the desired outcome has already been reached. @nevoodoo do you mind having a quick look to confirm?

dancoates (approved):
This looks good to me, if we can just resolve the couple of pending comments then good to merge

db/python/tables/output_file.py
dancoates: hm, I tried to add a comment to this thread but it ended up being posted separately:

I think this is ok - it's not so much that it does nothing as that the desired outcome has already been reached. @nevoodoo do you mind having a quick look to confirm?


now it _is_ displaying above, who knows 🤷

nevoodoo: This is actually a bug I think? Like I just checked the prod DB and I've seen duplicates being added because of the behavior above. I need to add unique constraints to the table so we can prevent this and then also fix this. @hrrsheen is it okay if I pop a PR in tomorrow to fix this?

thank you for tagging me @dancoates I would not have discovered this bug otherwise

dancoates: There's already unique constraints on the table:


```sql
UNIQUE KEY `unique_analysis_file` (`analysis_id`,`file_id`,`json_structure`),
  UNIQUE KEY `unique_analysis_output` (`analysis_id`,`output`,`json_structure`),
CONSTRAINT `chk_file_id_output` CHECK (`file_id` is not null and `output` is null or `file_id` is null and `output` is not null)
```

which I think should force uniqueness?

dancoates: though if json_structure is null you could create duplicates

test/test_analysis_output_files.py
hrrsheen: @dancoates even though we want to continue to silently ignore duplicate output files being added to an analysis, would it be worth adding a test to codify that requirement in the event of any future database changes?

dancoates: There's no harm in more tests!

hrrsheen: Cool beans, I've added the test.

db/python/tables/output_file.py
nevoodoo: Yep, its all because of null `json_structure` values. I thought I had the constraints previously but didn't see it in the migration file so I figured I missed it. I will pop in a PR to improve that unique constraint that we already have so it handles the null `json_structure` too. I will also clean up the DELETE behaviour which is currently a bit messy.

nevoodoo: https://github.com/populationgenomics/metamist/pull/1232

hrrsheen: I'm a bit confused as to how these duplicates came about in the first place.
Looking at the code for `create_or_update_analysis_output_files_from_output`, I can't think of a scenario where a `None` value would be passed to `json_structure`.

nevoodoo: Based on what I see, the only way is:

A user uses the old `output` field, a file would be created and a record added to `analysis_outputs` but then because it wasn't created with a json dict, the `json_structure` field would be NULL, and this can be repeated, causing duplicates.

violetbrina (approved):

PR #1224, opened 2026-02-23
[SET-989] Migrate project insights queries

rameshka:
This PR migrate project insights queries to use PostgreSQL syntax.
There aren't many test cases for this work. But the existing two testcases invoke almost all the DB functions.
In addition, all the new queries were manually tested.

### DIFF
--- a/db/python/layers/project_insights.py
+++ b/db/python/layers/project_insights.py
@@ -6,6 +6,7 @@ from datetime import datetime
 from typing import Any, NamedTuple
 
 from databases.interfaces import Record
+from psycopg import sql
 
 from db.python.enum_tables import SequencingPlatformTable as SeqPlatformTable
 from db.python.enum_tables import SequencingTechnologyTable as SeqTechTable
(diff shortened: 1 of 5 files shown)


db/python/layers/project_insights.py
```diff
@@ -815,38 +798,36 @@ async def _details_stripy_reports(
...
-) latest_analysis ON a.id = latest_analysis.max_analysis_id;
+        _query = t"""
+        SELECT
+            a.project,
+            a.id,
+            coalesce(a.output, ao.output, of.path) as output,
+            a.timestamp_completed,
+            asg.sequencing_group_id,
+            a.meta -> 'outliers_detected' as outliers_detected,
+            a.meta -> 'outlier_loci' as outlier_loci
```
dancoates, line 806: In `get_sg_web_report_links` this `outlier_loci` field is run through `json.loads` but it will already be a dict as this `->` returns jsonb which psycopg automatically converts

dancoates (commented):
This looks good, just one small bug I found while testing

db/python/layers/project_insights.py
rameshka: Oh, thanks for catching that. I will update `get_sg_web_report_links` function

rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1224/commits/94128fac08ef8714554b9e8a1a2b4f2f58fbe54a

dancoates (approved):

rameshka:
Thanks, @dancoates, for reviewing.

PR #1225, opened 2026-02-23
Add test project name markers.

hrrsheen:
Adds the ability to set the name of the test project that the `connection_with_project` fixture uses.
This can be done using the `@pytest.mark.project_name(<name>)` decorator.


dancoates (approved):
Nice! very handy addition!

PR #1226, opened 2026-02-24
Fix sg assay queries

hrrsheen:
Adds a check for None values to the `SequencingGroupAssayFilter` when querying graphql for sequencing groups.


violetbrina (changes requested):
Actually, could you add a test to the `test_assays.py` to cover this error? Just so we don't miss it again. At the very least the one I had would be good. Thank you!

violetbrina (approved):
LGTM!

PR #1228, opened 2026-02-25
[PG] Migrate web queries

hrrsheen:
Migrates the web questies to postgres and migrates the web tests to pytest.
Additionally identified and fixed some small query bugs in `tables/assay.py`, `tables/participant.py` and `tables/sample.py`.

### DIFF
--- a/db/python/tables/sample.py
+++ b/db/python/tables/sample.py
@@ -659,13 +659,12 @@ class SampleTable(DbBase):
         if len(sample_ids) == 0:
             return {}
         _query = t"""
-            SELECT id, MIN(lower(s.sys_period)) as date_created
-            FROM sample
-            INNER JOIN (
+            SELECT id, MIN(lower(sys_period)) as date_created
+            FROM (
                 SELECT id, sys_period FROM sample
                 UNION ALL
                 SELECT id, sys_period FROM sample_history
-            ) sample_hist ON sample.id = sample_hist.id
+            ) AS sample_hist
             WHERE id = ANY({sample_ids})
             GROUP BY id
         """
--- a/db/python/layers/web.py
+++ b/db/python/layers/web.py
@@ -139,9 +139,9 @@ class WebDb(DbBase):
         )
         seqr_links = self.get_seqr_links_from_project(project)
 
-        atable = AnalysisTable(self._connection)
-        seqtable = AssayTable(self._connection)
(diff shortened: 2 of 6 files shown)


db/python/tables/sample.py
```diff
@@ -659,13 +659,12 @@ async def get_samples_create_date(self, sample_ids: list[int]) -> dict[int, date
...
-            FROM sample
-            INNER JOIN (
+            SELECT id, MIN(lower(sys_period)) as date_created
+            FROM (
                 SELECT id, sys_period FROM sample
                 UNION ALL
                 SELECT id, sys_period FROM sample_history
-            ) sample_hist ON sample.id = sample_hist.id
+            ) AS sample_hist
             WHERE id = ANY({sample_ids})
```
rameshka, line 668: Minor optimisation: can we push down the filtering on sample_ids to the sub query rather than union on all and filter later?
something like:
```
SELECT id, MIN(lower(sys_period)) as date_created
FROM (
    SELECT id, sys_period FROM sample
    WHERE id = ANY({sample_ids})
    
   UNION ALL
    
    SELECT id, sys_period FROM sample_history
    WHERE id = ANY({sample_ids})
) AS sample_hist
GROUP BY id
```

rameshka (commented):
@hrrsheen changes are looking good. I just left one small comment.

db/python/tables/sample.py
dancoates: This is an interesting one - postgres does predicate pushdown so the filtering by id will be pushed into the sub queries.

This is the query plan for the query with the id filter on the outside of the subqueries:

https://explain.dalibo.com/plan/333da0a26c293h7b

If you expand out the bottom two parallel scans and look in the `misc` tab you'll see that the filtering is done at that stage. 

image


The query plan of including the id filter twice in the subqueries is very similar, though does eliminate one aggregation step - most of the time is spent in the seq scan of the history table so the difference should be negligible 

https://explain.dalibo.com/plan/heeg6dfgdbg21581

I'm happy to go with either, with a slight preference for keeping the single id filter on the outside as it will mean sending less data to the db server.

dancoates (approved):
This is looking good to me

db/python/tables/sample.py
rameshka: this is interesting. No objection from my end to leave the query as it is.

rameshka (approved):

PR #1229, opened 2026-02-25
[SET-1011] Revisit family and family participant test cases

rameshka:
This PR unskips the test cases in:

- test/test_external_id.py
- test/test_graphql.py
- test/test_pedigree.py
- test/test_search.py
- test/test_update_participant_family.py

The test cases in these files were skipped until all the dependent entity queries are migrated to PostgreSQL syntax. 
While unskipping them, some test cases were failing and required some minor fixes.

**There are still some test cases skipped until _querying JSON keys_ and _case insensitive string matching_ work.**

In addition, the test cases in test/test_cohort_status.py file are updated to reuse the `connection/db` created in the test class `set_up` fixture instead of creating a new test DB both in the set_up and again in the testcase.

### DIFF
--- a/db/python/tables/participant.py
+++ b/db/python/tables/participant.py
@@ -374,6 +374,7 @@ class ParticipantTable:
                     'audit_log_id': audit_log_id,
                 }
                 for name, external_id in external_ids.items()
+                if external_id is not None
             ]
 
             async with self.connection.pg_connection.cursor() as cur:
@@ -442,17 +443,16 @@ class ParticipantTable:
                     ]
 
                     # Batch update
-                    cur = conn.cursor()
                     async with conn.cursor() as cur:
                         await cur.executemany(
                             """
                                 MERGE INTO participant_external_id AS target
                                 USING (VALUES (%(project)s, %(participant_id)s, %(name)s, %(external_id)s, %(audit_log_id)s))
                                     AS source (project, participant_id, name, external_id, audit_log_id)
-                                ON target.project = source.project
+                                ON target.name = source.name
                                     AND target.participant_id = source.participant_id
                                 WHEN MATCHED THEN
-                                    UPDATE SET name = source.name, external_id = source.external_id, audit_log_id = source.audit_log_id
+                                    UPDATE SET external_id = source.external_id, audit_log_id = source.audit_log_id
                                 WHEN NOT MATCHED THEN
                                     INSERT (project, participant_id, name, external_id, audit_log_id)
                                     VALUES (source.project, source.participant_id, source.name, source.external_id, source.audit_log_id)
@@ -551,14 +551,14 @@ class ParticipantTable:
     ) -> tuple[set[ProjectId], dict[int, list[ParticipantInternal]]]:
         """Get list of participants keyed by families, duplicates results"""
 
-        keys_str = sql.SQL(', ').join(self.keys)
         _query = t"""
-            SELECT fp.family_id, {keys_str:q}
+            SELECT fp.family_id, p.id, jsonb_object_agg(peid.name, peid.external_id) as external_ids,
+            p.reported_sex, p.reported_gender, p.karyotype, p.meta, p.project, p.audit_log_id
             FROM participant p
             INNER JOIN family_participant fp ON fp.participant_id = p.id
             INNER JOIN participant_external_id peid ON p.id = peid.participant_id
             WHERE fp.family_id = ANY({family_ids})
-            GROUP BY p.id
+            GROUP BY p.id, fp.family_id
         """
         conn = self.connection.pg_connection
         cur = await conn.execute(_query)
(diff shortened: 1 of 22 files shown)


db/python/tables/participant.py
```diff
@@ -374,6 +374,7 @@ async def create_participant(
                     'audit_log_id': audit_log_id,
                 }
                 for name, external_id in external_ids.items()
+                if external_id is not None
```
rameshka, line 382: encountered a failing test case with a `None` `external_id`. So included `None` check - this check is also in the previous version (before migration).

db/python/tables/participant.py
```diff
@@ -551,14 +551,14 @@ async def get_participants_by_families(
     ) -> tuple[set[ProjectId], dict[int, list[ParticipantInternal]]]:
         """Get list of participants keyed by families, duplicates results"""
 
-        keys_str = sql.SQL(', ').join(self.keys)
         _query = t"""
-            SELECT fp.family_id, {keys_str:q}
```
rameshka, line 561: psycopg was not interpreting the columns specified in `{keys_str:q}`  correctly (when debugging, I could see it was only able to identify `fp.family_id` and `external_ids`).

db/python/tables/participant.py
```diff
@@ -551,14 +551,14 @@ async def get_participants_by_families(
...
+            SELECT fp.family_id, p.id, jsonb_object_agg(peid.name, peid.external_id) as external_ids,
+            p.reported_sex, p.reported_gender, p.karyotype, p.meta, p.project, p.audit_log_id
             FROM participant p
             INNER JOIN family_participant fp ON fp.participant_id = p.id
             INNER JOIN participant_external_id peid ON p.id = peid.participant_id
             WHERE fp.family_id = ANY({family_ids})
-            GROUP BY p.id
+            GROUP BY p.id, fp.family_id
```
rameshka, line 566: Fixes `psycopg.errors.GroupingError`

db/python/tables/participant.py
```diff
@@ -630,9 +635,9 @@ async def get_external_participant_id_to_internal_sequencing_group_id_map(
         participants with multiple sequencing groups.
         """
         wheres = [t'p.project = {project}']
-        wheres.append(t'sg.type = {sequencing_type}' if sequencing_type else t'')
+        if sequencing_type:
+            wheres.append(t'sg.type = {sequencing_type}')
 
-        wheres = [w for w in wheres if w]
```
rameshka, line 640: For `None` sequencing_type scenarios, the empty Template string did not get dropped.

db/python/tables/sample.py
```diff
@@ -707,7 +707,7 @@ async def get_history_of_sample(self, id_: int):
                 FROM history.sample_history
                 WHERE id = {id_}
             )
-            ORDER BY lower(sys_period) DESC
+            ORDER BY lower(sys_period)
```
rameshka, line 709: One of the existing test cases was failing, as the rows are now returned in reverse chronological order. But with this update, a new test is failing (`TestSample.test_get_history_of_sample`).

db/python/tables/participant.py
```diff
@@ -442,17 +443,16 @@ async def update_participant(
...
                     # Batch update
-                    cur = conn.cursor()
                     async with conn.cursor() as cur:
                         await cur.executemany(
                             """
                                 MERGE INTO participant_external_id AS target
                                 USING (VALUES (%(project)s, %(participant_id)s, %(name)s, %(external_id)s, %(audit_log_id)s))
                                     AS source (project, participant_id, name, external_id, audit_log_id)
-                                ON target.project = source.project
+                                ON target.name = source.name
```
rameshka, line 457: encountered a  `psycopg.errors.UniqueViolation` for one of the test cases. Therefore, updated the query to use the primary key columns of `participant_external_id` table.

db/python/tables/participant.py
dancoates: Could you please update the type of external_ids to make it clear that it can contain `None`s?

dancoates (approved):
One tiny thing, but otherwise looks great. Nice to see so many unskipped tests with minimal fixes required!

db/python/tables/sample.py
rameshka: @dancoates shall I update the new test case (now failing) then, retaining the original behaviour?

dancoates: yeah that's fine - I don't think the order matters too much here, though I don't really understand how they worked pre-migration if two tests expected different sort orders?

rameshka: I think it's a new test case. I could not find it in the `dev` branch.

db/python/tables/participant.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1229/commits/fd4af8f93884f6eb40faad070116037c58ad94c1

db/python/tables/sample.py
dancoates: ah gotcha, definitely fine to update then

rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1229/commits/995931e7266798528fb260517d257073ab4ed4e8

PR #1230, opened 2026-02-25
Tweaks to generic filter dict querying to reduce duplication

dancoates:


violetbrina (approved):
That's wildly clever. I had no idea how to get that json_path query inline working with the exiting to_sql function. The join with the dummy row is wild. Would have never thought of it!

PR #1231, opened 2026-02-26
[SET-1020] Revisit TestComment test class 

rameshka:
This PR migrate `comment` query-related test cases.

### DIFF
--- a/models/models/project.py
+++ b/models/models/project.py
@@ -43,7 +43,7 @@ class Project(SMBase):
     @property
     def is_test_project(self):
         """Returns whether this is a main or a test project"""
-        return self.name.endswith('-test')
+        return self.name.startswith('test-')
 
 
 class ProjectMemberUpdate(SMBase):
--- a/models/models/project.py
+++ b/models/models/project.py
@@ -43,7 +43,7 @@ class Project(SMBase):
     @property
     def is_test_project(self):
         """Returns whether this is a main or a test project"""
-        return self.name.startswith('test-')
+        return self.name.endswith('-test')
 
 
 class ProjectMemberUpdate(SMBase):
--- a/test/test_comment.py
+++ b/test/test_comment.py
@@ -512,17 +512,58 @@ class CommentHelpers:
         return restore_comment_result['data']['comment']['restoreComment']
 
 
+def get_participant_to_insert(id_suffix='1'):
+    """Helper function to create a participant object for insertion into the database"""
+    return ParticipantUpsertInternal(
+        external_ids={PRIMARY_EXTERNAL_ORG: 'P0' + id_suffix},
+        meta={'pmeta': 'pvalue'},
+        reported_sex=2,
+        reported_gender='FEMALE',
+        karyotype='XX',
+        samples=[
+            SampleUpsertInternal(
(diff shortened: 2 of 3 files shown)


models/models/project.py
```diff
@@ -43,7 +43,7 @@ class Project(SMBase):
     @property
     def is_test_project(self):
         """Returns whether this is a main or a test project"""
-        return self.name.endswith('-test')
+        return self.name.startswith('test-')
```
rameshka: One of the test cases invoking the function [`project.delete_project_data`](https://github.com/populationgenomics/metamist/blob/c829422d6ba3ab5860fd909d853eac366985cc02/db/python/tables/project.py#L210) was failing as the new project name is `test-project`.
This function only allows executing `DELETE` queries for a test project (verified from this function). 
In the pre-migration logic [TEST_PROJECT_NAME = 'test-test'](https://github.com/populationgenomics/metamist/blob/28164f5930776875f918386d14fb4d1e198bb477/test/testbase.py#L30).  So `endswith` check works. 

We can simply use `startswith` check -> as long as the prefix `test-` is only used for test projects.  
or
Rename the new `test-project` (has to update multiple places in the code)

test/test_comment.py
```diff
@@ -512,17 +512,58 @@ async def restore_comment(self, id_: int) -> dict[str, Any]:
         return restore_comment_result['data']['comment']['restoreComment']
 
 
+def get_participant_to_insert(id_suffix='1'):
```
rameshka: Extracted `get_participant_to_insert` function in the `test_participant.py` file here. The `test_participant.py` has `samples` section commented out, resulting in a failing test here.

models/models/project.py
rameshka: One of the test cases `TestProjectAccess.test_delete_project_data` is failing due to this change

dancoates: This is a bit confusing but "test project" here is not referring to a project used for unittests, but for a project used by analysts for testing metamist scripts, so it should stay with the check for the test suffix.

test/test_comment.py
dancoates: Can we uncomment the samples section in `test_participant`?

models/models/project.py
dancoates: @hrrsheen added a parameterisation to the fixture that creates the project for testing, so you can change the project name to allow testing things that relate to test projects like project deletion. 

https://github.com/populationgenomics/metamist/pull/1225

test/test_comment.py
rameshka: yes. I tried uncommenting and can see all the tests are passing.

models/models/project.py
rameshka: This is cool. thanks @dancoates

dancoates (approved):
Nice!

models/models/project.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1231/changes/c34b31785b54d6689b1817aad8e80dcfaecbc01e

test/test_comment.py
rameshka: addressed in https://github.com/populationgenomics/metamist/pull/1231/changes/c34b31785b54d6689b1817aad8e80dcfaecbc01e

PR #1232, opened 2026-02-26
fixed analysis outputs duplication

nevoodoo:
## Problem
The analysis_outputs table had no unique constraints handling NULL values for `json_structure`, which caused two bugs:

`ON CONFLICT DO NOTHING` was a no-op — Since there was no unique constraint to conflict on, every `INSERT` always succeeded. Calling update multiple times on the same analysis created duplicate rows.

Stale rows persisted on key renames/moves — The cleanup `DELETE` logic only checked whether a `file_id` was still referenced, not whether its `json_structure` path had changed. Renaming a key (e.g. "vcf" → "variant_calls") or moving a file between primary/secondary left ghost rows with the old `json_structure`.

### Schema
- Added UNIQUE partial index on `(analysis_id, file_id)` where `file_id IS NOT NULL`
- Added UNIQUE partial index on `(analysis_id, output)` where `output IS NOT NULL`

### Application Logic
- Replaced the insert-then-selective-delete approach with delete-all-then-insert within the existing transaction
- The old approach built up `file_ids` and `outputs` lists, then ran a complex `DELETE ... WHERE file_id <> ALL(...)` query that couldn't detect `json_structure` changes
- The new approach simply deletes all `analysis_outputs` rows for the analysis first, then re-inserts - this is atomic (wrapped in a transaction), idempotent, and correctly handles all edge cases

Scenario | Before | After
-- | -- | --
Same update called twice | Duplicate rows | Idempotent ✅
Rename a JSON key | Old + new rows coexist | Only new row ✅
Move primary → secondary | Old row persists | Clean update ✅
Move secondary → primary | Old row persists | Clean update ✅
Remove a file | ✅ Already worked | ✅ Still works
Add a file | Duplicates on retry | Idempotent ✅

### DIFF
--- a/db/migrations/20260120061420_create_tables.sql
+++ b/db/migrations/20260120061420_create_tables.sql
@@ -52,14 +52,6 @@ CREATE TABLE analysis_outputs (
     )
 );
 
-CREATE UNIQUE INDEX uq_analysis_outputs_file
-    ON analysis_outputs (analysis_id, file_id)
-    WHERE file_id IS NOT NULL;
-
-CREATE UNIQUE INDEX uq_analysis_outputs_output
-    ON analysis_outputs (analysis_id, output)
-    WHERE output IS NOT NULL;
-
 
 
 -- analysis_runner table
--- a/db/migrations/20260120061420_create_tables.sql
+++ b/db/migrations/20260120061420_create_tables.sql
@@ -52,6 +52,14 @@ CREATE TABLE analysis_outputs (
     )
 );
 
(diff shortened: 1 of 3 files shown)


violetbrina (commented):
Harrison and I were discussing this on his PR. I was going to open a ticket in metamist for this but if you've gone and fixed it then fantastic! Although it might be worth making the ticket anyway to add to the postgres-migration epic. Thanks for the fix!

dancoates (approved):

PR #1234, opened 2026-02-27
Set 1003 generic filter text case insensitive comparison

violetbrina:
As described in the ticket for SET-1003, Mariadb was automatically doing very flexible, case ignoring string comparisons. 

This update to the GenericFilter makes all string comparisons `ILIKE` or `CASEFOLD(value)` to replicate that pre-exiting functionality after the move to Postgres.

The tests have been updated accordingly.

### DIFF
--- a/db/python/filters/generic.py
+++ b/db/python/filters/generic.py
@@ -124,7 +124,7 @@ class GenericFilter[T](SMBase):
         # CASEFOLD(value) and the column query is also wrapped in CASEFOLD
         original_gf: GenericFilter[T] = self
         gf: GenericFilter[T] = self.copy_with_casefolded_strings()
-        value_type = infer_pg_type_from_filter(self)
+        value_type = self.infer_pg_type_from_filter()
 
         original_column_query = (
             column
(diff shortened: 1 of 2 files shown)


db/python/filters/generic.py
```diff
@@ -331,15 +382,19 @@ def infer_pg_type_from_filter(value: GenericFilter[Any]) -> str:
             cast_types.add('bool')
         elif isinstance(v, (int, float)):
             cast_types.add('numeric')
-        else:
+        elif isinstance(v, Enum):
+            cast_types.add('enum')
```
dancoates: Can we add some handling to `prepare_query_from_dict_field` for this as the `returning {pg_type:i}` won't work with `enum`. I think it'd be fine to raise an error if the pg type isn't numeric, bool, or text as we wouldn't expect anything other than those for dict filters.

db/python/filters/generic.py
```diff
@@ -297,7 +346,7 @@ def to_sql(
         return sql.SQL(' AND ').join(filters_rm_none)
 
 
-def infer_pg_type_from_filter(value: GenericFilter[Any]) -> str:
+def infer_pg_type_from_filter(value: GenericFilter[Any]) -> str | None:
```
dancoates: Should we move this to a method on the `GenericFilter` class?

dancoates (commented):
Thanks for this, the updates to the tests show that the case insensitive comparisons are working correctly. I do think there are some changes that we can make to the implementation to ensure best practices and scalability though.

The main things are trying to remove the use of `sql.Literal` and avoiding constructing the big `ARRAY[` strings with each array value wrapped individually in `CASEFOLD`. 

`sql.Literal` should only be used if there is no other alternative as it means the value is sent to the db server in the query string, rather than as a parameter. And `_in` queries are often used with 1000s of values so constructing arrays with the values wrapped in CASEFOLD will create some huge query strings. 

I'd recommend going with an approach where we add some small methods to the top of `to_sql` that allow wrapping values and lists conditionally in the `CASEFOLD` function. Something like:

```python
        column_query = (
            column
            if isinstance(column, Template)
            else t'{sql.Identifier(*column.split(".")):i}'
        )

        pg_type = infer_pg_type_from_filter(self)

        if pg_type == 'text':
            column_query = t'CASEFOLD({column_query:q})'

        def wrap_val(val: T | str):
            if pg_type == 'text':
                return t'CASEFOLD({val})'
            return t'{val}'

        def wrap_val_list(val_list: list[T]):
            if pg_type == 'text':
                return t'(SELECT CASEFOLD(val) FROM unnest({val_list}::text[]) AS val)'
            return t'{val_list}'
```

That also includes an alternative to wrapping the value list individually in casefolds, using a subquery with unnest which still allows the value list to be sent to the db as a param.

That way only minimal changes are required to the existing query building code, and it doesn't require making a copy of the filter. 


```python

        # for single val queries
        if self.eq is not None:
            filters.append(t'{column_query:q} = {wrap_val(self.eq):q}')

        # For list val queries
        if self.in_ is not None:
            if not isinstance(self.in_, list):
                raise ValueError('IN filter must be a list')

            # in an empty list is always false
            if len(self.in_) == 0:
                return t'FALSE'

            filters.append(t'{column_query:q} = ANY({wrap_val_list(self.in_):q})')

```

db/python/filters/generic.py
```diff
@@ -134,17 +157,23 @@ def to_sql(self, column: str | Template) -> Template | None:
             if len(self.in_) == 0:
                 return t'FALSE'
 
-            filters.append(t'{column_query:q} = ANY({self.in_})')
+            # Converting the list a query string will work for Any type
+            # in GenericFilter.in_ including a sql.Composed value
```
dancoates: This might be me, but my brain can't quite parse this comment

db/python/filters/generic.py
```diff
@@ -179,20 +210,70 @@ def transform(self, func: Callable[[T], X]) -> GenericFilter[X]:
         Apply a function to each value in the filter
         """
         return GenericFilter(
-            eq=func(self.eq) if self.eq else None,
-            neq=func(self.neq) if self.neq else None,
+            eq=func(self.eq) if self.eq is not None else None,
```
dancoates: These changes to `is not None` aren't needed for this PR I think? It's a minor change but is functionally different so could potentially break something in an edge case. I think would be good to leave as it was before.

db/python/filters/generic.py
```diff
@@ -154,14 +183,16 @@ def to_sql(self, column: str | Template) -> Template | None:
...
         if self.icontains is not None:
-            search_term = escape_like_term(str(self.icontains))
-            filters.append(t"{column_query:q} ILIKE '%' || {search_term} || '%'")
+            search_term = wrap_val(escape_like_term(str(self.icontains)))
+            filters.append(t"{column_query:q} ILIKE '%' || {search_term:q} || '%'")
         if self.startswith is not None:
-            search_term = escape_like_term(str(self.startswith))
-            filters.append(t"{column_query:q} LIKE {search_term} || '%'")
+            search_term = wrap_val(escape_like_term(str(self.startswith)))
+            filters.append(t"{column_query:q} ILIKE {search_term:q} || '%'")
```
dancoates: Should all these queries that use ILIKE be using `original_column_query` like the `contains` one?

dancoates (approved):
Looks good, just a few very minor things and then good to go

PR #1235, opened 2026-02-27
SET-1017 - Metamist postgres deployment

dancoates:
- Add pulumi infrastructure config for metamist deploy
- Add new migration to create necessary roles, this uses a superuser role to create the server role and dev access roles. There is a certain amount of bootstrapping involved in the CloudSQL databases to allow for this to work. See this doc for info: https://cpg-populationanalysis.atlassian.net/wiki/spaces/ST/pages/975765522/Metamist+Postgres+Database+-+documentation+and+setup
- Update tests to correctly handle new roles and permissions
- Update package building to point to new production URLs
- Add new database migration script and cloudrun service that uses dbmate to run migrations and can be triggered via github actions. 
- Switch asgi framwork to hypercorn so that the cloudrun can be http2 and handle larger response sizes.
- Remove up old mariadb backup and migration code


deploy/infra/server.py
```diff
@@ -0,0 +1,313 @@
...
+            args={'BUILDKIT_INLINE_CACHE': '1'},
+            builder_version=BuilderVersion.BUILDER_BUILD_KIT,
+            platform='linux/amd64',
+        ),
+        opts=pulumi.ResourceOptions(depends_on=[image_repository]),
+    )
+
+    # Service account for API
+    service_account = gcp.serviceaccount.Account(
+        'metamist-service-account',
```
ckrishk, line 43: good idea to make this a configurable value too

deploy/Pulumi.production.yaml
```diff
@@ -0,0 +1,41 @@
+secretsprovider: gcpkms://projects/metamist-488300/locations/australia-southeast1/keyRings/metamist_pulumi_secrets_keyring/cryptoKeys/metamist_pulumi_secrets
+encryptedkey: CiQAelOKiUKO55T9/L58gQl4yuSCU6VRJMcuNAh377Foxf1qJvwSSQDdXNuFPODn1KcT+V++wl3SjTi181KXWF1fggSVGR7FS/NRHHmQrq58ggRKDoWjWxI5nUMiy5yDILQWyOmfOikK/L25EdznGSQ=
```
ckrishk: we will not want this key to be part of `.yaml` file

ckrishk (commented):
Looks great, have a couple of comments.

EddieLF:
Hey Dan, will moving from MySQL / MariaDB to Postgres break any existing SQL queries or statements found across the methods in the classes from [`/db/python/tables/`](https://github.com/populationgenomics/metamist/tree/dev/db/python/tables) or in the [`project_insights`](https://github.com/populationgenomics/metamist/blob/dev/db/python/layers/project_insights.py) layer? (Or anywhere else SQL code is in the codebase.)

I know there are some minor syntax differences in Postgres but I'm not across all of them. 

Some to flag might be `GROUP_CONCAT` which is used by the project Insights dashboard [here](https://github.com/populationgenomics/metamist/blob/dev/db/python/layers/project_insights.py#L548) and [here](https://github.com/populationgenomics/metamist/blob/dev/db/python/layers/project_insights.py#L188). Seems like it might be used a few other times across the `tables` classes too. There are also some `... ON DUPLICATE KEY ...` statements in the methods from the `tables` classes.

deploy/Pulumi.production.yaml
dancoates: This is the _encrypted_ key that is used to encrypt secrets, the key needs to be decrypted by kms before it can be used. Pulumi recommends storing this in version control and it is ok for it to be public:
https://www.pulumi.com/docs/iac/concepts/secrets/#committing-configuration-to-source-control

deploy/infra/server.py
dancoates: This line is just the resource name in pulumi, the line below is the service account name which includes the stack name so should be unique for dev/production.

dancoates:
> Hey Dan, will moving from MySQL / MariaDB to Postgres break any existing SQL queries or statements found across the methods in the classes from [`/db/python/tables/`](https://github.com/populationgenomics/metamist/tree/dev/db/python/tables) or in the [`project_insights`](https://github.com/populationgenomics/metamist/blob/dev/db/python/layers/project_insights.py) layer? (Or anywhere else SQL code is in the codebase.)
> 
> I know there are some minor syntax differences in Postgres but I'm not across all of them.
> 
> Some to flag might be `GROUP_CONCAT` which is used by the project Insights dashboard [here](https://github.com/populationgenomics/metamist/blob/dev/db/python/layers/project_insights.py#L548) and [here](https://github.com/populationgenomics/metamist/blob/dev/db/python/layers/project_insights.py#L188). Seems like it might be used a few other times across the `tables` classes too. There are also some `... ON DUPLICATE KEY ...` statements in the methods from the `tables` classes.

Hey Ed, thanks for taking a look - I didn't realise the genomic analysis team got tagged on this PR. We've done a (hopefully) comprehensive migration of all existing queries across to postgres syntax, including all the project insights ones. There's a test postgres deploy at https://metamist.populationgenomics.org.au/ if you wanted to check if project insights/summary is working as expected (we're still working on this so it might break as we redeploy things for testing).

deploy/Pulumi.production.yaml
ckrishk: Interesting. TIL . Thanks

ckrishk (approved):
Looks good

PR #1236, opened 2026-02-27
[PG] Revisit sequencing group tests

hrrsheen:
Fixes `sequencing_group` tests that were previously skipped due to missing functionality.
Also replaces fixture that relied on manually inserting into the database.


dancoates (approved):
Looks good!

PR #1237, opened 2026-03-02
SET-1035 Set up and integrate pyright

dancoates:
WIP


db/python/layers/family.py
```diff
@@ -113,9 +113,9 @@ async def get_families_by_participants(
     async def update_family(
         self,
         id_: int,
-        external_ids: dict[str, str] | None = None,
-        description: str = None,
-        coded_phenotype: str = None,
+        external_ids: dict[str, str | None] | None = None,
```
milo-hyben: Any particular reason we need dict[None] and not just None?

pyproject.toml
```diff
@@ -81,3 +82,20 @@ omit = [
...
+    "packages/metamist/src/metamist/model/*",
+    "packages/metamist/src/metamist/models/*",
+    "packages/metamist/src/metamist/*.py",
+]
+
+pythonVersion = "3.14"
+pythonPlatform = "Linux"
+
+executionEnvironments = [
+    { root = "packages/metamist", pythonVersion = "3.10" },
```
milo-hyben, line 107: on line 96 pythonVersion=3.14, here it is 3.10.
Should we make this consistent?

milo-hyben (approved):
I have couple of small things, otherwise it looks good to me!

jmarshall:
I've collected your comments here so they don't go missing next time I force-push the commits on this PR:

----

> `external_ids: dict[str, str | None] | None = None,`
>
> Any particular reason we need dict[None] and not just None?

It's not super well documented, but for updates and upserts, `external_ids = {'foo': 'id123', 'bar': None}` means to update the value for `foo` and delete the entry for `bar`. So the value type of the dict **for upserts** is indeed `str | None`.

For the basic models, the dict is `external_ids: dict[str, str]` as you'd expect.

----

> `executionEnvironments = [`
> `    { root = "packages/metamist", pythonVersion = "3.10" },`
>
> on line 96 pythonVersion=3.14, here it is 3.10.
> Should we make this consistent?

It deserves a comment. The bulk of metamist code is 3.14 but this overrides it for the _packages/metamist_ subtree. This is because we now build the metamist server with 3.14 (for template strings) so check most of the code with that, but the API needs to remain useable with older Pythons.

pyproject.toml
jmarshall: Explanatory comment now added.

jmarshall:
This monster is now fully ready for review! 🎉 

The changes mostly fill into one of three categories:

1. Purely type annotation changes, which cannot affect runtime behaviour (except perhaps to the extent that strawberry et al introspect types! but the strawberry-related types have almost entirely been left alone).
2. Additions of `assert`. If this changes runtime behaviour, it will generally be to crash tidily instead of messily when something has already gone wrong…
3. Actual code changes. These are mostly small, and mostly called out with CODE CHANGE in the commit message.

api/graphql/loaders.py
```diff
@@ -385,7 +385,7 @@ async def load_participants_for_ids(
         missing_pids = set(participant_ids) - set(p_by_id.keys())
         if missing_pids:
             raise NotFoundError(f'Could not find participants with ids {missing_pids}')
-        return [p_by_id.get(p) for p in participant_ids]
+        return [p_by_id[p] for p in participant_ids]
```
rameshka, line 388: Here, the returned list is populated based on `participant_ids`. So  I'm not sure if the returned list is expected to have `None` placeholders for the `None` values in the `participant_ids` list, and also for those participant IDs that could not fetch their `ParticipantInternal` objects from the DB and not in the  `persons` list (Invoking `get_participants_by_ids`).

jmarshall:
There are still a number of type errors in a few files:

**api/graphql/schema.py** 
Numerous strawberry-related type diagnostics that would be invasive and time-consuming to resolve.

**api/utils/db.py**
The functionality in question is unused, so it is unclear what the policy for this function should be.

**db/python/layers/participant.py**
Unclear whether `family_participant.paternal_id` et al can be None or not.

~~**db/python/layers/seqr.py** 
`get_participant_ids_sequencing_group_ids_for_sequencing_type()` was apparently wrongly removed in PR #1204. PR #1269 puts it back.~~ Now fixed on **pg-dev**.

**db/python/layers/sequencing_group.py**
This is [SET-1103](https://cpg-populationanalysis.atlassian.net/browse/SET-1103).

**db/python/layers/project_insights.py** 
**packages/metamist/src/metamist/audit/…** 
Audit code that is not part of the metamist core.

**models/models/sequencing_group.py** 
Internal and external classes disagree as to whether these fields can be None.

**packages/metamist/src/metamist/parser/…**
Parser code is a self-contained part of the metamist code that seems relatively fragile.

**test/data/generate_seqr_project_data.py** 
Not part of core metamist.

These are mostly in fairly self-contained parts of the code. So this PR's proposal is to leave these as is for now. Most future PRs will touch only type-clean parts of the code base, so will be unaffected by these latent type checking errors. As we need to touch file with latent type problems, we can fix them then or otherwise decide what to do when the time comes.

[SET-1103]: https://cpg-populationanalysis.atlassian.net/browse/SET-1103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

db/python/layers/analysis.py
```diff
@@ -156,14 +164,15 @@ async def get_cram_size_proportionate_map(
         if not start_date:
             raise ValueError('start_date must be set')
         start_date = check_or_parse_date(start_date)
+        assert start_date is not None
```
rameshka: @jmarshall what do you think if we move the `start_date` sanity check (that raise a `ValueError`) here.
`check_or_parse_date` function will return `None` if `start_date` is empty and this will trigger the `ValueError` now. This will help us to remove this double verification.

rameshka (commented):
Hey @jmarshall thanks for taming this beast. Added a few comments to check. 
I could not go through all the changes. But thought of releasing what I have reviewed, as it might take a while to review the whole thing.

violetbrina:
I had a question about this PR. Is the goal to get the pyright checking to pass? I'm assuming we can't have dirty type checking errors on the main branch moving forward, it'll make the checks effectively useless.

I guess I'm asking, are we fixing all the errors, or are we going to ignore them in the config at least to get a clean check on main?

jmarshall:
@violetbrina: The goal is to get it as clean as practical, and especially to get all the commonly worked on files clean.

The change to _.github/workflows/lint.yaml_ means that for PRs pre-commit linters will only run on the files that are actually changed in the PR. So even if there are files with type checking errors on the **main** branch, PRs that don't touch those files will be unaffected.

If e.g. Eddie wants to do some work on his auditing code, there will be a problem… If you touch those files, your PR will need to tidy them up. (Or we could in the meantime have tidied them up at our leisure. This PR tries to hit all the core metamist code so we can be confident in that and so that future core-code-affecting PRs will be unaffected.)

Or if we need to when the time comes, we can add problem files to `[tool.pyright] exclude` in _pyproject.toml_. But IMHO we should only do that when we need to.

We've also talked about having a mode in which pre-commit runs for `--all-files` on **main** on manual dispatch or on a schedule or on pushes to **main**. That would fail for files that haven't been tidied up, but could be instructive nonetheless. (This PR doesn't re-add such a mode and is planning to leave that as a future improvement.)

TL;DR The tree is mostly clean; the few dirty files shouldn't cause trouble for most future PRs; we can eventually tidy up the rest hopefully at our leisure.

api/graphql/loaders.py
jmarshall: All the records returned by `player.get_participants_by_ids()` have their `id` fields filled in (i.e., non-None).

So if there were any Nones in the `participant_ids` argument or if there were any `participant_ids` entries that could not be found in the database, they will be present in `missing_pids`. And so we will have raised an exception on the previous line.

(Yep, I went through a reasoning exercise like this for every one of these…!)

db/python/layers/analysis.py
jmarshall: Yep, good idea.

api/graphql/loaders.py
rameshka: yes ofcourse, It is already handled in the previous line (my brain has skipped that logic 😄). Then the reasoning for iterating over `participant_ids` for the returned list has to be to ensure the order of participants in the list.

jmarshall: Yes exactly, so you get them back in the same order you asked for them.

(It doesn't really explain why the code a few lines above has `…if p is not None` because if there were any Nones in there it always raises NotFoundError, but 🤷 …)

db/python/filters/generic.py
```diff
@@ -366,18 +372,13 @@ def to_sql(
                         f'Filter {field.name} must be a GenericFilter or dict[str, GenericFilter]'
                     )
 
-        filters_rm_none: list[Template] = [f for f in filters if f is not None]
-
-        if len(filters_rm_none) == 0:
-            return None
```
rameshka, line 372: @jmarshall I recall having logic to raise an exception if the returned value is `None`. e.g., in the [tables/family_participants.py](https://github.com/populationgenomics/metamist/blob/2898b2f5ff16167e0ec0e554e94bf3d437bda3c5/db/python/tables/family_participant.py#L104). 
Also in [tables/cohort.py](https://github.com/populationgenomics/metamist/blob/2898b2f5ff16167e0ec0e554e94bf3d437bda3c5/db/python/tables/cohort.py#L72)

We may have to update that behaviour based on the new function signature.

rameshka (commented):
Hey @jmarshall the remaining changes are also looking good. Just left a tiny comment.

db/python/layers/family.py
```diff
@@ -297,8 +301,12 @@ async def import_pedigree(
...
                     family_id=external_family_id_map[row.family_id],
                     individual_id=external_participant_ids_map[row.individual_id],
-                    paternal_id=external_participant_ids_map.get(row.paternal_id),
-                    maternal_id=external_participant_ids_map.get(row.maternal_id),
+                    paternal_id=external_participant_ids_map.get(row.paternal_id)
+                    if row.paternal_id
+                    else None,
+                    maternal_id=external_participant_ids_map.get(row.maternal_id)
+                    if row.maternal_id
+                    else None,
```
hrrsheen: Could this be shortened to `paternal_id=external_participant_ids_map.get(row.paternal_id, None)`, etc?

db/python/layers/seqr.py
```diff
@@ -54,6 +54,8 @@
     SeqrDatasetType.MITO: 'MtToEsMito',
 }
 
+FAR_FUTURE = datetime.datetime(datetime.MAXYEAR, 12, 31)
```
hrrsheen: Since this is defined in two different files, I wonder if it would make sense to put its definition in `settings.py`. There's precedent for constants being stored there, which can be seen with some BQ values.

db/python/layers/family.py
jmarshall: That would be equivalent to the existing code. 😄 

The issue is that `external_participant_ids_map` has type `dict[str, int]`, so if you try to index it with `row.paternal_id` when that could possibly not be a `str` you get a type diagnostic.

db/python/layers/seqr.py
violetbrina: I like this idea. I thought settings.py was just pulling environment variables, but there are constants in there as well. As good a place as any.

db/python/layers/analysis.py
```diff
@@ -32,6 +32,7 @@
 
 
 ES_ANALYSIS_OBJ_INTRO_DATE = datetime.date(2022, 6, 21)
+FAR_FUTURE = datetime.datetime(datetime.MAXYEAR, 12, 31)
```
violetbrina: Seconding the move for these constants to go to settings.py

db/python/filters/generic.py
```diff
@@ -39,6 +39,16 @@ def get_hashable_value(value):  # noqa: PLR0911
...
+def join_sql_with_AND(clauses: list[Template]) -> Template:  # noqa: N802 allow ..._AND name
+    """Join SQL snippets with AND, dropping redundant '...AND TRUE AND...' entries"""
+
+    def _literally_true(s: Template):
+        return len(s.strings) == 1 and s.strings[0] == 'TRUE'
+
+    nontrivial = [sql for sql in clauses if not _literally_true(sql)]
+    return sql.SQL(' AND ').join(nontrivial) if len(nontrivial) > 0 else t'TRUE'
+
+
```
violetbrina: Love this, definitely should have had this ages ago we merge where conditions so often.

violetbrina (changes requested):
Wow what a PR. ALL the type errors haha. It's looking good.

db/python/layers/family.py
jmarshall: What we could do to make this less horrible is add a helper function similar to `pmap_lookup()`. Possibly even refactor things to reuse the same function in both places… 🤔

jmarshall: Can't use the same function, but it's improved with its own function.

db/python/filters/generic.py
jmarshall: Good catch. There were also a few other places quietly checking these things for None, and nearby manual joining-templates-together-with-AND that could be improved.

Fortunately most of these bad None tests would just lead to redundant `WHERE TRUE` clauses in the emitted SQL, but the exception raising sites you identify are probably a bit more important.

violetbrina (approved):
LGTM

rameshka (approved):

jmarshall:
The final counts of remaining type diagnostics as run on CI are:

```
  30   api/graphql/schema.py
   1   api/utils/db.py
   3   db/python/layers/participant.py
   1   db/python/layers/sequencing_group.py
   4   models/models/sequencing_group.py
  40   packages/metamist/src/metamist/parser/generic_metadata_parser.py
  18   packages/metamist/src/metamist/parser/generic_parser.py

  44   db/python/layers/project_insights.py
   2   packages/metamist/src/metamist/audit/adapters/storage_client.py
   4   packages/metamist/src/metamist/audit/services/reporter.py
   5   test/data/generate_seqr_project_data.py
```

These are all as previously described in https://github.com/populationgenomics/metamist/pull/1237#issuecomment-4349797810.

PR #1238, opened 2026-03-04
Call refresh_projects() only once per dependable_get_connection_getter call

jmarshall:
The `get_connection()` function returned by `dependable_get_connection_getter` is called repeatedly for related graphql resolvers. Caching the project maps in `dependable_get_connection_getter` avoids repeated calls to get the same information from `refresh_projects()` and speeds up this hot path.

But don't cache them for other `refresh_projects()` call sites for which it can be important to get a fresh list of projects.


dancoates (approved):
Very nice!

PR #1239, opened 2026-03-04
Add dbmate migration for project-creators & members-admin groups

jmarshall:
Adds a migration for the remaining initial `INSERT INTO` I needed to be able to apply (more or less) the existing instructions to get a local database up and running. This corresponds to `2023-10-06_inbuilt-permissions` in the old _db/project.xml_.

I assume it's not appropriate to add to the existing _20260120061616_add_default_data.sql_ as it's already long-since been applied to most people's installations.

And I assume we're happy to do this via a dbmate migration…

### DIFF
--- a/db/migrations/20260305092300_add_default_data.sql
+++ b/db/migrations/20260305092300_add_default_data.sql
@@ -1,0 +1,7 @@
+-- migrate:up
+
+INSERT INTO "group" (name) VALUES ('project-creators'), ('members-admin');
+
+-- migrate:down
+
+DELETE FROM "group" WHERE name IN ('project-creators', 'members-admin');


db/migrations/20260305092300_add_default_data.sql
hrrsheen, line 1: I don't think it'd cause any issues to just have these in the existing default data migration file. If anyone needs these groups for testing or development then they can add them manually easily enough, especially since it's a small group of us working on the Postgres migration. It shouldn't break anything when running `dbmate down`. Saves having an extra migration file when it comes time to deploy. Keen to hear what others think on this though.

dancoates:
Closing this as it is implemented in https://github.com/populationgenomics/metamist/pull/1242 as well - and I think it is ok to update existing migrations until we get this released. It is also that important that these are added before the foreign keys so that when migrating down if you have data in your db the foreign keys will already be gone and you won't get errors.

PR #1240, opened 2026-03-04
[SET-1032] Migrate billing tests to pytest

rameshka:
Migrate billing-related test cases from unittest to pytest.
Remove `unittest.mock.MagicMock` usages and introduce dummy mock classes (in `test/testbqbase.py`) and pytest `monkeypatch`.

### DIFF
--- a/db/python/tables/bq/generic_bq_filter.py
+++ b/db/python/tables/bq/generic_bq_filter.py
@@ -1,3 +1,4 @@
+import re
 from datetime import datetime
 from enum import Enum
 from typing import Any, TypeVar
@@ -8,6 +9,7 @@ from db.python.filters import GenericFilter
 
 
 T = TypeVar('T')
+NONFIELD_CHARS_REGEX = re.compile(r'[^a-zA-Z0-9_]')
 
 
 class GenericBQFilter(GenericFilter[T]):  # noqa: PLW1641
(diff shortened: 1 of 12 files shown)


db/python/tables/bq/generic_bq_filter.py
```diff
@@ -28,6 +30,20 @@ def __eq__(self, other):
         # all attributes are equal
         return True
 
+    @staticmethod
+    def generate_field_name(name):
```
rameshka, line 34: This function, previously in the `generic.py` (later removed with migration refactoring work), is used in GenericBQFilter. I could not find an alternative function in the generic.py; therefore added `generate_field_name` here.

dancoates (commented):
I think this looks pretty good, but the tests are very slow, around 5 second for each billing test. I found that both `MockBqClient` and `MockBqConnection` take around 2 second to instantiate. This is because they are both calling their parent classes which results in real bigquery connections being created. Removing the calls to `super().__init__()` from the `__init__` methods of these makes them complete pretty much instantaneously.

Could you also please add these tests to the include list in pytest.toml

rameshka:
Thanks for reviewing @dancoates . I will check on the suggestion.
Yes, of course I will add them to the `pytest.toml` (I often miss this step).

dancoates (approved):
Looks good

PR #1241, opened 2026-03-04
Minor PostgreSQL migration fixes

jmarshall:
Minor fixes:

* Use `sql.SQL(…)` to make the Explore / Project Summary Stats page work


dancoates (approved):

jmarshall:
I anticipated there would be further minor fixes to add to this PR, but that one ended up being the only failure I saw on the main metamist UI pages.

PR #1242, opened 2026-03-06
[PG] Update installation docs

hrrsheen:
Updates the `docs/installation.md` file to reflect the process for setting up with Postgres.
Additionally adds necessary default information for the groups table to the db migrations.

### DIFF
--- a/db/migrations/20260120061616_add_default_data.sql
+++ b/db/migrations/20260120061616_add_default_data.sql
@@ -1,5 +1,7 @@
 -- migrate:up
 
+INSERT INTO "group" (name) VALUES ('project-creators');
+INSERT INTO "group" (name) VALUES ('members-admin');
 INSERT INTO assay_type (id, name) VALUES ('sequencing', 'sequencing');
 INSERT INTO sample_type (id, name) VALUES ('blood', 'blood');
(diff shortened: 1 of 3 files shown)


db/migrations/20260120061616_add_default_data.sql
```diff
@@ -27,6 +29,7 @@ INSERT INTO analysis_type (id, name) VALUES ('analysis-runner', 'analysis-runner
 
 -- migrate:down
 
+DELEtE FROM "group" WHERE name IN ('project-creators', 'members-admin');
```
jmarshall: FUnKy :smile:

jmarshall:
Note that the _db/migrations_ bit duplicates PR #1239 and the same question applies about whether we are early enough in development that it's kosher to modify an already existing and applied migration script.

It might be worth adding a section summarising “what environment variables etc you need to change from your existing mariadb setup”.

dancoates (approved):
I just pushed a couple of tiny changes but happy with this otherwise. I'm ok to be updating migrations retrospectively until we get this released. 

> It might be worth adding a section summarising “what environment variables etc you need to change from your existing mariadb setup”.

This is a good idea, but I think better off as a confluence doc as it will be useful at a point in time and doesn't need to be long term documentation. I'll add a ticket to write this up.

PR #1243, opened 2026-03-11
SET-999 Add PBMC viability report to OurDNA dashboard

hrrsheen:
Adds the following visualisations for PBMC samples to the OurDNA reports dashboard under the "Sample Processing" tab:

Percent Viability by Collection Date
- This is a scatter plot of percent viability against collection date.
- A small drop-down menu has been added just above the plot that allows data points to be colour coded according to either biobank, collection event type or collection lab.

A box-and-whisker plot of PBMC viabilities by the biobank, event-type, collection-lab combination
- Visually displays summary statistics for different collection/processing conditions.

A table of the statistics from the box-and-whisker plot
- Contains minimum and maximum (excluding outliers), median (Q2), first quartile value (Q1), third quartile value (Q3) and average for each condition.

**Additionally** the `generate_ourdna_data.py` script was modified to eliminate the need to manually create an OurDNA project in the database before generating test data.

### DIFF
--- a/test/data/generate_ourdna_data.py
+++ b/test/data/generate_ourdna_data.py
@@ -4,6 +4,8 @@ This is a simple script to generate some participants and samples for testing ou
 Local Backend API needs to run prior executing this script
 
 """
+import os
+import sys
 
 import argparse
 import datetime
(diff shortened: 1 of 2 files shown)


dancoates (commented):
This is looking good, I think we just need to exclude outliers from the first two charts too. At the moment there's one sample with a typo for viability that is throwing out the chart.

Image

Looking at the prod data there's one value of 998 and one of 0, the next lowest is about 47 and the next highest is 100 so potentially could cap between 50 and 100?

ckrishk (approved):
LGTM

dancoates (approved):
This looks good, nice work!

PR #1244, opened 2026-03-11
SET-1048 Migrate remaining tests, fix some errors along the way

dancoates:
**test updates:**

There were some updates required to tests to work with pytest.

- Fairly major updates to the test_cohort_builder.py script to get the tests passing. The script it was trying to test uses sync code and the mixing with the async code of the api was causing a bunch of problems. I don't think this script is used so should be ok to make the tests a bit more basic, and just test the cohort functionality itself rather than the script.
- Some fixes for test_layers_billing to handle the changes to mocking made in #1240
- Some major changes to expected results in test_web, these tests were incorrectly named in dev so were never actually being run, the test names were fixed during the migration but the expected results had always been incorrect. We probably could have just deleted these tests as they were never run, but they're fixed now so may as well keep them around.

**bugfixes:**

- Fix GenericFilter isnull: False querying of json fields. Previously if the json path query the user passed resulted in a non-scalar value then NULL would be returned, resulting in the filter failing. This was different to the mariadb implementation which would correctly return rows. The fix was to switch from json_value to json_query.
- Fix an existing bug in assay mutations where the the project would not be passed through to table methods when upserting an assay. This was broken in graphql mutations but not picked up because previously the graphql tests used a connection with a project attached which doesn't match the realworld implementation.
- Fix for external id updates. External ID columns have two unique indexes, one on name and the entity id, and a second on on project and external id. The mariadb implementation would handle conflicts on both indexes but during the migration we switched to only handling conflicts on the entity_id, name index. This is fixed by switching to a merge statement that checks both conditions. In the future we'd be better off avoiding the entity_id, name index conflcits by deleting all external ids for an entity and then re-adding them.
- Fix project meta not being correctly converted to Jsonb
- Fix some bad query building in sample table where spaces weren't added between query template sections resulting in syntax errors.
- Fix sample meta table meta columns not being cast to text


test/test_audit_log.py
```diff
@@ -22,8 +24,10 @@ async def test_insert_sample(self):
             )
         )
 
-        sample_cl_id = await self.connection.connection.fetch_val(
-            'SELECT audit_log_id FROM sample WHERE id = :sid', {'sid': sample.id}
+        cur = await connection_with_project.pg_connection.execute(
+            'SELECT audit_log_id FROM sample WHERE id = %(sid)s', {'sid': sample.id}
```
hrrsheen: Could this just be a t-string?

test/test_api_billing.py
```diff
@@ -182,12 +189,12 @@ async def test_get_running_cost(
...
             author=TEST_API_BILLING_USER,
         )
-        self.assertEqual(mockup_record, records)
+        assert mockup_record == records
 
         # Verify the layer method was called with the query object
         mock_get_running_cost_with_filters.assert_called_once_with(query)
 
-    @run_as_sync
+    @pytest.mark.asyncio
```
ckrishk: Nice!

test/test_cohort_builder.py
```diff
@@ -124,96 +86,38 @@ async def setUp(self):
...
-        self.assertEqual(spec.description, 'Every criterion')
-
-        criteria = body['cohort_criteria']
-        self.assertIsInstance(criteria, metamist.models.CohortCriteria)
-        self.assertListEqual(criteria.projects, [self.project_name])
-        self.assertListEqual(criteria.excluded_sgs_internal, [self.sgB, self.sgC])
-        self.assertListEqual(criteria.sg_technology, ['short-read'])
-        self.assertListEqual(criteria.sg_platform, ['illumina'])
-        self.assertListEqual(criteria.sg_type, ['genome'])
-        self.assertListEqual(criteria.sample_type, ['blood'])
```
hrrsheen, line 212: What's the reason that we don't need to test for these fields?

dancoates: This test was previously testing the `create_custom_cohort.py` script which does some funny stuff to manually build the cohort criteria and return it. I couldn't get testing of that script to work with pytest as it runs sync and these tests are async, and I'm pretty sure we don't use the script anyway so I didn't want to put a bunch of time into it. So this switches to just testing the API endpoints instead which only return the cohort and not the criteria.

test/test_audit_log.py
dancoates: Thanks, I've updated that one

test/test_upsert.py
```diff
@@ -211,9 +214,9 @@ async def test_insert_participants(self):
         """
 
         # Table interfaces
-        pt = ParticipantLayer(self.connection)
+        pt = ParticipantLayer(connection_with_project)
 
-        await pt.upsert_participants(all_participants, open_transaction=False)
```
rameshka, line 216: Hey @dancoates, this comment isn’t related to this PR.

When I saw this change to remove the `open_transaction` parameter, I did a quick search across the project and saw two usages of `AssayLayer.upsert_assays` in the Sample Layer where `open_transaction` is still being passed. Not sure if those were addressed in another PR.

rameshka (approved):
Hey @dancoates, thanks for taking up this task. There were a lot of tests to update. I’ve just left a small comment.

test/test_upsert.py
dancoates: Thanks, that's a good pickup. I think that should be detected by pyright but I'll make a note to double check that is fixed in the pyright PR

hrrsheen (approved):
That's a lotta tests! LGTM.

PR #1245, opened 2026-03-17
Fix bug in graphql sg analyses resolver

dancoates:
This resolver was causing problems with exhausting connections in the connection pool. A connection was obtained from the pool and then the call to the loader would cause another connection to be obtained from the pool without returning the first one. In the case where lots of calls were made to this resolver, this resulted in exhausting pool connections before any could be returned, resulting in timeouts. The solution is to move the loader out from the connection context manager so that the connection is returned to the pool before the loader runs.

### DIFF
--- a/api/graphql/schema.py
+++ b/api/graphql/schema.py
@@ -1191,8 +1191,6 @@ class GraphQLSequencingGroup:
         project: GraphQLFilter[str] | None = None,
     ) -> list[GraphQLAnalysis]:
         async with info.context['get_connection']() as connection:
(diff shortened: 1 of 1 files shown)


jmarshall (approved):
LGTM. Just checking `SET-1017-…` was your intended base branch? I guess it will make its way to all the other branches quickly enough in any case!

dancoates:
> LGTM. Just checking SET-1017-… was your intended base branch? I guess it will make its way to all the other branches quickly enough in any case!

Thanks, yeah the PRs are stacked so I won't merge this one till the ones it depends on are merged

PR #1246, opened 2026-03-18
Update api auth to accept requests from legacy sample metadata proxy

dancoates:
While migrating, there will still be requests to the old sample metadata server, we'll need those requests to continue working - so the old server will be switched to act as a proxy and forward requests to the new server. It will authenticate with the new server with its own service account and forward the requesting user in a header. This update checks the service account and header and if they exist and are correct, will act as if that user authenticated directly.


ckrishk (approved):
Looks good. Lets give this go . However codecov seems to unhappy about missing line coverage

PR #1247, opened 2026-03-19
migrate queries that we missed during previous migration

dancoates:
There were a few database queries hidden away that we missed earlier in the migration, this catches all of those and removes the databases library as it is no longer needed.

### DIFF
--- a/db/python/tables/bq/billing_ar_batch.py
+++ b/db/python/tables/bq/billing_ar_batch.py
@@ -116,25 +116,26 @@ class BillingArBatchTable(BillingBaseTable):
         Get min timestamp by sequencing groups
         """
         # Find the minimum day when samples has been added
-        # using FOR SYSTEM_TIME ALL gives us full history
-        _query = """
+        # Using history.sequencing_group to get previous states
+        _query = t"""
         SELECT sg.id, MIN(l.timestamp) as start_date
-        FROM sequencing_group FOR SYSTEM_TIME ALL sg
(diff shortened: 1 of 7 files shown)


db/python/tables/bq/billing_ar_batch.py
```diff
@@ -145,18 +146,13 @@ async def get_number_of_sequencing_groups(
...
+            WHERE s.project = ANY({[project_id]})
         """
-        info_record = await connection.connection.fetch_one(
-            _query,
-            {
-                'projects': [project_id],
-            },
-        )
+        info_record = await (await connection.pg_connection.execute(_query)).fetchone()
         return dict(info_record).get('cnt', 0)
```
rameshka: small suggestion: `info_record` is already a dictionary.

db/python/tables/bq/billing_ar_batch.py
```diff
@@ -165,21 +161,16 @@ async def get_total_crams_info(
...
         """
-        info_record = await connection.connection.fetch_one(
-            _query,
-            {
-                'projects': [project_id],
-            },
-        )
+        info_record = await (await connection.pg_connection.execute(_query)).fetchone()
 
         total_cram_info = dict(info_record)
```
rameshka: similarly here as well

rameshka (approved):
Hey @dancoates . Changes are looking good to me. Just left a tiny comment, which is really minor.

PR #1248, opened 2026-03-20
[PG] audit case insensitive matching

hrrsheen:
The mariadb database used a collation of `utf8mb4_general_ci` which makes text comparisons case insensitive, ie. `'TEXT' = 'text'` is `TRUE`. In order to maintain backwards compatibility, queries where text comparison is performed between user input and database values have been modified to be case-insensitive.


db/python/layers/project_insights.py
```diff
@@ -415,7 +415,7 @@ async def _total_families_by_project_id_and_seq_fields(
             LEFT JOIN sequencing_group sg on sg.sample_id = s.id
         WHERE
             f.project = ANY({project_ids})
-            AND sg.type = ANY({sequencing_types_param})
+            AND LOWER(sg.type) = ANY({sequencing_types_param})
```
rameshka: hey @hrrsheen,
A general comment on casting the enum types.
In this DB query, I could see that the [enum types are cast to lowercase and persisted to the DB](https://github.com/populationgenomics/metamist/blob/603255c6d16ca82935d42daecf3829d91cb8f677/db/python/enum_tables/enums.py#L62). 
And this `type` column in the `sequencing_group` table has a FK reference to the 'sequencing_type' table. So I think we can ignore the case casting `sg.type` here.

dancoates: I can also confirm that in prod all the `*_type` tables have only lower case values

hrrsheen: Thanks @rameshka and @dancoates. I've gone and removed calls to `LOWER` on rows that have a FK relationship to enum tables.

dancoates (approved):
This looks good, just need to fix the linting and then happy to merge

PR #1249, opened 2026-03-23
SET-921 pg script to migrate data from mariadb to postgres

dancoates:
This has been tested on both the production and development databases and seems to work well. Copying the production database takes around ~2 minutes.

The script uses duckdb as it is able to efficiently download data from mariadb, and then upload it to postgres, while performing transformations. It also allows us to copy all the data from mariadb system versioned tables into the new postgres history tables. 

The script does some handling of migrations, first migrating all the way down to remove any existing data in the database, then migrating up until all the tables are created, but stopping before foreign keys are added so that table data can be inserted quickly and in any order. Then after all the data has been copied, it applies all the final migrations which acts as a sort of integrity check of the data that was just copied. If any referential data was incorrectly copied, the adding of fks and constraints will fail. 

This is a one-off temporary script and can be removed once data has been copied and we're happy with the results. 

@populationgenomics/genomic-analysis-team you will be automatically tagged to review this as it is in the scripts folder, but no need to review unless anyone particularly wants to.

### DIFF
--- a/scripts/copy_mariadb_data/copy_from_mariadb_to_postgres.py
+++ b/scripts/copy_mariadb_data/copy_from_mariadb_to_postgres.py
@@ -1,0 +1,904 @@
+# /// script
+# dependencies = [
+#   "duckdb",
+#   "python-dotenv",
+#   "click",
+# ]
+# ///
+
+
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import urllib.parse
+from dataclasses import dataclass
+from pathlib import Path
+
+import click
+import dotenv
+import duckdb
+
+
+dotenv.load_dotenv()
+
+
+@dataclass
+class ColumnInfo:
+    """Information about a database column."""
+
+    source_name: str  # mariadb column name
+    is_json: bool = False  # whether the column contains JSON data
+    dest_name: str | None = None  # postgres column name
+    is_deleted: bool = False  # whether the column isn't present in postgres
+    is_added: bool = False  # whether the column is new in postgres
+    to_lowercase: bool = False  # whether to convert value to lowercase
+
+
+@dataclass
+class TableInfo:
+    """Information about a database table."""
+
+    name: str
+    columns: list[ColumnInfo]
+    has_system_versioning: bool = False
+
+
+maria_db_database = os.getenv('MARIADB_DB')
+maria_db_password = os.getenv('MARIADB_PWD')
+maria_db_user = os.getenv('MARIADB_USER')
+maria_db_port = os.getenv('MARIADB_PORT')
+maria_db_host = os.getenv('MARIADB_HOST')
+
+postgres_password = os.getenv('POSTGRES_PASSWORD')
+postgres_database = os.getenv('POSTGRES_DB')
+postgres_user = os.getenv('POSTGRES_USER')
+postgres_port = os.getenv('POSTGRES_PORT')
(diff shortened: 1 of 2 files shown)


scripts/copy_mariadb_data/copy_from_mariadb_to_postgres.py
```diff
@@ -0,0 +1,931 @@
...
+        columns=[
+            ColumnInfo(source_name='id'),
+            ColumnInfo(source_name='parent_id'),
+            ColumnInfo(source_name='content'),
+            ColumnInfo(source_name='status'),
+            ColumnInfo(source_name='audit_log_id'),
+        ],
+    ),
+    TableInfo(
+        name='family',
```
hrrsheen, line 292: This table is missing the `meta` field.

scripts/copy_mariadb_data/copy_from_mariadb_to_postgres.py
```diff
@@ -0,0 +1,931 @@
...
+            SELECT * FROM postgres_query(
+                'pg_db',
+                'SELECT setval(pg_get_serial_sequence(''main.{table_name}'', ''{column_name}''), COALESCE((SELECT MAX({column_name}) FROM main.{table_name}), 1))'
+            );
+        """)
+
+    print('All identity sequences reset successfully.')
+
+
+def insert_postgres_data():  # noqa: D103
```
hrrsheen, line 741: Could it be worth while to make all of these inserts part of the same transaction?

scripts/copy_mariadb_data/copy_from_mariadb_to_postgres.py
dancoates: Good pickup! I thought I'd added that but must have missed it. Have put it in there now.

scripts/copy_mariadb_data/copy_from_mariadb_to_postgres.py
dancoates: Not in this case - if anything in the migration fails you'd start the migration again by deleting all the tables and re-importing the data, so there's no need to use transactions for the inserts. I'm honestly not sure if it'd work, it'd consume a lot of memory putting so many inserts in one transaction without committing them

scripts/copy_mariadb_data/copy_from_mariadb_to_postgres.py
```diff
@@ -0,0 +1,939 @@
...
+    ('participant', 'id'),
+    ('project', 'id'),
+    ('sample', 'id'),
+    ('sequencing_group', 'id'),
+]
+
+
+def reset_identity_sequences():
+    """Reset PostgreSQL identity sequences to the max value in each table."""
+    # duck.execute("""
```
hrrsheen: Can this comment be removed? The same query is executed before `reset_identity_sequences()` is called.

dancoates: Yep, good pickup, this was left over from testing. I've removed it now.

scripts/copy_mariadb_data/copy_from_mariadb_to_postgres.py
hrrsheen: Makes sense, thanks for clarifying!

hrrsheen (approved):
LGTM.

PR #1250, opened 2026-03-23
SET-1063 migration testing bugfixes

dancoates:
Misc group of bugfixes from testing migration work


rameshka (approved):
Thanks fixing these queries @dancoates. Changes are looking good.

PR #1253, opened 2026-03-23
Update metamist python package to support IAP

dancoates:
Previously the metamist package would call the cloud run instance rather than the proper metamist URL. As of https://github.com/populationgenomics/cpg-utils/pull/190 cpg utils now supports getting an id token for users via desktop oauth flow so we can switch to calling the proper metamist url which hits the load balancer behind IAP.

This new flow only affects human users. Service accounts and compute instances continue to authenticate in the same way they always have.

In addition to the new auth flow, this PR includes a basic metamist cli to allow users to run `metamist auth init` to start the auth process and cache a refresh token so that subsequent usages of the metamist module won't require going through the manual auth flow.

The cli makes available some other existing cli tools within the metamist package which were previously not surfaced.

To test, check out this branch, then run `uv sync` then `uv metamist auth init`


packages/metamist/src/metamist/metamist_config.py
```diff
@@ -0,0 +1,127 @@
...
+    env='production',
+    audience='313488420322-5l74h33pn88m9hl9t9ffvoqe6vs62lg2.apps.googleusercontent.com',
+    desktop_oauth_creds_secret='projects/313488420322/secrets/metamist-python-client-desktop-credentials',
+)
+
+LOCAL_CONFIG = MetamistConfig(
+    url='http://localhost:8000',
+    env='local',
+    audience='',
+    desktop_oauth_creds_secret='',
```
hrrsheen, line 104: When running `uv run metamist auth init` with the environment set to `local`, `get_google_identity_token()` raises an exception because these fields are empty.

dancoates: Thanks, `auth init` doesn't need to be run for local env but you're right that we should still handle it. I've updated the cli to print a message explaining that dev or production env needs to be used to run `auth init`

hrrsheen (approved):

PR #1254, opened 2026-03-26
SET-1061 set up existing cloud run to proxy requests to new metamist

dancoates:
Once we are ready to move users/traffic across to the new metamist service, we will need to continue supporting requests to the old service for some time, as the urls for the old cloud run instance are hard-coded into the python package. This PR replaces the server handler with one that proxies API requests to the new cloud run, or redirects web requests so that users accessing via the browser will be redirected to the new metamist website.

This works in conjunction with the changes from #1246 which accept these proxy requests. 

The two env vars that this uses have already been added to the `sample-metadata-api-dev` and `sample-metadata-api` cloud run services. 

Github doesn't display the diff of this very well, basically server_old.py is an unchanged version of server.py, kept the same so that the deploy action including `regenerate_api.py` will still work. And the new `server.py` is essentially a new file, not updates to the old `server.py`

__NOTE: This should NOT be deployed until we are happy to move all traffic across to the postgres database.__

### DIFF
--- a/api/server.py
+++ b/api/server.py
@@ -1,0 +1,129 @@
+import os
+import time
+from typing import Annotated
+
+import httpx
+from fastapi import FastAPI, Request
+from fastapi.concurrency import run_in_threadpool
+from fastapi.middleware.cors import CORSMiddleware
+from fastapi.params import Depends
+from starlette.background import BackgroundTask
+from starlette.responses import RedirectResponse, StreamingResponse
+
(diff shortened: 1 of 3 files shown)


hrrsheen (approved):
LGTM

dancoates:
Closing in favour of https://github.com/populationgenomics/metamist/pull/1328

PR #1255, opened 2026-03-26
Change metamist project env var to be set to deploy gcp project

dancoates:
This var wasn't used in many places but still needed updating to ensure gcp actions in billing and output files are based on the correct project. This goes with https://github.com/populationgenomics/metamist-private/pull/32 which removes the value for this env var from the config.


hrrsheen (approved):
LGTM

PR #1257, opened 2026-03-26
Make cloud run instance and db pool size configurarable

dancoates:
These are the parameters that we can tweak to stay on top of connections to the database and ensure we're not under or over utilizing connections. A config update in metamist-private will attempt to implement sensible initial values for these.

Accompanying metamist-private PR is here: https://github.com/populationgenomics/metamist-private/pull/33


deploy/infra/server.py
```diff
@@ -176,6 +177,14 @@ def create_server_resources(
                             name='SM_LEGACY_PROXY_SA',
                             value=config.server.sm_legacy_proxy_sa,
                         ),
+                        gcp.cloudrunv2.ServiceTemplateContainerEnvArgs(
+                            name='DB_POOL_MIN_SIZE',
+                            value=config.server.sm_legacy_proxy_sa,
+                        ),
+                        gcp.cloudrunv2.ServiceTemplateContainerEnvArgs(
+                            name='DB_POOL_MAX_SIZE',
+                            value=config.server.sm_legacy_proxy_sa,
```
hrrsheen: It seems like the `value`s here should be `config.server.db_min_pool_size` and `config.server.db_max_pool_size`, respectively.

milo-hyben: Thank you Harrison, Had addressed those. Please review again.

hrrsheen (approved):

PR #1260, opened 2026-04-09
Update analysis moving script to skip analyses with old project

EddieLF:
Updates the analysis mover script so that analyses that are still in the old project don't get touched.

For example, the STRipy index pages produced in the old project that include many SGs, including the one targeted. We don't want to move these multi-SG analyses, but they are picked up when you search for `analysis.type='web'` (Although we may want to archive those analyses and create them fresh, since they now have SGs across multiple dataset)

Caught on a dry run: https://batch.hail.populationgenomics.org.au/batches/1127974/jobs/1
```
Analysis: 123456. SG: CPGXXXXXX. Type: web.
    New outputs path: gs://cpg-{new_dataset}-main-web/stripy/1.2/{new_dataset}_index.html
    New meta dataset: {new_dataset}
DRY RUN :: Skipping updating analysis 123456
```

In this case, analysis 123456 is associated with the old project, so by adding this condition it becomes out of scope.

### DIFF
--- a/scripts/move_analyses_across_projects.py
+++ b/scripts/move_analyses_across_projects.py
@@ -48,6 +48,9 @@ ANALYSES_QUERY = gql(
                     type
                     meta
                     outputs
+                    project {
+                        name
+                    }
                 }
             }
         }
@@ -134,6 +137,10 @@ def get_analyses_to_update_and_files_to_move(
     files_to_move = []
     for sg in analysis_query_result['project']['sequencingGroups']:
         for analysis in sg['analyses']:
+            if analysis['project']['name'] != new_dataset:
+                # This skips the analyses that haven't been updated with the new project
+                continue
+
             if (
                 analysis['type'] == 'sv'
                 and analysis['meta'].get('stage') != 'GatherSampleEvidence'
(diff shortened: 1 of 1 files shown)


scripts/move_analyses_across_projects.py
```diff
@@ -134,6 +138,10 @@ def get_analyses_to_update_and_files_to_move(
     files_to_move = []
     for sg in analysis_query_result['project']['sequencingGroups']:
         for analysis in sg['analyses']:
+            if analysis['project']['name'] != new_dataset:
```
Johnnyassaf, line 141: just making sure I understand this change, This implies there is a step before this function runs where the analysis project is changed in the database for some of thse entries and only those changed entries are considered

Johnnyassaf (approved):

scripts/move_analyses_across_projects.py
EddieLF: @Johnnyassaf yup, check out the docstring at the very top of the script

```
This script assumes that the project field for all the sequencing group and
analysis records has been updated in the database with the new project id.
```

PR #1265, opened 2026-04-29
Update md5 script to produce md5 files matching the standard

EddieLF:
MD5 checksum files created by `md5sum` typically look like `<hash>  <filename>`
```bash
cat filetohashA.txt.md5
595f44fec1e92a71d3e9e77456ba80d1  filetohashA.txt
```

Since files are streamed through stdin to `md5sum`, the "filename" aspect of the written checksum files is just the stdin character `'-'`. To conform to the expected md5 file format, we need to replace the stdin character with the filename.

This PR also replaces `gsutil` with `gcloud storage` and `os.path` with `cpg_utils.Path` to modernise it and remove the linting violations.

### DIFF
--- a/scripts/create_or_validate_md5s.py
+++ b/scripts/create_or_validate_md5s.py
@@ -1,15 +1,15 @@
-import os
 from shlex import quote
 
 import click
 from google.cloud import storage
 
+from cpg_utils import to_path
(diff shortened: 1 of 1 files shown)


amyminiter (approved):
👍

PR #1266, opened 2026-05-01
Fix for Billing Cost page, bug introduced with conversion to pg.

milo-hyben:
Small fix for Billing page `Cost Across Invoice Months (Topic only)`.
Currently page is crashing due to `psycopg.errors.AmbiguousColumn: column reference "id" is ambiguous`


violetbrina (approved):
LGTM

PR #1267, opened 2026-05-01
Fix number 2 for Billing Cost page, bug introduced with conversion to pg.

milo-hyben:
One more fix for the same SQL, this time:
`psycopg.errors.UndefinedTable: missing FROM-clause entry for table "s"`


rameshka (approved):

violetbrina (approved):
LGTM

PR #1268, opened 2026-05-01
Take 3 to fix for Billing Cost page, bug introduced with conversion to pg.

milo-hyben:
Another `psycopg.errors.AmbiguousColumn: column reference "id" is ambiguous`, should be the last one.


violetbrina (approved):
Haha third time's the charm?

PR #1269, opened 2026-05-04
Reinstate function that was erroneously thought to be unused

jmarshall:
`get_participant_ids_sequencing_group_ids_for_sequencing_type()` was removed in #1204 on the basis that it was unused, but in fact it is used in `SeqrLayer.sync_dataset()`.

This PR reinstates it and converts it to use a Postgres query.


violetbrina (approved):
LGTM

PR #1270, opened 2026-05-11
SET-1045: Add toggle for Avg. Sample Cost rows in Cost Across Invoice Months

nevoodoo:
## Summary

Adds a toggle on the **Cost Across Invoice Months (Topic only)** page (`/billing/costByMonth`) that lets users include or exclude the `Avg. Sample Storage Cost (Est.)` and `Avg. Sample Compute Cost (Est.)` rows in both the displayed table and CSV/TSV exports.

PM team currently deletes these rows from every export — this toggle removes that manual step.

## Behaviour

- New `Include Avg. Sample Cost (Est.) rows` toggle, rendered under the date/topic filters.
- Defaults to **on** so existing behaviour is preserved.
- When off, the two `Avg. Sample …` rows are hidden from the table *and* omitted from CSV / TSV exports.
- Setting persists in the URL (`?includeAvgSampleCost=false`) so users can bookmark/share their preferred view.
- All-Topics row is unchanged (it never carried the Avg. Sample Cost values).

## Files

- `web/src/pages/billing/BillingCostByMonth.tsx`
- `web/src/pages/billing/components/BillingCostByMonthTable.tsx`

## Test plan

- [x] Load `/billing/costByMonth`; confirm Avg. Sample Storage/Compute rows render by default
- [x] Toggle off; confirm the two Avg. Sample rows disappear from the table and the URL gains `includeAvgSampleCost=false`
- [x] Export to CSV with toggle off — confirm the exported file contains only `Compute Cost` and `Storage Cost` rows per topic
- [x] Export to TSV with toggle on — confirm Avg. Sample rows are present (unchanged behaviour)
- [x] Reload page with `?includeAvgSampleCost=false` in URL; confirm toggle initialises to off

### DIFF
--- a/web/src/pages/billing/BillingCostByMonth.tsx
+++ b/web/src/pages/billing/BillingCostByMonth.tsx
@@ -1,9 +1,16 @@
-import { Box, Link as MuiLink, Modal as MuiModal, Typography } from '@mui/material'
+import {
+    Box,
+    FormControlLabel,
+    Link as MuiLink,
+    Modal as MuiModal,
+    Switch,
+    Typography,
+} from '@mui/material'
 import { SelectChangeEvent } from '@mui/material/Select'
 import { debounce } from 'lodash'
 import * as React from 'react'
 import { useLocation, useNavigate, useSearchParams } from 'react-router-dom'
-import { Button, Card, Checkbox, Dropdown, Grid, Message } from 'semantic-ui-react'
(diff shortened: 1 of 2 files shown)


milo-hyben (approved):
Looking good, just the default should be OFF, had made a commit with that change.
Not many people would need the avg sample cost and it is expensive calculations.

PR #1271, opened 2026-05-14
[SET-1153] Fix create analysis runner log mutation

rameshka:
As of now, the `createAnalysisRunnerLog` GraphQL mutation fails. This is because the `self.project_id` (in the `connection` object) is `None` when invoked from the GraphQL. The GraphQL `loader` sets a `get_projectless_db_connection`; unlike in the fast-api routes, the connection object is populated with the project information. 
In this fix, the `project id` value is retrieved from the `AnalysisRunnerInternal` object, which is set in both [Fast-API routes](https://github.com/populationgenomics/metamist/blob/12dc4150f78ac6cf4a63f4281ed986fb2800a177/api/routes/analysis_runner.py#L59) as well as in the [GraphQL mutation](https://github.com/populationgenomics/metamist/blob/12dc4150f78ac6cf4a63f4281ed986fb2800a177/api/graphql/mutations/analysis_runner.py#L72).

**Testing**

- Existing unit tests passing
- Added new unit test for the GraphQL api
- Manually tested GraphQL mutation
- Manually tested Rest API

### DIFF
--- a/api/graphql/mutations/analysis_runner.py
+++ b/api/graphql/mutations/analysis_runner.py
@@ -72,7 +72,8 @@ class AnalysisRunnerMutations:
                     project=target_project.id,
                     audit_log_id=None,
                     output_path=analysis_runner.output_path,
-                )
+                ),
+                target_project.id
             )
 
             return analysis_id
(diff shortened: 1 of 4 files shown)


violetbrina (approved):
Nice fix for this bug. New test looks good.

PR #1272, opened 2026-05-20
[PG] `not-included-sample-ids` endpoint fix

hrrsheen:
The query for this endpoint was attempting to access a column that didn't exist on the `sequencing_group` table. This PR adds a `JOIN` statement to allow the information to be accessed.


rameshka (approved):
LGTM !

PR #1274, opened 2026-05-20
[PG] fix `latest-complete` endpoint

hrrsheen:
The query for this endpoint was still using MariaDB conventions. This PR updates the query and surrounding code to work with postgres.

### DIFF
--- a/db/python/tables/analysis.py
+++ b/db/python/tables/analysis.py
@@ -349,7 +349,7 @@ class AnalysisTable(DbBase):
         meta_query = t''
         if meta:
             for k, v in meta.items():
-                meta_query += t" AND meta::json->>{k} = {v}"
+                meta_query += t' AND meta::json->>{k} = {v}'
 
         _query = t"""
             SELECT a.id as id, a.type as type, a.status as status,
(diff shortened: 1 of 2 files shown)


db/python/tables/analysis.py
```diff
@@ -345,17 +345,11 @@ async def get_latest_complete_analysis_for_type(
...
+        meta_query = t''
         if meta:
             for k, v in meta.items():
-                k_replacer = f'meta_{k}'
-                meta_str += f" AND json_extract(meta, '$.{k}') = :{k_replacer}"
-                if v is None:
-                    # mariadb does a bad cast for NULL
-                    v = 'null'  # noqa: PLW2901
-                values[k_replacer] = v
+                meta_query += t' AND meta::json->>{k} = {v}'
```
rameshka: Hey @hrrsheen, just checking out this PR. If `v` is `None`, then psycopg builds the query with value `NULL`. For example: `meta::json->>'size' = NULL`. But for NULL comparison, [it will be best to use `IS NULL` ](https://www.postgresql.org/docs/17/functions-comparison.html) check.

hrrsheen: Thanks @rameshka for the comment! I've updated the query and the test to account for `NULL` values in the `meta`.

rameshka (approved):
Thanks @hrrsheen. LGTM !

PR #1276, opened 2026-05-26
Support multi-SG-type samples in create_test_subset.py

michael-harper:
## Summary
- Pull every sequencing group (and its assays + analyses) for each selected sample, so test subsets can include co-located SG types — specifically WGS + genotyping array on the same sample.
- Fix `transfer_analyses` to attach each analysis to the SG matched by `(type, platform, technology)` instead of the arbitrary first SG of the sample (the previous behaviour produced wrong `sequencing_group_ids` on the update path and wrong SG-id substitution in the rewritten file path).
- Source analysis paths from `outputs.path` (the structured field) with a fallback to the legacy `output` string, so newer-style records — including the `genotypingarray_gtc` analyses on array SGs whose `output` is empty — aren't silently dropped on transfer.
- Widen the analyses-type filter on `QUERY_ALL_DATA` to `["cram", "gvcf", "genotypingarray_gtc"]`.

## Context
Needed to produce local test data for [single_sample_qc_popgen#13](https://github.com/populationgenomics/single_sample_qc_popgen/pull/13), which compares WGS and genotyping-array SGs for the same sample.

## Test plan
- [ ] Run locally against a real source project: confirm both genome and genotypingarray SGs (plus assays) land in `-test` for a sample that has both.
- [ ] Confirm CRAM/gVCF and GTC files are copied to `gs://cpg-<project>-test/...` and that each new analysis is attached to the correct SG (not both).
- [ ] Confirm the rewritten file paths use the correct new SG id per analysis (no cross-wiring between genome and array).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -234,11 +234,12 @@ def get_sids_by_random_sampling(
     logger.info(f'Querying all samples in {project}')
     sid_output = query(SG_ID_QUERY, variables={'project': project})
 
-    # all_sids here is a set of all internal IDs in the project, minus any we already selected
+    # all_sids here is a set of all internal IDs in the project, minus any we already selected.
+    # Samples with no active sequencing groups are skipped — nothing to transfer.
     all_sids = {
         sid['id']
         for sid in sid_output.get('project').get('samples')
-        if sid.get('sequencingGroups')  # skip samples with no active SGs — nothing to transfer
+        if sid.get('sequencingGroups')
     } - samples_so_far
 
     logger.info(
@@ -249,7 +250,7 @@ def get_sids_by_random_sampling(
     return random.sample(sorted(all_sids), sample_count)
 
 
-def main(
+def main(  # noqa: PLR0913
     project: str,
     samples_n: int,
     families_n: int,
@@ -1135,7 +1136,7 @@ def extra_file_commands(batch, job, new_path: str):
         batch.write_output(job.new_md5, new_path)
 
 
-def copy_files_in_dict(
+def copy_files_in_dict(  # noqa: PLR0911
     d,
     dataset: str,
     sid_replacement: tuple[str, str] = None,
@@ -1207,7 +1208,9 @@ def copy_files_in_dict(
     if isinstance(d, list):
         return [copy_files_in_dict(x, dataset, skip_gcs=skip_gcs) for x in d]
     if isinstance(d, dict):
-        return {k: copy_files_in_dict(v, dataset, skip_gcs=skip_gcs) for k, v in d.items()}
+        return {
+            k: copy_files_in_dict(v, dataset, skip_gcs=skip_gcs) for k, v in d.items()
+        }
     return d
 
 
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -236,7 +236,9 @@ def get_sids_by_random_sampling(
 
     # all_sids here is a set of all internal IDs in the project, minus any we already selected
     all_sids = {
-        sid['id'] for sid in sid_output.get('project').get('samples')
+        sid['id']
+        for sid in sid_output.get('project').get('samples')
+        if sid.get('sequencingGroups')  # skip samples with no active SGs — nothing to transfer
     } - samples_so_far
 
     logger.info(
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -257,12 +257,15 @@ def main(
     cohorts: set[str],
     skip_ped: bool,
     update_embedded_ids: bool,
+    local_mode: bool = False,
+    dry_run: bool = False,
 ):
     """
     Script creates a test subset for a given project.
     A new project with a suffix -test is created, and for any files in sample/meta,
     sequence/meta, or analysis/output a copy in the -test namespace is created.
     """
+    skip_gcs = local_mode or dry_run
     if not any(
         [additional_families, additional_samples, samples_n, families_n, cohorts]
     ):
@@ -333,16 +336,31 @@ def main(
         upserted_participant_map = transfer_participants(
             target_project=target_project,
             participant_data=participant_data,
+            dry_run=dry_run,
         )
 
     else:
         logger.info(f'Transferring {len(participant_data)} participants. ')
         family_ids = transfer_families(
-            project, target_project, internal_participant_ids
+            project, target_project, internal_participant_ids, dry_run=dry_run
         )
-        upserted_participant_map = transfer_ped(project, target_project, family_ids)
+        upserted_participant_map = transfer_ped(
+            project,
+            target_project,
+            family_ids,
+            participant_data=participant_data,
+            dry_run=dry_run,
+        )
 
     existing_data = query(EXISTING_DATA_QUERY, {'project': target_project})
+
+    # Skipping GCS means no real files to reheader; disable the Hail Batch jobs.
+    if skip_gcs and update_embedded_ids:
+        logger.info(
+            '--local-mode / --dry-run: disabling --update-embedded-ids '
+            '(no real files to reheader)'
+        )
+        update_embedded_ids = False
 
     # Create batch if needed for running header updating jobs
     if update_embedded_ids:
@@ -360,6 +378,8 @@ def main(
         upserted_participant_map=upserted_participant_map,
         target_project=target_project,
         project=project,
+        skip_gcs=skip_gcs,
+        dry_run=dry_run,
     )
 
     logger.info('Transferring analyses')
@@ -371,6 +391,8 @@ def main(
         old_sid_to_new_sid,
         sample_to_sg_attribute_map,
         update_embedded_ids,
+        skip_gcs=skip_gcs,
+        dry_run=dry_run,
     )
     logger.info('Subset generation complete!')
 
@@ -385,6 +407,8 @@ def transfer_samples_sgs_assays(
     upserted_participant_map: dict[str, int],
     target_project: str,
     project: str,
+    skip_gcs: bool = False,
+    dry_run: bool = False,
 ):
     """
     Transfer samples, sequencing groups, and assays from the original project to the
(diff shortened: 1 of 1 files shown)


scripts/create_test_subset.py
```diff
@@ -902,6 +1006,19 @@
             # Participants are being created before the samples are, so this will be empty for now.
             participants_to_transfer.append(transfer_participant)
 
+    if dry_run:
+        logger.info(
+            f'[dry-run] Would papi.upsert_participants(project={target_project!r}) with {len(participants_to_transfer)} participants:'
+        )
+        for p in participants_to_transfer:
+            logger.info(f'[dry-run]   {p}')
```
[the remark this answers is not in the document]
joshuamschmidt: @michael-harper to me this actually looks like a good suggestion.

scripts/create_test_subset.py
```diff
@@ -332,17 +336,32 @@ def main(
         upserted_participant_map = transfer_participants(
             target_project=target_project,
             participant_data=participant_data,
```
joshuamschmidt, line 421: `participant_data` can contain the same participant twice when the random sampler picks two samples that share a participant. On local testing the upsert errored due to duplicate IDs. Worth filtering the sampler to `whole-blood` and/or sampling at participant granularity?

scripts/create_test_subset.py
joshuamschmidt, line 300: Footgun. despite the name of the query at https://github.com/populationgenomics/metamist/pull/1276/changes/BASE..c20fed1a350173e8d6e36deb4a72e97a67add81f#diff-a0a08a9dd29ea8495cf15c3e60d2c724d8f1048bcc9fc056963d6f19f053fab6L142
`sid['id'] for sid in sid_output.get('project').get('samples')` gives sampleIDs not seq group IDs!
This is the ultimate cause of the duplicate ID upsert error i think.

joshuamschmidt: relatedly, as sampling is on samples not seq groups, we can get samples without seq groups, which is useless for building test data bucket.

scripts/create_test_subset.py
michael-harper: I had to deep-dive this. `get_sids_by_random_sampling` is only invoked under if `samples_n`, so if we pass `--cohorts` without `-n`, the random sampler never runs. This is therefore not on the code path we'd exercise with a just cohort invocation.

However, the duplicate-participant isn't unique to `get_sids_by_random_sampling`. It also comes from a loop in main at:
```
# Pull Participant Data
    participant_data = []
    internal_participant_ids: list = []
    for sg in original_project_subset_data.get('project').get('samples', []):
        participant = sg.get('participant')
        if participant:
            participant_data.append(participant)
            internal_participant_ids.append(participant.get('id'))
```
that builds `participant_data` per-sample regardless of how the samples were chosen. So if our cohort happens to contain two SGs whose parent samples share a participant, we'd still trip the upsert error via the cohort path. 

We could either sample at participant-level and guarantee there'd be no duplicate participants (major change) or as you said, filter samples on `whole-blood`, I'm not sure how restrictive it is to hard-code filtering on whole blood samples? I can imagine it being useful in the future to allow the user to filter on different sample types? Thoughts? Bit of scope creep happening here if we want to expand to selecting sample types

scripts/create_test_subset.py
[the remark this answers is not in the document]
michael-harper: Done. Also replaced family and ped file logging in other places

scripts/create_test_subset.py
michael-harper: Good call. pushed a one-line filter in `get_sids_by_random_sampling` so the candidate pool excludes samples whose sequencingGroups list is empty (8da84c87). The GraphQL resolver already defaults to `active_only=True` for SGs, so samples whose only SGs are archived are naturally skipped too.

michael-harper: `sid` itself is the established convention in this file for sample id, SG ids are formatted `sgid` elsewhere in the codebase

Alexander-Stuckey (commented):
This all looks fine to me, but I am not a big user of this script, so I will defer to someone like @EddieLF

scripts/create_test_subset.py
michael-harper: Went with the source-side dedup option since the duplication is structural to the per-sample loop, independent of which selection path (`--cohorts`, `--samples`, `--families`, `-n`) brought us in. Build a `dict[participant_id → participant]` from the returned samples and derive `participant_data` / `internal_participant_ids` from the map at the end — single source of truth, dedup falls out of dict semantics, and the two derived lists can't drift.
Pushed as 82f2e8f9.
Punted on the sample-type filter / participant-level sampler for now — agree both are reasonable but they're separate features, and a --sample-types whole-blood,pbmc flag is the more general version of "filter on whole-blood" if we want to go that way later. Happy to scope that into a follow-up issue.

joshuamschmidt (commented):
LGTM

MattWellie:
Looks reasonable. From the description it looks like this would solve issues in the transfer of DNA + RNA samples, but I don't think we've had issues so far/haven't attempted it.

Having two separate dry-run/local args is odd, if this is run with `local` it will transfer new samples to metamist, and if it later runs with the same input in full will it duplicate, or will it just transfer the files and attach them to the existing copied samples/SGs? I think the latter, but haven't used this in a while. To me local implies using a local/dev instance of metamist, but that doesn't appear to be the case

michael-harper:
@MattWellie  

Before each upsert it runs `EXISTING_DATA_QUERY` against the target project and matches by external id (samples, participants), (type, platform, technology) (SGs), type (analyses), so a second run updates existing records rather than creating new ones. So `--local-mode` followed by a full run wouldn't duplicate it'd update the existing rows and additionally do the GCS copies the first run skipped.

On "local" being misleading: agreed. I added a check in 95fee6b2 where `--local-mode` now refuses to run unless `SM_ENVIRONMENT=local`. That way it can't accidentally write subset records into prod metamist pointing at GCS paths that were never populated.

Kept the two-flag form for now because the use cases are distinct: `--local-mode` is "seed a local DB without touching GCS", `--dry-run` is "log what would happen, safe to run anywhere read-only." 

On DNA + RNA transfers: yep, same problem class as WGS + genotyping-array (one sample, multiple SG types of different type), which is what the main fix in this PR addresses. So this should unblock that scenario when you get to it.

jmarshall:
The fundamental concern with trying to review this PR is that it seems to do several unrelated things:

1. I understand it fixes buggy behaviour around sequencing group selection (see second bullet point in the PR description).
2. It adds new `genotypingarray_gtc` analysis creation, though it's unclear to this reviewer where the dividing line between bug fixes (1 above) and new behaviour is.
3. It adds two new modes `local_mode` and `dry_run` (which is not mentioned in the PR description). Admittedly the existing script is fairly spaghetti-like, but the proposed implementation adds quite a bit more spaghetti and special cases.

As discussed at office hours today, it might be good to split this up into multiple smaller PRs, or at least to describe the motivation for the various facets of the change in more detail.

jmarshall:
With the addition of “local mode”, would you ever **not** want local mode when operating against `SM_ENVIRONMENT=LOCAL`?

scripts/create_test_subset.py
```diff
@@ -830,14 +931,23 @@ def transfer_families(
...
 def transfer_ped(
-    initial_project: str, target_project: str, family_ids: list[int]
+    initial_project: str,
+    target_project: str,
+    family_ids: list[int],
+    participant_data: list[dict] | None = None,
+    dry_run: bool = False,
 ) -> dict[str, int]:
     """Pull pedigree from the input project, and copy to target_project"""
     ped_tsv = fapi.get_pedigree(
```
jmarshall, line 985: This is the **one** API call not gated by `if dry_run`. I assume this is intentional, and because this is getting some information rather than updating the database?

joshuamschmidt:
> With the addition of “local mode”, would you ever **not** want local mode when operating against `SM_ENVIRONMENT=LOCAL`?

@jmarshall thanks for that observation - i think you are right and probably suggests a cleaner way to handle this than yet another command line flag. 

It wasn't clear to us (well certainly me) how you do a full test against a local metamist deploy given none of the files/analysis objects etc actually exist.

joshuamschmidt:
@jmarshall re 

> 'fixes buggy behaviour around sequencing group selection .... adds new genotypingarray_gtc analysis creation, though it's unclear to this reviewer where the dividing line between bug fixes (1 above) and new behaviour is.

for you is the main issue the mixing of concerns in a single PR? or the lack of
clear delimitation of, and separate justification, of touching both at once? or both?

In this case it seemed like we could only add a genotypingarray_gtc analysis into copy over to test by fixing the sequencing group selection buggy behaviour.

But if we were to pull out just the buggy sequencing group selection, then why would we stop just there? As you suggested this morning, i think this entire module should be refactored, but being in the metamist repo, such an undertaking would feel like a faux-pas coming from outside team software.

scripts/create_test_subset.py
michael-harper: Yep, the updating of the pedigree is done in `fapi.import_pedigree`, no need to gate `fapi.get_pedigree`.

As an aside we can also technically move the
```
tmp_ped_tsv = 'tmp_ped.tsv'
# Work-around as import_pedigree takes a file.
with open(tmp_ped_tsv, 'w') as tmp_ped:  # noqa: PTH123
    tmp_ped.write(ped_tsv)
```
below the `if dry_run` check to avoid an unnecessary write of `tmp_ped.tsv`

jmarshall:
My concern about the dry-run implementation is that it adds a bit of spaghetti and in particular changes the behaviour of various functions depending on whether we're in dry-run mode.

PR #1282 suggests an alternative approach, of stubbing out `aapi` et al with subclasses that just log the operation instead of doing any updates. It might need a bit of massaging but may be interesting to test.

scripts/create_test_subset.py
jmarshall: For the record, the other “read-only” API calls that aren't gated are — of course — a handful of graphql requests.

jmarshall (approved):
With the reduction in dry-run spaghetti :smile: I think this is now merge-worthy.

I suspect the `skip_gcs` stuff might benefit from some similar kind of refactoring, but it's not so bad and so that could occur later if at all. So in the interests of getting to use the new script facilities, this is LGETGTM (GETG = good enough to go).

Up to you whether you want to change `--local-mode` to just an `SM_ENVIRONMENT` check in this PR, or also leave that for an optional future date.

michael-harper:
Thanks @jmarshall ! In the spirit of de-spaghettification I have removed the `--local-mode` flag.

Confirmed tested and working locally

PR #1277, opened 2026-05-26
Extend ourdna test generator to create SGs, assays, and analyses

joshuamschmidt:
## Summary
The PopGen team has multiple datatypes in prod that need to be captured in test datasets. Currently,
`test/data/generate_ourdna_data.py` only provides participants + nested samples. This PR extends functionality to also produce two data (sequencing group) types representing WGS and snp genotyping data by making:

- Two sequencing groups on every whole-blood sample: a `genome` SG (`illumina` / `short-read`, one `sequencing` assay with R1+R2 fastq reads) and a `genotypingarray` SG (`illumina` / `infinium-global-diversity-array-v1.0`, no assays).
- One `cram` and one `gvcf` analysis on each genome SG; one `genotypingarray_gtc` analysis on each array SG.
- The sequencing / assay / analysis enum types this generator depends on.

The result is a local/dev metamist DB ourdna dataset with multi-SG-type whole-blood samples.

Re PR Checklist:

no existing tests for this script

### DIFF
--- a/test/data/generate_ourdna_data.py
+++ b/test/data/generate_ourdna_data.py
@@ -64,9 +64,9 @@ SG_QUERY = gql(
     """
 )
 
-WHOLE_BLOOD_WITHOUT_SGS_QUERY = gql(
+WHOLE_BLOOD_QUERY = gql(
     """
-    query WholeBloodWithoutSGs($project: String!) {
+    query WholeBlood($project: String!) {
         project(name: $project) {
             samples(type: {eq: "whole-blood"}) {
                 id
@@ -475,21 +475,13 @@ def register_enums(enums_api: EnumsApi) -> None:
     for typ in sample_types:
         enums_api.post_sample_types(new_type=typ)
 
-    for typ in (SG_TYPE_GENOME, SG_TYPE_ARRAY):
-        enums_api.post_sequencing_types(new_type=typ)
-
-    for plat in {defn['platform'] for defn in SG_DEFINITIONS.values()}:
-        enums_api.post_sequencing_platforms(new_type=plat)
-
-    for tech in {defn['technology'] for defn in SG_DEFINITIONS.values()}:
-        enums_api.post_sequencing_technologys(new_type=tech)
-
     enums_api.post_assay_types(new_type='sequencing')
-
-    for analysis_type in {
-        a for defn in SG_DEFINITIONS.values() for a in defn['analyses']
-    }:
-        enums_api.post_analysis_types(new_type=analysis_type)
+    for sequencing_type, defn in SG_DEFINITIONS.items():
+        enums_api.post_sequencing_types(new_type=sequencing_type)
+        enums_api.post_sequencing_platforms(new_type=defn['platform'])
+        enums_api.post_sequencing_technologys(new_type=defn['technology'])
+        for analysis_type in defn['analyses']:
+            enums_api.post_analysis_types(new_type=analysis_type)
 
 
 def attach_sgs_to_whole_blood_samples(sample_api: SampleApi, project: str) -> None:
@@ -500,7 +492,7 @@ def attach_sgs_to_whole_blood_samples(sample_api: SampleApi, project: str) -> No
     SGs declared on nested whole-blood samples are silently dropped. Run this
     as a second pass against the freshly-created whole-blood sample IDs.
     """
-    resp = query(WHOLE_BLOOD_WITHOUT_SGS_QUERY, {'project': project})
+    resp = query(WHOLE_BLOOD_QUERY, {'project': project})
     samples = [s for s in resp['project']['samples'] if not s.get('sequencingGroups')]
     if not samples:
         return
@@ -524,8 +516,7 @@ def create_analyses_for_new_sgs(
 
     Genome SGs get a cram + gvcf, array SGs get a genotypingarray_gtc. Paths use
     a FAKE:// scheme so metamist skips GCS validation and stores the path as a
-    plain string in analysis_outputs — Harper's PR exercises the SG-matching
-    code regardless of whether outputs comes back as a string or a dict.
+    plain string in analysis_outputs.
     """
     sg_resp = query(SG_QUERY, {'project': project})
     new_sgs = [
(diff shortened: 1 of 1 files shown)


test/data/generate_ourdna_data.py
```diff
@@ -363,11 +472,100 @@ def main(project='ourdna', num_participants=10):
...
+                    type=analysis_type,
+                    status=AnalysisStatus('completed'),
+                    sequencing_group_ids=[sg_id],
+                    output=path,
+                    meta={'sequencing_type': sg_type},
+                ),
+            )
+
+
+def main(project='ourdna', num_participants=5):
```
jmarshall: I'd just remove the defaults here — because the parser sets `default=` these arguments will always be present, and the default values are already specified below. And if they're not present, you'll want the `main` invocation to fail.


```suggestion
def main(project: str, num_participants: int):
```

test/data/generate_ourdna_data.py
```diff
@@ -363,11 +472,100 @@ def main(project='ourdna', num_participants=10):
...
+
+    for tech in {defn['technology'] for defn in SG_DEFINITIONS.values()}:
+        enums_api.post_sequencing_technologys(new_type=tech)
+
+    enums_api.post_assay_types(new_type='sequencing')
+
+    for analysis_type in {
+        a for defn in SG_DEFINITIONS.values() for a in defn['analyses']
+    }:
+        enums_api.post_analysis_types(new_type=analysis_type)
```
jmarshall: I think this would be better written more straightforwardly:


```suggestion
    enums_api.post_assay_types('sequencing')
    for sequencing_type, defn in SG_DEFINITIONS.items():
        enums_api.post_sequencing_types(sequencing_type)
        enums_api.post_sequencing_platforms(defn['platform'])
        enums_api.post_sequencing_technologys(defn['technology'])
        for analysis_type in defn['analyses']:
            enums_api.post_analysis_types(analysis_type)
```

test/data/generate_ourdna_data.py
```diff
@@ -13,8 +13,70 @@
...
+                    id
+                    type
+                }
+            }
+        }
+    }
+    """
+)
+
+WHOLE_BLOOD_WITHOUT_SGS_QUERY = gql(
```
jmarshall: This query is really `WHOLE_BLOOD_QUERY`; we're going to do the `WITHOUT_SGS` part later in code that filters the result of this query.

test/data/generate_ourdna_data.py
```diff
@@ -363,11 +472,100 @@ def main(project='ourdna', num_participants=10):
...
+
+def create_analyses_for_new_sgs(
+    analysis_api: AnalysisApi, project: str, existing_sg_ids: set[str]
+) -> None:
+    """
+    Create one analysis per analysis type for every SG newly added to the project.
+
+    Genome SGs get a cram + gvcf, array SGs get a genotypingarray_gtc. Paths use
+    a FAKE:// scheme so metamist skips GCS validation and stores the path as a
+    plain string in analysis_outputs — Harper's PR exercises the SG-matching
```
jmarshall: Which PR? This is not really a useful comment for posterity… :smile:

test/data/generate_ourdna_data.py
```diff
@@ -363,11 +472,100 @@ def main(project='ourdna', num_participants=10):
...
+        sg
+        for sg in sg_resp['project']['sequencingGroups']
+        if sg['id'] not in existing_sg_ids
+    ]
+    print(f'creating analyses for {len(new_sgs)} new sequencing groups')
+
+    for sg in new_sgs:
+        sg_id = sg['id']
+        sg_type = sg['type']
+        existing_analysis_types = {a['type'] for a in sg.get('analyses') or []}
```
jmarshall:
```suggestion
        existing_analysis_types = {a['type'] for a in sg.get('analyses', [])}
```

test/data/generate_ourdna_data.py
```diff
@@ -363,11 +472,100 @@ def main(project='ourdna', num_participants=10):
...
+
+        for analysis_type in SG_DEFINITIONS[sg_type]['analyses']:
+            if analysis_type in existing_analysis_types:
+                continue
```
jmarshall: I'd invert this:

```suggestion
            if analysis_type not in existing_analysis_types:
                path = …
                analysis_api.create_analysis(…
```

test/data/generate_ourdna_data.py
joshuamschmidt: thanks - agreed.

test/data/generate_ourdna_data.py
joshuamschmidt: d'oh

test/data/generate_ourdna_data.py
joshuamschmidt: @jmarshall q: you suggested this change is so that a stored `None` value is returned as `None`, and not modified with `or []`?

jmarshall: Oh, can that come back as “database entry is None” as opposed to omitted altogether from the returned data?

Okay, fair enough — whether it's a likely scenario or not, probably best to be defensive w.r.t. that possibility and leave the `sg.get(…) or []` code as you have it.

joshuamschmidt:
Thanks for the review @jmarshall. I made changes for 5 of your 6 suggestions. Just had a follow up question: https://github.com/populationgenomics/metamist/pull/1277#discussion_r3322294594

test/data/generate_ourdna_data.py
joshuamschmidt: Yeah TIL that in GraphQL, the specification requires the server to return the field key explicitly with a null value.

joshuamschmidt: Actually @jmarshall, having thought about this some more:
if we were really worried that the `analyses` key was absent from the db, `sg.get('analyses') or []` seems like a poor way to actually handle that.

 whereas, your change to `sg.get('analyses', [])}` makes no pretensions to that fact, so is better. right?

joshuamschmidt:
@jmarshall all of your suggestions implemented now

jmarshall (approved):

PR #1278, opened 2026-05-27
[PG] fix `merge_samples` endpoint

hrrsheen:
The query for this endpoint was attempting to modify an incorrect table. It was also not correctly deleting the merged sample. This PR fixes these issues and adds a test for the endpoint.

### DIFF
--- a/db/python/tables/sample.py
+++ b/db/python/tables/sample.py
@@ -515,7 +515,7 @@ class SampleTable(DbBase):
         # Query to update the sequencing group table replacing the merge sid with
         # the id to keep
         queries.append(t"""
-            UPDATE analysis_sequencing_group
+            UPDATE sequencing_group
             SET sample_id = {id_keep}, audit_log_id = {audit_log_id}
             WHERE sample_id = {id_merge}
         """)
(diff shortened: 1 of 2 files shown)


db/python/tables/sample.py
```diff
@@ -515,7 +515,7 @@ def dict_merge(meta1, meta2):
         # Query to update the sequencing group table replacing the merge sid with
         # the id to keep
         queries.append(t"""
-            UPDATE analysis_sequencing_group
+            UPDATE sequencing_group
```
jmarshall, line 518: So this was always a typo and should have been `sequencing_group` all along, in particular because `analysis_sequencing_group` doesn't have a `sample_id` field! And we want to move the sequencing groups from sample `merge_id` to `keep_id` => okay, LGTM.

db/python/tables/sample.py
```diff
@@ -527,6 +527,12 @@ def dict_merge(meta1, meta2):
             WHERE id = {id_merge}
         """)
 
+        # Query to delete the sample's external id
```
jmarshall, line 530: Samples can have multiple external ids. We might want to think about this, but it seems like it would be good to merge the external ids too. Probably like this: reassign all of `merge_id`'s external ids to `keep_id`, but where they both have an entry for the same `name` then `keep_id`'s value wins.

PR #1279, opened 2026-06-03
Add CONTEXT.md domain glossary

folded:
## Summary

Adds a root `CONTEXT.md` fixing the domain vocabulary for metamist, inferred from the codebase. The goal is one word per concept so code and conversations stay consistent.

Covers:
- **Core entities** — Project, Participant, Family, Sample, SequencingGroup, Assay, Analysis, Cohort, Comment/Discussion, AnalysisRunner, AuditLog
- **Identity & access** — Internal ID vs Formatted ID (prefix + Luhn checksum), External ID/org, ProjectMemberRole and the ReadAccessRoles/FullWriteAccessRoles sets
- **Model-variant convention** — the Internal / External / Upsert / Nested / GraphQL suffixes
- **Navigation terms** — Layer / Table / Filter
- **Relationships**, an example dialogue, and flagged ambiguities

## Opinionated calls worth a look

These were inferred, not read from existing docs — please sanity-check against how the team actually speaks:
- Named the wet-lab record **Assay** and reserved **SequencingGroup** for the immutable analysis unit (the code uses "sequence/sequencing" loosely).
- Coined **Formatted ID** vs **Internal ID** rather than "CPG ID", since prefixes are env-configurable (`XPGLCL`/`CPGLCL`/`COH`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,153 @@
+# Metamist Context
+
+Metamist is a database and API for storing **de-identified -omics metadata**: the
+participants, biological samples, sequencing, and downstream analyses involved in
+population-genomics studies. This file fixes the domain vocabulary so code and
+conversations use one word per concept.
+
+## Language
+
+### Core entities
+
+**Project**:
+The top-level organisational and access-control unit; every other entity belongs to
+exactly one. Has a `name`, a `dataset`, and per-user `roles`.
+_Avoid_: dataset (a Project *has* a dataset name, but the access-scoped entity is the Project), workspace, tenant
+
+**Participant**:
+A de-identified study subject (a person), carrying `external_ids`, reported sex/gender,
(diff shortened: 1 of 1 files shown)


MattWellie (commented):
Appears accurate in terms of the object hierarchy. The flagged ambiguities are well resolved, probably out of scope for this to track back and clean up all the "sequence*" instances in code. 

This was generated from the contents of this repo, but would the better location for it be in cpg-skills' existing metamist.md which is currently pretty thin, and already has a route to being accessed as a file-free skill in any claude session?

folded:
It could be referenced in a skill, but I also believe that it's useful documentation to have live alongside the code - assuming it's right, and stays current - because humans can refer to it as well.

PR #1280, opened 2026-06-03
SET 1103 - [PG] Sequencing groups upsert does not insert new assays

hrrsheen:
The behavior of sequencing groups when being upserted is unclear when looking at some of the code in the `upsert_sequencing_groups` layer function. What should happen is that the existing sequencing group is archived and the upsert assays are attached to a new SG.
This PR removes logic that obfuscates this functionality. Additionally a test has been added to ensure functionality is as expected when modifying a sequencing group's assays.

### DIFF
--- a/db/python/layers/sequencing_group.py
+++ b/db/python/layers/sequencing_group.py
@@ -243,11 +243,6 @@ class SequencingGroupLayer(BaseLayer):
                     to_update.append(sg)
                     continue
 
-                # if we need to insert any assays, then the group will have to change
-                if any(not assay.id for assay in sg.assays):
-                    to_replace.append(sg)
-                    continue
-
                 assert sg.id is not None
                 existing_sequences = set(sequence_to_group.get(int(sg.id), []))
                 new_assay_ids = set(sq.id for sq in sg.assays)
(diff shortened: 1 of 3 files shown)


test/test_sequencing_groups.py
```diff
@@ -158,6 +158,74 @@ async def test_update_sequencing_group(
...
+        new_sg_id = new_sg[0].id
+        assert new_sg_id
+        assert new_sg_id != initial_sg_id
+
+        # Check that the first assay belongs to both the old and new sequencing group
+        old_sg_assays = await a_layer.get_assays_for_sequencing_group_ids(
+            [initial_sg_id]
+        )
+        new_sg_assays = await a_layer.get_assays_for_sequencing_group_ids([new_sg_id])
+        old_sg_assays_ids = set(a.id for a in old_sg_assays[initial_sg_id])
```
jmarshall: I think what the linter is hinting at is `old_sg_assays_ids = {a.id for a in …}`

db/python/layers/sequencing_group.py
```diff
@@ -271,9 +271,16 @@ async def upsert_sequencing_groups(
...
+            new_assay_ids = []
+            for assay in sg.assays or []:
+                if assay.id is None:
+                    raise ValueError(
+                        'Attempting to replace sequencing-group using non-existent assays'
+                    )
+                new_assay_ids.append(assay.id)
             sg.id = await self.recreate_sequencing_group_with_new_assays(
                 sequencing_group_id=int(ensure_nonnone(sg.id)),
-                assays=[s.id for s in sg.assays] if sg.assays else [],
```
jmarshall, line 281: Looking at it again, and looking at the `assay_ids = […]` line in the `to_insert` bit above, I think this can just be a job for `ensure_nonnone`:

```
assays=[ensure_nonnone(a.id) for a in sg.assays] if sg.assays else [],
```

(and change the `s` variable to `a`) and that'll shut up the linter.

jmarshall (approved):

PR #1281, opened 2026-06-03
SET-1046 Add Total column to Cost Across Invoice Months (Topic only)

nevoodoo:
## Summary

The PM team requested a per-row summary on the **Cost Across Invoice Months (Topic only)** report (`/billing/costByMonth`). This adds a final **Total** column that sums each row across all visible invoice months.

## Changes

- **`BillingCostByMonthTable.tsx`** (on-screen table)
  - New `rowTotal(key, compType)` helper — null-safe sum of `data[key][month][compType]` across all months.
  - Added a bold `Total` cell as the last cell of every body row.
  - Added the matching `Total` header, plus a spacer cell in the top "Invoice Month" header row to keep columns aligned.
- **`BillingCostByMonth.tsx`** (CSV/TSV export)
  - Added `'Total'` to the header fields and a `categoryTotal()` helper, appending the per-row total to all four exported row types so the export mirrors the table exactly.

Existing behaviour preserved: the synthetic "All Topics" row and the Avg. Sample Cost row-skipping logic are unchanged.

## Testing

- ESLint: clean on both files.
- `tsc`: 14 errors on clean `dev` vs 14 with these changes — identical, and none in the edited regions. The pre-existing errors are in unrelated billing files (`BillingCostBySample`, `BillingHome`, `SequencingGroupsByMonth`) plus a stale `sm-api` model mismatch — not introduced here.

### DIFF
--- a/web/src/pages/billing/BillingCostByMonth.tsx
+++ b/web/src/pages/billing/BillingCostByMonth.tsx
@@ -438,7 +438,6 @@ const BillingCostByMonth: React.FunctionComponent = () => {
                         const val = data[topic]?.[m]?.[CloudSpendCategory.SAMPLE_STORAGE_COST]
                         return val === undefined ? '' : val.toFixed(2)
                     }),
-                    categoryTotal(topic, CloudSpendCategory.SAMPLE_STORAGE_COST),
                 ]
                 matrix.push(sampleCostStorageRow)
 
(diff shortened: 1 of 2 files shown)


milo-hyben (approved):
This looking good, I have only removed the Totals for Avg Sample cost rows as it does not make sense to just summed them and those are not used by PM team.

PR #1282, opened 2026-06-03
Introduce DryRunFooApi subclasses to replace dry-run bespoke code

jmarshall:
Suggestion for PR #1276 which refactors the dry-run code to be less invasive.

### DIFF
--- a/scripts/create_test_subset.py
+++ b/scripts/create_test_subset.py
@@ -45,6 +45,71 @@ aapi = AnalysisApi()
 fapi = FamilyApi()
 papi = ParticipantApi()
 
+
+class DryRunAnalysisApi(AnalysisApi):
+    """Dry-run AnalysisApi that merely logs update methods."""
+
+    def create_analysis(self, project, analysis):
+        logger.info(f'[dry-run] Would aapi.create_analysis({project=}): {analysis}')
+
+    def update_analysis(self, analysis_id, analysis_update_model):
+        logger.info(
+            f'[dry-run] Would aapi.update_analysis({analysis_id=}): {analysis_update_model}'
+        )
+
+
+class DryRunFamilyApi(FamilyApi):
+    """Dry-run FamilyApi that merely logs update methods."""
+
+    def import_families(self, file, project):
+        families = file.readlines()
+        logger.info(
+            f'[dry-run] Would fapi.import_families({project=}) with {len(families)} families'
+        )
+
+    def import_pedigree(self, file, has_header, project, create_missing_participants):  # noqa: ARG002
+        ped_row_count = max(0, file.read().count('\n') - 1)
+        logger.info(
+            f'[dry-run] Would fapi.import_pedigree({project=}) with {ped_row_count} pedigree rows'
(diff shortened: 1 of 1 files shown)


michael-harper:
Thanks @jmarshall, definitely less spaghetti-like. Once we've confirmed these produce what we need locally, we'll approve from our end. 

Regarding local mode: I agree there's no case where we'd want GCS copies against `SM_ENVIRONMENT=local`, so I'll drop `--local-mode` entirely and derive `skip_gcs` from the environment instead. I'll fold that into #1276 after merging this refactor, to keep 1282 scoped to the dry-run change.
(There's a possible follow up to add file-existence checks into dry-run as a pre-flight, but that's out of scope here. I'll note it as a separate issue.)

michael-harper:
Just tried testing locally (not using `local_mode` only specifying `--dry-run`) and ran into the following: 
The logic in `DryRunSampleApi` doesn't call to the database and allocate a new id, instead it always returns `'NEW_SAMPLE_ID'` leading to a `KeyError` in `sample_to_sg_attribute_map[seid_sid_key].get(new_sg_attributes)`. 

This happens in the case where we have a freshly minted `ourdna-test` project with no participants or anything in there so `get_existing_sample` finds nothing for every sample meaning `existing_sid = None` and therefore `sample_upsert.id = None`. Therefore `sample_upsert.id if sample_upsert.id else 'NEW_SAMPLE_ID'` always returns `'NEW_SAMPLE_ID'`. 
That means all the values in `old_sid_to_new_sid` are set to `'NEW_SAMPLE_ID'`so in `transfer_analyses` the match `new_sample['id'] == old_sid_to_new_sid[old_sid]` gets the first sample for every lookup. It then builds `seid_sid_key = (new_sample['externalId'], old_sid)` a pair that was never inserted into `sample_to_sg_attribute_map` leading to a `KeyError`

joshuamschmidt:
~~@michael-harper, as a short term work around, can you use the generic generate test data script and create a few samples in ourdna-test?~~

Scrap that, misread

joshuamschmidt:
Ok think I understand now: the problem is that `DryRunSampleApi.create_sample` returns a constant 'NEW_SAMPLE_ID' for every sample that isn't already in the destination project. In the old code, the dry-run branch did new_sample_id = s['id'] (i.e. "old==new"). The new stub can't do that as it only sees sample_upsert, whose id is the destination id (or None for a genuinely new sample), never the source s['id']. Thus, every new sample collapses to the same value.

jmarshall:
Yeah, I was hoping to get away with that, but it's not surprising that we need to invent distinct ids. Should be easy to fix, as I believe these aren't going into the database or through any validation that they look like CPG ids, so we can just make them something distinctive like `NEW_SAMPLE_nnn` as they're just getting printed out…

jmarshall:
@michael-harper Okay, can you re-test with this version?

michael-harper:
Confirmed now working, thanks @jmarshall

michael-harper (approved):

PR #1285, opened 2026-06-05
SET 1097 - [PG] sequencing groups upsert does not handle external ids

hrrsheen:
Though the sequencing group model has a field for external IDs, this is never used during upsert. This PR adds a query to insert the sequencing group's external IDs to the `sequencing_group_external_id` table. Additionally a test has been added to ensure that this functions as expected.

### DIFF
--- a/db/python/tables/sequencing_group.py
+++ b/db/python/tables/sequencing_group.py
@@ -425,7 +425,7 @@ class SequencingGroupTable(DbBase):
             new_sg_id = await cur.fetchone()
             if not new_sg_id:
                 raise InternalError('A new sequencing_group row was not created')
-            
+
             if external_ids:
                 eid_values = [
                     {
--- a/db/python/tables/sequencing_group.py
+++ b/db/python/tables/sequencing_group.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from collections.abc import Mapping
 from datetime import date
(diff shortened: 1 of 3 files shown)


db/python/tables/sequencing_group.py
```diff
@@ -401,6 +403,13 @@ async def create_sequencing_group(
         RETURNING id;
         """
 
+        external_id_query = """
+        INSERT INTO sequencing_group_external_id
+            (project, sequencing_group_id, external_id, name, audit_log_id)
```
jmarshall, line 408: Minor, but canonically `name` is the key and `external_id` is the value so usually we put them the other way around.

test/test_sequencing_groups.py
```diff
@@ -62,7 +62,7 @@ def sequencing_group_model(test_sample: int) -> SequencingGroupUpsertInternal:
             'meta-key': 'meta-value',
         },
         sample_id=test_sample,
-        external_ids={'ext': 'some-ext-id'},
+        external_ids={'ext': 'some-ext-id', '': 'some-other-ext-id'},
```
jmarshall, line 65:
```suggestion
        external_ids={'ext': 'some-ext-id', PRIMARY_EXTERNAL_ORG: 'some-other-ext-id'},
```

dancoates (approved):

PR #1288, opened 2026-06-10
[SET-1169] Validate cohort SGs belong to the cohort's project

### LINKED ISSUE #1287: Cohort creation does not validate that member SGs belong to the cohort's project
## Summary
`createCohortFromCriteria` lets you create a cohort in `project-test` comprised of sequencing groups belonging to `project` (main). This results in broken cohorts, unusable because the sequencing groups and their associated assays etc are unusable within the scope of `project-test`. This is an easy error to make: pass a list of main/prod sg id's when creating a cohort in the dataset's `-test` project.

I would say this is unexpected behaviour.

## what is expected

sequencing groups passed in `sg_ids_internal` are checked against the destination project for membership.
trying to create a cohort in a project containing one ore more sequencing groups not belonging to that project should result in an error.

## what is observed

cohort creation proceeds without error, but with broken scope (note project is `{dataset}-test`, but cram path is `gs://cpg-{dataset}-main/exome/cram/CPGXXXXX.cram`:

```
"data": {
    "cohorts": [
      {
        "name": "{dataset}-test-exome-twist-20260528-dragen",
        "project": {
          "dataset": "{dataset}",
          "name": "{dataset}-test"
        },
        "sequencingGroups": [
          {
            "id": "CPGXXXXX",
            "sample": {
              "project": {
                "name": "{dataset}"
              }
            },
            "analyses": [
              {
                "outputs": {
                  "id": XXXXX,
                  "parent_id": null,
                  "path": "gs://cpg-{dataset}-main/exome/cram/CPGXXXXX.cram",
                  "basename": "CPGXXXXX.cram",
                  "dirname": "gs://cpg-{dataset}-main/exome/cram",
                  "nameroot": "CPGXXXXX",
                  "nameext": ".cram",
                  "file_checksum": "pziOag==",
                  "size": 921791013,
                  "meta": null,
                  "valid": true,
                  "secondary_files": {}
                }
```

nevoodoo:
## Summary

Closes #1287 

`createCohortFromCriteria` let you create a cohort in `project-test` comprised of sequencing groups belonging to `project` (main), producing broken, unusable cohorts (SGs and their assays/analyses are out of scope for the cohort's project).

The root cause: when SG IDs were supplied via `sg_ids_internal`, the layer derived the project filter *from the SGs themselves* (`get_projects_given_sg_ids`), never from the destination project, so SGs from any project passed validation.

## Change

In `CohortLayer.create_cohort_from_criteria`, when `sg_ids_internal` is provided, query the requested SGs (active + inactive) and verify each belongs to `project_to_write`. Any SG that doesn't triggers a clear `ValueError` naming the offending CPG IDs and the destination project. The check runs before the dry-run branch, so dry runs are validated too.

Scope matches the issue's "what is expected" section (the `sg_ids_internal` path).

## Tests

Added:
- `test_create_cohort_fails_when_sg_in_another_project`
- `test_create_cohort_dry_run_fails_when_sg_in_another_project`

Both create a second project, add an SG to it, and assert cohort creation in the original project rejects the foreign SG.

### DIFF
--- a/db/python/layers/cohort.py
+++ b/db/python/layers/cohort.py
@@ -18,6 +18,7 @@ from models.models.cohort import (
     NewCohortInternal,
 )
 from models.models.project import FullWriteAccessRoles, ProjectId, ReadAccessRoles
+from models.utils.sequencing_group_id_format import sequencing_group_id_format
 
 
 logger = get_logger()
@@ -218,9 +219,40 @@ class CohortLayer(BaseLayer):
                     )
                 raise ValueError(self.COHORT_SG_CRITERIA_ERROR_MSG)
 
-            projects = list(
-                await self.sglayer.get_projects_given_sg_ids(sg_ids_internal_raw)
+            # Every sequencing group passed in must belong to the project the
+            # cohort is being created in. Otherwise the cohort would reference
+            # sequencing groups (and their assays/analyses) that are out of
+            # scope for the project, producing a broken, unusable cohort.
+            requested_sgs = await self.sglayer.query(
+                SequencingGroupFilter(
+                    id=GenericFilter(in_=sg_ids_internal_raw),
+                    active_only=None,  # consider both active/inactive SGs
+                )
             )
+            if not requested_sgs:
+                raise ValueError(
+                    'Could not find any sequencing groups for the given ids'
+                )
+
+            external_sg_ids = sorted(
+                sg.id
+                for sg in requested_sgs
+                if sg.id and sg.project != project_to_write
+            )
+            if external_sg_ids:
+                dest_project = self.connection.project_id_map.get(project_to_write)
+                dest_project_name = (
+                    dest_project.name if dest_project else str(project_to_write)
+                )
+                raise ValueError(
+                    'The following sequencing groups do not belong to project '
+                    f'{dest_project_name}: '
+                    f'{", ".join(sequencing_group_id_format(s) for s in external_sg_ids)}. '
+                    'All sequencing groups in a cohort must belong to the project '
+                    'the cohort is created in.'
+                )
+
+            projects = [project_to_write]
         else:
             projects = cohort_criteria.projects
 
--- a/test/test_cohort.py
+++ b/test/test_cohort.py
@@ -1,4 +1,5 @@
 import datetime
+import uuid
 from random import randint
 
 from pymysql.err import IntegrityError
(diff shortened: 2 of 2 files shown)


rameshka:
Thanks for raising this PR @nevoodoo. 
The suggested changes seem to address https://github.com/populationgenomics/metamist/issues/1287,

I'm unclear about a few things related to the original functionality of the API. 

- The original method allows creating cohorts across different projects (SGs and Cohort belong to different projects)
   - So are there any valid use cases that still rely on this behaviour (If so, is performing this validation only during `dry_run` is sufficient)
- Here, the issue was raised when creating a cohort given the `sg_ids_internal` criterion (the project that SGs belong to and the destination project are different). But a similar scenario could happen if the cohort is created from other criteria as well. So does this validation need to be generic for cohort creation, given any criteria?
   - If that's the intended behaviour, it seems we would restrict cohort creation to sequencing groups within the same project as the cohort, which would make the `projects` criterion in `createCohortFromCriteria` redundant.

db/python/layers/cohort.py
```diff
@@ -218,9 +219,40 @@ async def create_cohort_from_criteria(
                     )
                 raise ValueError(self.COHORT_SG_CRITERIA_ERROR_MSG)
 
-            projects = list(
-                await self.sglayer.get_projects_given_sg_ids(sg_ids_internal_raw)
+            # Every sequencing group passed in must belong to the project the
+            # cohort is being created in. Otherwise the cohort would reference
+            # sequencing groups (and their assays/analyses) that are out of
+            # scope for the project, producing a broken, unusable cohort.
```
MattWellie, line 225: This isn't quite true, it's a bit more complicated. Here we want to guarantee that the project the cohort is created in has either direct or transitive access to the SGs. 

An example would be a Cohort in the `Seqr` umbrella project. If a Cohort contains SG IDs from multiple different projects (a common use case), the Cohort can't belong to a single dataset, it has to belong to a higher level project which has permission across multiple lower level projects. `seqr` doesn't contain any SG IDs itself, but through the transitive ownership via the rare-disease project group, it has permission to contain all RD SG IDs.

Not sure how easy it is to evaluate this in code or unit tests, but the final condition to satisfy is:

"all SG IDs in this new Cohort must be in the project the Cohort belongs to, or in any project which the owning project can transitively access"

rameshka: Thanks @MattWellie. This answers the unclear points I had earlier. Recalling from a [previous conversation](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1761715088843799)- Metamist doesn't have visibility of  the hierarchy of datasets/projects at the moment.

MattWellie: Ahh, that's not ideal. A really coarse version of this would be to make sure the destination project for the cohort and the per-SG datasets both either contain `-test` or not, but that's only solving a small portion of this problem

dancoates:
Closing this as we'll need a different approach that still allows multi-project cohorts

PR #1289, opened 2026-06-12
Increase GQL async client execute timeout and fix some scripts

EddieLF:
Increase the GQL async client default execute timeout and fix edge cases in the `create_test_subset` and `duplicate_sequencing_group` script s

## GQL async client
- Current timeout for the async client is 10 seconds which is ridiculously short. For samples with a number of assays and analyses this quickly triggers the timeout, e.g. https://batch.hail.populationgenomics.org.au/batches/1132974/jobs/1
- Changing this to 30s is a very minor and completely and backwards compatible change 

## create_test_subset.py 
- Only copy one of each unique analysis output to avoid cases where the same file is pointlessly copied and reheadered multiple times, e.g. in [this driver](https://batch.hail.populationgenomics.org.au/batches/1132984/jobs/1) and [submitted batch](https://batch.hail.populationgenomics.org.au/batches/1132986)

## duplicate_sequencing_group.py 
- Allow destination dataset to be a `-test` dataset - previously this was incompatible due to the find and replace of old dataset name with new dataset name when formulating the new file paths, leading to destination bucket paths like `cpg-dataset-test-main`, e.g. https://batch.hail.populationgenomics.org.au/batches/1132976/jobs/1

### DIFF
--- a/packages/metamist/src/metamist/graphql/__init__.py
+++ b/packages/metamist/src/metamist/graphql/__init__.py
@@ -85,7 +85,7 @@ async def configure_async_client(
     schema: str = None,
     auth_token: str = None,
     force_recreate=False,
-    execute_timeout: int | float | None = 30,
+    execute_timeout: int | float = 30,
 ) -> Client:
     """Configure an async client for use with the Metamist GraphQL API"""
     global _async_client  # noqa: PLW0603
--- a/packages/metamist/src/metamist/graphql/__init__.py
+++ b/packages/metamist/src/metamist/graphql/__init__.py
@@ -81,7 +81,11 @@ def configure_sync_client(
 
 
 async def configure_async_client(
-    url: str = None, schema: str = None, auth_token: str = None, force_recreate=False
+    url: str = None,
+    schema: str = None,
+    auth_token: str = None,
+    force_recreate=False,
+    execute_timeout: int | float | None = 30,
 ) -> Client:
     """Configure an async client for use with the Metamist GraphQL API"""
     global _async_client  # noqa: PLW0603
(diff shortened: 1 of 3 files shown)


MattWellie (approved):
LGTMatt

Johnnyassaf (approved):

packages/metamist/src/metamist/graphql/__init__.py
```diff
@@ -81,7 +81,11 @@ def configure_sync_client(
 
 
 async def configure_async_client(
-    url: str = None, schema: str = None, auth_token: str = None, force_recreate=False
+    url: str = None,
+    schema: str = None,
+    auth_token: str = None,
+    force_recreate=False,
+    execute_timeout: int | float | None = 30,
```
rameshka: Hey @EddieLF, [looking at the documentation](https://gql.readthedocs.io/en/stable/modules/gql.html), it looks like allowing `None` for `execute_timeout` won't be safe for the gql client. 

> execute_timeout – The maximum time in seconds for the execution of a request before a TimeoutError is raised. Only used for async transports. Passing None results in waiting forever for a response.

EddieLF: @rameshka those docs say `execute_timeout: Optional[Union[int, float]] = 10` - which is the same as `int | float | None` based on [typing.Optional](https://docs.python.org/3/library/typing.html#typing.Optional). So I think it's fine to allow `Nonetype` here. 

Sorry for taking a long time to get back to you.

rameshka: While the type annotation allows `None`, it causes the client to wait indefinitely for a response.

In this case, I think we should either hide/disallow `None`, since waiting forever generally isn't desirable, or explicitly document what `None` does.

EddieLF: That's a good observation, we wouldn't want indefinite runs to be allowed. I've removed the Nonetype type hint and explicitly set it to `int | float = 30`

rameshka (approved):
LGTM !!

PR #1299, opened 2026-07-10
[SET-1184] Inconsistent sample_type filtering in create_cohort_from_criteria

rameshka:
**Current behaviour** 
When creating a cohort using the **sample_type** criterion (optionally combined with other cohort criteria), if no samples match the requested sample_type, Metamist silently ignores the sample_type filter and applies only the remaining criteria to filter the SGs.

**Expected behaviour**
The created cohort should contain only SGs that satisfy both the requested sample_type and any other specified cohort criteria.

**Fix**
The API now validates the requested sample_type before creating the cohort. If no samples in the database match the specified sample_type, it returns an error instead of silently dropping the filter. 

For more details, See https://cpg-populationanalysis.atlassian.net/browse/SET-1184

### DIFF
--- a/db/python/layers/cohort.py
+++ b/db/python/layers/cohort.py
@@ -238,6 +238,10 @@ class CohortLayer(BaseLayer):
 
             _, samples = await self.sampt.query(sample_filter)
             sample_ids = [s.id for s in samples]
+            if not samples:
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


dancoates (approved):
Nice, looks like a good simple fix, and good to have the added test to avoid regressions.

rameshka:
haha thanks @dancoates for the quick review.

PR #1305, opened 2026-07-16
Parser more flexible sample type

EddieLF:
Updating the generic parser / generic metadata parser with added flexibility for sample type

- Set default sample type to 'unknown' (previously 'blood' or None)
- Allowing flexible sample type ingestion using a column in the sample file map manifest (like how sequencing type works)
- Forbid `sample_type = None` - users must either provide a populated `Sample Type` column in their manifest and set it when instantiating the parser OR set the `default_sample_type` value when instantiating the parser

### DIFF
--- a/packages/metamist/src/metamist/parser/generic_metadata_parser.py
+++ b/packages/metamist/src/metamist/parser/generic_metadata_parser.py
@@ -114,7 +114,7 @@ class GenericMetadataParser(GenericParser):
         # Default values
         default_reference_assembly_location: str | None = None,
         ora_reference_assembly_location: str | None = None,
-        default_sample_type: str | None = None,
+        default_sample_type: str | None = 'unknown',
         default_sequencing=DefaultSequencing(
             seq_type='genome', technology='short-read', platform='illumina'
         ),
@@ -277,16 +277,26 @@ class GenericMetadataParser(GenericParser):
         """Get internal cpg id from a row using get_sample_id and an api call"""
         return row.get(self.cpg_id_column, None)
 
+    def get_sample_type_from_row(self, row: SingleRow) -> str | None:
+        """Get sample type from row"""
+        if self.sample_type_column and self.sample_type_column in row:
+            value = row[self.sample_type_column]
+            if value:
+                return str(value).lower()
+        return self.default_sample_type
+
     def get_sample_type(self, row: GroupedRow) -> str:
         """Get sample type from row or default"""
-        value = (
-            row.get(self.sample_type_column, None)
-            or self.default_sample_type
-        )
-        value = value.lower()
-        if value:
-            return value
-        return None
+        if isinstance(row, dict):
+            return self.get_sample_type_from_row(row)
+        types = [
+            self.get_sample_type_from_row(r) for r in row
+        ]
+        if len(set(types)) > 1:
+            raise ValueError(
+                f'Conflicting sample types for sample {self.get_primary_sample_id(row[0])}: {types}'
+            )
+        return types[0] if types else self.default_sample_type
 
     def get_sequencing_types(self, row: GroupedRow) -> list[str]:
         """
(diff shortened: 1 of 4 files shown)


packages/metamist/src/metamist/parser/generic_metadata_parser.py
amyminiter, line 1020: This is `required=True` whereas all the other columns provided as click options are not. Deliberate?

packages/metamist/src/metamist/parser/generic_metadata_parser.py
```diff
@@ -273,11 +277,22 @@ async def get_cpg_sample_id_from_row(self, row: SingleRow) -> str | None:
...
+        if self.sample_type_column and self.sample_type_column in row:
+            value = row[self.sample_type_column]
+            if value:
+                return str(value).lower()
+        return self.default_sample_type
+
+    def get_sample_type(self, row: GroupedRow) -> str:
+        """Get sample type from row or default"""
+        types = [self.get_sample_type_from_row(r) for r in row]
+        if len(set(types)) > 1:
```
amyminiter, line 291: This could break if the data provider has forgotten to record sample types in the manifest for some rows i.e. `blood` in the first row, NULL for others. The NULL/empty values will be transformed to `unknown` and there will be > 1 sample type. 

We have received metadata before where (inexplicably) the first row only had a value for a column. It might be overkill to account for this. May happen if you're receiving metadata from sequencing providers and they don't have robust SOPs in place.

packages/metamist/src/metamist/parser/generic_metadata_parser.py
```diff
@@ -55,6 +55,7 @@
 python parse_generic_metadata.py \
     --project $dataset \
     --sample-name-column "Sample ID" \
+    --sample-type-column "unknown" \
```
MattWellie:
```suggestion
    --default-sample-type "unknown" \
```
Is this argument setting the column name to pick this up from, or the default value? The name and value seem to conflict

EddieLF: Good note thanks. Your suggestion is what it's meant to be.

packages/metamist/src/metamist/parser/generic_parser.py
amyminiter:
```suggestion
        self.default_sample_type: str | None = default_sample_type.lower() if default_sample_type else default_sample_type
```

Parameterised values are not normalised; values read from the manifest are. Suggestion guards against different formatting and an eventual incorrect comparison between strings with inconsistent capitalisation. Example when manifest is read.
Row 1: value 'blood' → lowercased → 'blood'.
Row 2: empty → falls back to self.default_sample_type → 'Blood' (not lowercased).
types = ['blood', 'Blood'] → set(types) = {'blood', 'Blood'} → len > 1 → raises "Conflicting sample types" on data that's semantically identical.

packages/metamist/src/metamist/parser/generic_metadata_parser.py
EddieLF: Yup, this is intentional. The sample is the core entity in Metamist and any manifest ingested through these parsers is expected to have a column like "Sample ID", so this is enforced as mandatory.

packages/metamist/src/metamist/parser/generic_metadata_parser.py
EddieLF: Thanks Amy, I have been thinking about this, and I think it's ok to have the ValueError raised in this case. 

Scenario
---

The data ingestion handler attempts to ingest a manifest that looks like this:
```
Sample ID  Filenames     Sample Type
S01        S01_R1.fastq  Blood
S01        S01_R2.fastq 
```

As you mention, the value error will be raised
```
Conflicting sample types for sample S01: ['blood', 'unknown']
```

The data ingestion handler should then correct the manifest by filling in the rows with missing sample type before re-attempting ingestion.

---

IMO this explicit failure is better than carving out logic to handle this as an edge case by inferring intent and backfilling rows accordingly. What do you think?

amyminiter: Agree: a fix would require a lot of code to support an edge case. If you're happy, then I'm happy 👍

amyminiter (approved):
Thanks Ed. Looking good.

dancoates (approved):
Thanks @EddieLF and @amyminiter looking good

PR #1306, opened 2026-07-27
SET-1218 Support directory-like outputs (.mt/.ht/.vds) in OutputFiles model

nevoodoo:
### Problem

MatrixTable (`.mt`), Table (`.ht`) and VDS (`.vds`) outputs live on GCS as folders of blobs, not single objects. `get_file_info` only matches a blob by exact name, which never happens for a folder, so these outputs always came back `valid=False`. When a file is invalid we don't create an `output_file` row, so the output ends up stored as a plain string in `analysis_outputs.output` and there's no structured `outputs` dict to hang a `secondary_files` block off. That's why VDS/HT/MT analyses have been effectively second-class here.

### What this does

Instead of trying to fully validate a folder (summing every child blob's size, which could mean listing thousands of VDS shards), I treat these three extensions as directory-like and skip the checksum/size while still creating a real row:

- `get_file_info` now recognises `DIRECTORY_FORMAT_EXTENSIONS = {.mt, .ht, .vds}`. For those it does a cheap existence check (`list_blobs` with `max_results=1`) and, if anything is under the prefix, marks the output valid with a null checksum and size 0. That's enough to get an `output_file` row, which is what makes `secondary_files` work for these formats.
- `extract_bucket_params` strips a trailing slash, so `gs://bucket/foo.vds/` is handled the same as `gs://bucket/foo.vds` (previously the trailing slash gave an empty basename/extension).
- Removed the old `!= '.mt'` guard — it sat inside the exact-name match that never fires for a folder, so it was dead and didn't cover `.ht`/`.vds` anyway.

I also fixed the backfill script (`migrate_analysis_outputs.py`). It lists a whole bucket prefix with a `/` delimiter, which folds folder contents into prefixes rather than returning them, so the pre-fetched blob list can never confirm a folder exists. Directory-format outputs now get reprocessed with `blobs=None` so the bounded existence check runs instead.

### Worth calling out

`valid=True` for a directory now means "the folder exists", not "checksum verified". Existing MT/VDS/HT analyses that are already stored as strings won't change until the migrate script is re-run.


models/models/output_file.py
```diff
@@ -105,6 +110,33 @@ def list_blobs(
             print(f'Could not find bucket {bucket}: {e}')
             return []
 
+    @staticmethod
+    def directory_exists(
```
MattWellie, line 120: There's something more specific we can do here - following the existence check model we have elsewhere e.g. [CPG-Utils](https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/existence_checks.py#L48-L51) (re-used/stolen from CPG-Flow)

The final file written as part of a MT/HT/VDS is a sentinel `_SUCCESS` - without that, the Hail process didn't finish writing. We can check for existence of that file using that method?

models/models/output_file.py
```diff
@@ -105,6 +119,39 @@ def list_blobs(
...
+        blobs: list[Blob] | None = None,
+    ) -> bool:
+        """
+        Check whether a directory-like path (a folder of blobs on GCS, e.g. a
+        MatrixTable or VDS) holds a completely written output, by looking for the
+        sentinel Hail writes last.
+
+        If a pre-fetched list of blobs is supplied, it is checked in-memory;
+        otherwise the sentinel is looked up on its own, which avoids enumerating
+        the potentially thousands of shards inside the folder. Listing rather than
```
dancoates: This comment is a bit funny, would be clearer to say that this uses list rather than getting the sentinel because the metamist server service account only has list permissions, at the moment the comment implies there are multiple users with different permissions.

models/models/output_file.py
```diff
@@ -16,6 +16,20 @@
 
 GCS_CLIENT = None
 
+# Outputs that are stored on GCS as folders of blobs rather than a single object
+# (Hail MatrixTable/Table and VDS). These have no single blob to checksum or size,
+# so we verify existence only. Hail writes a sentinel as the final step of the
+# write, which is what we look for — a folder without one is the wreckage of a run
+# that died partway through, not a usable output. A VDS is a pair of nested
+# MatrixTables, so it carries no sentinel of its own.
+# Mirrors cpg_utils.existence_checks.
```
dancoates: Might also be worth noting why we're not directly using cpg_utils.existence_checks here? Is it because metamist can only list blobs?

models/models/output_file.py
nevoodoo: Yep good point!

models/models/output_file.py
nevoodoo: Yep metamist only has the list permissions, have verified with cpg_infra to confirm this :(

